PR #174, opened 2026-04-11
perf: reduce memory usage and speed up board exports

swahlquist:
## Summary

Addresses repeated OOM kills on the staging web service and slow board copy/export operations. Combines Render infrastructure tuning with targeted code-level optimizations identified during a memory investigation.

The lingolinq-staging service was crashing 3-7 times per day, with memory hovering at 730-900MB at idle (on a 2GB plan) and spiking to 1.5GB+ during board operations. Root cause: Puma was running 3 forked workers by default, consuming ~950MB before any traffic.

### Render config (`render.yaml`)
- Set `WEB_CONCURRENCY=2` on web service (was unset, defaulting to 3 Puma workers consuming ~950MB at idle)
- Add `RUBY_YJIT_ENABLE=1` to web and worker for 15-30% Ruby speedup
- Add jemalloc (`LD_PRELOAD` + `MALLOC_CONF`) to worker service (was only set on web)

### Code optimizations
- **Add `oj` gem** with `Oj.mimic_JSON` initializer. 5-10x faster JSON parsing across the app, including BoardDownstreamButtonSet blobs and board content that gets parsed repeatedly during exports.
- **`BoardContent.load_content`**: add request/job-scoped Thread.current cache to avoid re-parsing the same board content multiple times within a single request or job.
- **`WordData.inflection_locations_for`**: add request/job-scoped cache for inflection lookups to avoid redundant DB queries.
- **`ApplicationController` + `Worker`**: clear request-scoped caches after each request/job to prevent data leaking between contexts.
- **`Converters::LingoLinq.to_external`**: load translations once per board instead of once per button (was an N+1 inside a loop).
- **`Converters::LingoLinq.to_external_nested`**: deduplicate images and sounds early to keep array sizes manageable across many boards.

## Verified Locally

All checks pass via Rails runner:
- Rails 7.2.3.1 boots cleanly with all changes
- Oj 3.16.12 loads at boot, `JSON.parse` returns same results as `Oj.load`
- `BoardContent.load_content` cache hit on 2nd call confirmed working
- `Worker.perform` override with `super` chain works correctly
- All edge cases handled (nil/empty inputs)

## Memory Impact (already observed)

Render metrics show staging memory dropped from **730-900MB baseline** to **426-480MB** after just the env var changes. Service has stopped restarting.

## Test plan

- [ ] Auto-deploy to lingolinq-staging completes successfully
- [ ] Existing test suite passes in CI
- [ ] Manual smoke test: copy a board set on staging, confirm it completes without OOM
- [ ] Manual smoke test: download a board set as PDF and OBZ
- [ ] Monitor Render memory metrics for 24 hours -- baseline should settle around 600-700MB with WEB_CONCURRENCY=2

## Follow-up Work (separate PRs)

These are larger items identified during research that should fix the slow (10-15 minute) board copy times:
- Replace rubyzip with `zip_kit` for streaming ZIP exports
- Adopt `Clowne` gem to replace the hand-rolled board copying logic in `relinking.rb`
- Replace Resque with GoodJob (database-backed, no Redis dependency, threaded model)
- Migrate Thread.current usage to `ActiveSupport::CurrentAttributes` for safer thread handling

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -36,6 +36,10 @@ gem 'concurrent-ruby', '~> 1.3'
 gem 'rails', '>= 7.2.3.1', '< 7.3'
 # CVE-2026-33210 (format string); bundler-audit advisory minimum
 gem 'json', '>= 2.19.2'
+# oj is a faster JSON parser/generator (5-10x faster than stdlib json).
+# Used via Oj.mimic_JSON in config/initializers/oj.rb to transparently
+# replace the JSON module across the app, including Rails internals.
+gem 'oj', '~> 3.16'
 # GHSA-46fp-8f5p-pf2m (allowed_uri?); rails-html-sanitizer 1.7.0 depends on loofah ~> 2.25; ensure >= 2.25.1
 gem 'loofah', '>= 2.25.1'
 gem 'pg', '~> 1.5'
--- a/Gemfile
+++ b/Gemfile
@@ -81,6 +81,8 @@ gem 'ttfunk', '1.7'
 gem 'ruby-saml'
 gem 'rotp'
 
+gem 'clowne', '~> 1.4' # Declarative model cloning DSL for board copy optimization
+
 gem 'sinatra', '~> 4.2'
 gem 'sanitize'
 gem 'anthropic', '~> 1.23'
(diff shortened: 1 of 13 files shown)


MelissaOneil (commented):
looks good

MelissaOneil:
@copilot apply changes based on the comments in [this thread](https://github.com/lingolinq/LingoLinq-AAC/pull/174#pullrequestreview-4095759833)

PR #184, opened 2026-04-15
fix: hide Tarheel Reader behind feature flag (off by default)

swahlquist:
## Summary
- Tarheel Reader was acquired by Building Wings in Sept 2024 and moved to Monarch Reader. The old `tarheelreader.org` JSON endpoints now 301-redirect to a closed Vue SPA, returning HTML instead of JSON.
- `Uploader.find_resources` was calling those endpoints and `JSON.parse`-ing the HTML response, surfacing a broken experience to users.
- Added a `tarheel_reader` feature flag (in `AVAILABLE_FRONTEND_FEATURES` but NOT `ENABLED_FRONTEND_FEATURES`) and guarded the `source=tarheel` / `source=tarheel_book` paths so they return `[]` by default.
- Users still see the Tarheel Reader category UI in `focus-words.js` but tapping through returns an empty result set instead of a broken error.

## Why not remove entirely?
- Individual users can still be opted in via `settings['feature_flags']['tarheel_reader'] = true` for testing when we negotiate Monarch Reader partner access or wire up an alternate book source (Unite for Literacy, Bookshare, etc.).
- Full UI removal is a larger, multi-file change; this is the minimum safe fix to stop silent prod errors.

## Test plan
- [ ] Confirm local rspec for `Uploader.find_resources` passes (existing specs stub Typhoeus; should be unaffected)
- [ ] Manually verify in dev: navigate to Tarheel Reader category in focus-words, tap a book, see empty result instead of JSON parse error
- [ ] Opt a single test user into `tarheel_reader` flag and confirm the old (broken) behavior still reaches tarheelreader.org (we want to keep this path wired so we can re-enable it the moment there is an API to call)
- [ ] Check Sentry for any existing Tarheel-related exceptions stopping after deploy

## Context
- Research finding: [Building Wings acquisition announcement](https://www.buildingwings.com/blog/tar-heel-reader-joins-building-wings-as-monarch-reader/)
- Monarch Reader has a Directus backend that returns 403 on all public probes. No public API exists today.

### DIFF
--- a/lib/feature_flags.rb
+++ b/lib/feature_flags.rb
@@ -11,7 +11,8 @@ module FeatureFlags
               'auto_inflections', 'remote_modeling', 'focus_word_highlighting', 'profiles',
               'skin_tones', 'lessons', 'other_menu', 'shallow_clones', 'ai_board_generation',
               'ai_word_prediction', 'ai_board_suggestions', 'ai_symbol_search',
-              'ai_compliance_logging', 'supervisor_consent_flow']
+              'ai_compliance_logging', 'supervisor_consent_flow',
+              'tarheel_reader']
   ENABLED_FRONTEND_FEATURES = ['subscriptions', 'assessments', 'custom_sidebar', 'snapshots',
               'video_recording', 'goals', 'modeling', 'geo_sidebar', 'edit_before_copying',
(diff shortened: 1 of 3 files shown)


MelissaOneil (approved):
good to go!

PR #200, opened 2026-04-20
feat(ops): auto-notify Google Chat on key rotation

swahlquist:
## Summary
- `sync-render-env.js --apply` now posts to Google Chat (space: AI & Integrations) when it actually changes a secret value on any Render service. Message lists key names + affected envs only — never secret values.
- Removes the older post-apply Chat notification step from the workflow (the in-script one is more accurate and fires for both local `--apply` and the hourly GitHub Action).
- Unifies failure notifications onto the same webhook (`GOOGLE_CHAT_WEBHOOK_KEY_ROTATION`).
- New `docs/ROTATING_KEYS.md` gives senior devs a self-contained runbook for rotating the four shared MCP keys (GitHub PAT, Render, Notion, n8n) without co-founder involvement.

## Why
After finding four keys in plaintext inside `claude_desktop_config.json`, we need a lighter-weight signal than "ping everyone in Slack" when shared secrets rotate. People with local `.env` copies need to know to pull the new values and restart their tools. The hourly sync-to-Render pipeline is the natural trigger.

## Config required for this to fire
Repo secret `GOOGLE_CHAT_WEBHOOK_KEY_ROTATION` is already set. Value also stored in:
- `~/ai-company-brain/config/.env`
- 1Password Shared Dev → "Google Chat Webhooks" → `Key Rotation Notifier`

## Test plan
- [x] Manual webhook ping via `curl` — 200 OK, message posted.
- [x] `node -c scripts/sync-render-env.js` — syntax OK.
- [x] `yaml.safe_load` — workflow YAML valid.
- [ ] Merge → wait for next hourly cron (runs at :15 UTC) → verify workflow run succeeds with no notification (no changes expected).
- [ ] Rotate the Render API key (still exposed), run sync manually → verify notification posts to AI & Integrations space with `RENDER_API_KEY (dev, staging, prod)`.

### DIFF
--- a/.github/workflows/sync-render-secrets.yml
+++ b/.github/workflows/sync-render-secrets.yml
@@ -4,9 +4,9 @@ name: Sync Render Secrets from 1Password
 # Only runs if there are actual changes to apply.
 #
 # REQUIRED REPO SECRETS (Settings > Secrets and variables > Actions):
-#   OP_RENDER_SYNC_TOKEN  -- 1Password service account token (read-only ideally)
-#   RENDER_API_KEY        -- Render API key with write access
-#   GOOGLE_CHAT_WEBHOOK   -- Google Chat webhook URL for notifications
+#   OP_RENDER_SYNC_TOKEN              -- 1Password service account token (read-only ideally)
+#   RENDER_API_KEY                    -- Render API key with write access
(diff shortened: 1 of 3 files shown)


MelissaOneil (approved):
looks good

PR #208, opened 2026-04-22
fix(proxy): regenerate button_set on cache miss instead of silent 400

swahlquist:
## Summary

When the search proxy endpoint fetches a `button_set_cache` URL and S3 returns 403/404, the proxy used to return a 400. The frontend surfaces this as a permanent "Loading..." on board copy, and the DB's cache pointer stays stale indefinitely. This PR makes the fetch self-healing.

**Before:** any cache-pointer drift = user stuck forever until an admin runs a rails console command.

**After:** any cache-pointer drift = automatic regenerate + retry in the same HTTP request.

## Why this matters

Prior PRs #197, #202, #207 each closed one drift path between the DB's `remote_paths` entry and the actual S3 object. But ANY future drift (new race, new bug, new deploy state) re-creates the same user-visible failure, because the fetch side has no fallback. This PR fixes the architectural gap by adding that fallback.

## What it does

In `SearchController#proxy`, when `BadFileError` is raised on a URL matching `/button_set_cache/<global_id>/`:

1. Apply a 60-second Redis cooldown (prevents regen thrash when many tabs open the same broken board)
2. Delete the matching `remote_paths` entry so `generate_for`'s 12-hour cooldown does not return the same stale URL
3. Synchronously call `BoardDownstreamButtonSet.generate_for(board_id, user_id)`
4. If the result has a NEW URL, retry the fetch and stream back the result

Non-cache URLs still surface `BadFileError` as a 400 exactly as before.

## Test plan

- [x] `bundle exec rspec spec/controllers/api/search_controller_spec.rb -e proxy` — 8 examples, 0 failures (3 new specs: happy path, generate_for-fails path, non-cache URL is not affected)
- [ ] Deploy to staging, clear existing stale cache entries
- [ ] Board copy on vocal-flair-40 completes (was hanging forever)
- [ ] Board copy on vocal-flair-60 completes
- [ ] Check worker logs for no new `long-running job` errors related to button_set fetches

## Follow-up

PR #2 (separate branch) will apply the same regenerate-on-miss pattern to the PDF worker's per-button image fetch, which is the root cause of the 8-minute print board timeouts.

### DIFF
--- a/app/controllers/api/search_controller.rb
+++ b/app/controllers/api/search_controller.rb
@@ -225,7 +225,7 @@ class Api::SearchController < ApplicationController
       end
     rescue BadFileError => e
       error = e.message
-      s3_cache_miss = e.message.match?(/\b(403|404)\b/)
+      s3_cache_miss = e.message.match?(/(?:status |, )(403|404)\b/)
       Rails.logger.error("Proxy error for #{url}: #{error}")
     rescue => e
       error = "Failed to fetch URL: #{e.message}"
--- a/app/controllers/api/search_controller.rb
+++ b/app/controllers/api/search_controller.rb
@@ -225,18 +225,22 @@ class Api::SearchController < ApplicationController
       end
     rescue BadFileError => e
       error = e.message
+      s3_cache_miss = e.message.match?(/\b(403|404)\b/)
       Rails.logger.error("Proxy error for #{url}: #{error}")
     rescue => e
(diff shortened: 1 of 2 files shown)


MelissaOneil:
@copilot apply changes based on the comments in [this thread](https://github.com/lingolinq/LingoLinq-AAC/pull/208#pullrequestreview-4158757656)

MelissaOneil (approved):
good to go

PR #209, opened 2026-04-22
fix(pdf): harden OBF::Utils.save_image against empty data and Tempfile race

swahlquist:
## Summary

Full-board-set PDF prints (Vocal Flair 60, etc.) were hanging for 501 seconds (10 minutes) on staging 2026-04-20..22. The Render worker then killed the job. The browser surfaces this as an infinite loading state with no error. Worker logs showed hundreds of ImageMagick failures per print attempt:

```
convert-im6.q16: unable to open image '/tmp/image_stash20260422-116-tqlq9b.png':
    No such file or directory
convert-im6.q16: no images defined '/tmp/image_stash20260422-116-tqlq9b.png.jpg'
...
ERROR -- : long-running job, Progress . perform_action (), 501s
```

## Root causes

Two bugs inside `obf` 0.9.9.3's `OBF::Utils.save_image`:

**1. Empty response body still spawns convert.** When an S3 GET returns 200 with an empty body (stale cache pointer, CDN miss), the `elsif image['raw_data']` branch fires because `""` is truthy in Ruby. Zero bytes get written to a Tempfile. Convert is spawned on the empty file and fails. Each print job accumulates 100+ of these failures until Resque's `long-running job` guard kills it at 501s.

**2. Tempfile finalizer race.** `save_image` returns `{thread:, image:, type:, pid:}` with **no reference to the Tempfile it created**. During `threads.each{|t| t[:thread].join }` in `obf/pdf.rb#344`, Ruby GC can collect the Tempfile and fire its finalizer, which unlinks `/tmp/image_stash*.png` out from under the still-running convert subprocess. That produces the exact "No such file or directory" error. The race is worse after #197 set `MAGICK_THREAD_LIMIT=1`, which made each convert call slower and widened the GC window.

## Fix

Single initializer (`config/initializers/obf_save_image_hardening.rb`) that `prepend`s a hardened `save_image` onto `OBF::Utils`:

- **Reject too-small data** (< 100 bytes) before spawning convert. The PDF layer at `obf/pdf.rb#413` already handles `nil` returns gracefully and renders the button without a symbol.
- **Post-write file-size sanity check** in case data writes short for any other reason.
- **Include `tempfile: file` in the threadable-spawn return hash** so the Tempfile stays alive until the caller drops the reference (which is after the convert thread joins and the PDF image path is consumed).

## Net effect

Full-board-set print either completes normally or completes with missing-symbol placeholders where the S3 data is genuinely gone. No more 10-minute hangs.

## Test plan

- [x] `bundle exec rspec spec/initializers/obf_save_image_hardening_spec.rb` - 4 examples, 0 failures
- [x] `bundle exec rspec spec/lib/converters/` - 84 examples, 0 failures (no regression on existing PDF export paths)
- [ ] Deploy to staging
- [ ] Print full Vocal Flair 60 board set; verify PDF completes in < 3 minutes
- [ ] Watch staging worker logs for absence of "long-running job" errors
- [ ] If convert still OOMs on some SVGs, that is PR #197's 64MiB cap and is a separate tuning question (bump to 128MiB as follow-up)

## Related

Companion PR: #208 (proxy regenerate-on-miss) - fixes the same S3-drift class of bug for the board-copy path. Together they eliminate the two user-visible failure modes from missing or stale S3 cache objects.

### DIFF
--- a/config/initializers/obf_save_image_hardening.rb
+++ b/config/initializers/obf_save_image_hardening.rb
@@ -1,0 +1,119 @@
+require 'tempfile'
+
+# Hardens OBF::Utils.save_image against two failure modes observed in staging
+# 2026-04-20..22 on full-board-set PDF prints (Vocal Flair 60, etc.):
+#
+# 1. Empty or too-small response bodies: when a button image URL returns a 0-
+#    byte 200 (stale S3 cache entry, CDN miss, etc.), the original method still
+#    writes the empty bytes to a Tempfile and spawns `convert`, which fails
+#    with "unable to open image ... No such file or directory" and "no images
+#    defined ...jpg". We skip the image instead — the PDF layer handles nil
+#    gracefully ("missing image" log line, button renders without symbol).
+#
+# 2. Tempfile finalizer race: the original save_image returns
+#    `{thread:, image:, type:, pid:}` with NO reference to the Tempfile it
+#    created. When GC runs during `threads.each{|t| t[:thread].join }` in
+#    pdf.rb, the Tempfile finalizer unlinks /tmp/image_stash*.png out from
+#    under the still-running `convert` subprocess. We add `tempfile: file` to
+#    the returned hash so the Tempfile object stays alive until the caller
+#    drops the reference.
+#
+# Together these turn the observed 501s worker timeout into a job that
+# completes in normal time with any genuinely-missing symbols skipped.
+module OBFSaveImageHardening
+  MIN_IMAGE_BYTES = 100
+
+  def save_image(image, zipper = nil, background = nil)
+    if image['data']
+      image['content_type'] = image['data'].split(/;/)[0].split(/:/)[1] if !image['content_type']
+    elsif image['raw_data']
+      # already processed
+    elsif image['path'] && zipper
+      image['raw_data'] = zipper.read(image['path'])
+      if !image['content_type']
+        types = MIME::Types.type_for(image['path'])
+        image['content_type'] = types[0] && types[0].to_s
+      end
+    elsif image['url']
+      OBF::Utils.log "  retrieving #{image['url']}"
+      url_data = OBF::Utils.get_url(image['url'])
+      OBF::Utils.log "  done!"
+      image['raw_data'] = url_data['data']
+      image['content_type'] = url_data['content_type']
+    elsif image['symbol']
+      # not supported
+    end
+
+    if image['raw_data'] && image['raw_data'].to_s.bytesize < MIN_IMAGE_BYTES && !image['data']
+      OBF::Utils.log "  skipping image with too-small data (#{image['raw_data'].to_s.bytesize} bytes) url=#{image['url']}"
+      return nil
+    end
+
+    type = MIME::Types[image['content_type']]
+    type = type && type[0]
+    extension = nil
+    if type.respond_to?(:preferred_extension)
+      extension = type && ('.' + type.preferred_extension)
+    elsif type.respond_to?(:extensions)
+      extension = type && ('.' + type.extensions.first)
+    end
+    file = Tempfile.new(['image_stash', extension.to_s])
+    file.binmode
+    if image['data']
+      str = Base64.strict_decode64(image['data'].split(/\,/, 2)[1])
+      file.write str
+    elsif image['raw_data']
+      file.write image['raw_data']
+    else
+      file.close
+      return nil
+    end
+    file.close
+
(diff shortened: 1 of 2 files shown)


MelissaOneil:
@copilot apply changes based on the comments in [this thread](https://github.com/lingolinq/LingoLinq-AAC/pull/209#pullrequestreview-4158886221)

swahlquist:
## Review notes + retrospective (for team learning, not blocking)

@MelissaOneil and anyone reading this later: the last commit (`bb49bdff4`) is a small correctness fix on top of the Copilot agent's changes. Wanted to leave an honest walk-through of what was done, what I agreed with, what I'd have done differently, and what to watch for. This is for the retro, not to re-litigate anything.

### What the Copilot agent did well

**Full argv-form for `convert`** (`Process.spawn(*args)` / `system(*args)`). This is objectively better than my original shell-string interpolation and better than the narrower background-whitelist approach we'd agreed to in the earlier review. It eliminates the shell-injection surface entirely rather than shutting only one input. Given unconstrained scope I would have taken this path too; I only went narrower because the review plan we'd agreed on was "minimum viable." Worth noting for next time: if an argv refactor is small and local, prefer it over a string sanitizer.

### What I would have left alone

**JPEG fast-path change** (removed the `cp` and points `local_path = file.path`). Technically correct (prior version produced `.jpeg.jpeg`). But that code path is verbatim upstream `obf 0.9.9.3` behavior, it produced valid JPEG output that Prawn read correctly (Prawn reads by content, not extension), and it has no connection to the production bug this PR was opened to fix (501-second print timeouts from missing S3 images).

My preference would have been a separate two-line PR for that change. Reasons:
- Keeps the monkey-patch closer to upstream, which makes future obf gem upgrades easier
- Gives us a clean bisect target if any JPEG rendering regression surfaces later
- Smaller review surface for each PR

Not asking anyone to revert it. Just flagging that it widened the blast radius of this PR beyond the incident scope.

### What the Copilot agent got wrong

**Spec cleanup rescue syntax.** The agent's fix was:

```ruby
ensure
  res[:tempfile].close! rescue Errno::ENOENT if res && res[:tempfile]
end
```

Ruby's inline `rescue X` does NOT rescue only `X`. It rescues any `StandardError` and evaluates the expression to `X`. So this version silently swallows every exception during cleanup, which is the exact footgun the "fix" was supposed to address. Tests still passed because the happy path doesn't raise, but a real bug in `close!` would be invisible.

Commit `bb49bdff4` replaces it with a proper `begin / rescue Errno::ENOENT` block. Small change, but worth calling out because this is a Ruby idiom that bites people regularly.

### Things to watch for in staging/prod after this deploys

1. **Worker logs for `convert-im6.q16:` errors.** If we still see `cache resources exhausted`, that's PR #197's `MAGICK_MEMORY_LIMIT=64MiB` being too tight, not this PR. Tune separately.
2. **JPEG symbol rendering.** If any small square JPEG symbol renders incorrectly in a printed PDF, the JPEG fast-path change is suspect and should be bisected first.
3. **Long-running job errors.** If we see `ERROR -- : long-running job, Progress . perform_action (), 501s` again, the Tempfile pin (`:tempfile` in the returned hash) didn't fully solve the race and we may need to look at disabling Tempfile's ObjectSpace finalizer during PDF generation.

### Meta: autonomous agent working on the same branch

Heads-up that the Copilot agent (GitHub's `copilot-swe-agent[bot]`) pushed two commits to this PR branch autonomously while I was editing locally, one of them titled `Changes before error encountered`. Not a blocker, but worth knowing the pattern:

- Autonomous Copilot can and will push to the branch without warning
- It optimizes for "address every review comment" over "ship smallest safe change"
- It can introduce subtle bugs (see rescue-syntax above) because it's writing fast and not reading its own code carefully
- Its commit messages can be sloppy (`Changes before error encountered`)

Recommended practice: when both a human-driven agent (me) and Copilot's autonomous agent might be working on the same PR, either one should have a "lock" (a draft label, an in-progress indicator in the PR description) so we don't step on each other. Alternatively, disable Copilot autonomous edits on branches where a live collab is happening.

### Summary

Net, this PR is in a good state. Argv form is a real upgrade, the Tempfile pin addresses the staging race, and the empty-data guard addresses the cascading convert failures. The JPEG scope creep is worth a mental note for next time, and the rescue-syntax bug is now fixed.

Happy to discuss any of this in the next retro.

— Claude (working with Scot)

MelissaOneil (approved):
looks good

PR #222, opened 2026-04-27
compliance(coppa): retention enumeration + AiApiLog IP scrub schedule + school-official carve-out

swahlquist:
## Summary

Three quick-win fixes for the COPPA Final Rule (effective 2026-04-22, deadline missed). Tracks `docs/legal/COPPA_VERIFICATION_2026-04-26.md` punch-list rows **3a, 3b, 4b**. PR #211 was documentation only; this PR moves the needle on the running app.

- **3a Retention enumeration in the privacy notice.** Per Fenwick guidance, linking out from the privacy notice is insufficient under 16 CFR 312.10. Replaced the single generic "12 months inactive" sentence with per-category windows for active accounts, inactive accounts, AI request logs (AiApiLog), voice/audio recordings (ButtonSound), dwell/eye-gaze telemetry, school-tenant data on contract end, and children's data. Sourced from `docs/legal/DATA_RETENTION.md`.
- **3b AiApiLog IP scrub schedule.** `AiApiLog.redact_old_ip_addresses!` (`app/models/ai_api_log.rb:108-112`) is defined with a 90 day default but was never called. Wired it into the daily 6 AM UTC block in `lib/tasks/scheduler.rake` alongside `enforce_data_retention_policies`.
- **4b School-official as parental-consent substitute carve-out.** Old copy endorsed "parent, guardian, or authorized school official" as fully interchangeable. The 1999 FTC FAQ school-official carve-out only survives when the service has no AI, no profiling, no ads. LingoLinq has all three. Tightened copy to: school-official authorization substitutes for parental consent only for non-AI school-curriculum use; any AI feature use by a child under 13 requires verifiable parental consent under 16 CFR Part 312 regardless of school enrollment.

No feature flag (privacy notices must be live to users at update time; flag-gating defeats the disclosure obligation). No code path behavior change beyond the scheduler entry; no schema or migration impact.

## Files changed

- `app/frontend/app/templates/privacy.hbs` (Ember template, sections 5 and 7)
- `app/views/shared/_privacy.html.erb` (Rails partial, sections 5 and 7) - mirrored edits to keep templates in sync
- `lib/tasks/scheduler.rake` (added `redact_old_ai_api_log_ips` step in the daily 6 AM UTC branch)
- `spec/models/ai_api_log_spec.rb` (added 6 new examples covering `redact_old_ip_addresses!`)
- `spec/lib/tasks/scheduler_spec.rb` (new file) - asserts the redact step is invoked at hour 6 UTC and not at other hours

[...]

PR #226, opened 2026-04-28
test(logs_controller): pin event timestamps to fix flaky spec

swahlquist:
## Summary

Fixes the timestamp-drift + type-mismatch flake in \`spec/controllers/api/logs_controller_spec.rb:1118\` (\"should return events data if no encryption header sent\"). The test surfaced on the staging CI run for PR #222 (run 25029553351) but is unrelated to PR #222's contents.

## What was wrong

\`4.seconds.ago.to_i\` was evaluated twice -- once at fixture creation, once at expectation -- so on a slow runner the two calls fell into different seconds. The test also asserted Integer timestamps, but \`LogSession\` serializes them as Float, so strict \`eq\` would fail even without the drift.

## What changed

- Captured \`event1_ts\` and \`event2_ts\` once and reused them in both fixture and expectation
- Cast the expected timestamp to \`.to_f\` to match the model's serialization

## Hard rules check

- [x] Branch from \`origin/staging\`
- [x] No em dashes
- [x] No plaintext secrets
- [x] Test-only change, no behavior change

## Test plan

- [ ] CI runs the modified spec and it passes consistently across at least 3 runs
- [ ] No other specs in \`logs_controller_spec.rb\` regress

### DIFF
--- a/spec/controllers/api/logs_controller_spec.rb
+++ b/spec/controllers/api/logs_controller_spec.rb
@@ -1117,10 +1117,12 @@ describe Api::LogsController, :type => :controller do
 
     it 'should return events data if no encryption header sent' do
       token_user
+      event1_ts = 4.seconds.ago.to_i
+      event2_ts = 3.seconds.ago.to_i
       log = LogSession.process_new({
         :events => [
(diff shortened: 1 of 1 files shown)


MelissaOneil (approved):
good to go

PR #229, opened 2026-04-28
fix: resolve infinite loading hang in copy board workflow

swahlquist:
Summary of Changes:
1. Deadlock Prevention in Buttonset:
   - Added a 10-second safety timeout to the serialization tail (__loadButtonsSerialTail) in Buttonset.load_buttons. This ensures that if a promise fails to settle or hangs, the queue is force-cleared, preventing subsequent operations from stalling.
2. Robust Error Handling in UI:
   - Updated copy-board and copying-board components to ensure the 'loading' flag is consistently reset to false in rejection handlers.
   - Added isDestroyed and isDestroying lifecycle checks before calling .set() in asynchronous callbacks to prevent errors when components are unmounted during background tasks.
3. Visibility and Diagnostics:
   - Injected debug logging in copying-board.js to track the hierarchy load and editManager.copy_board promise lifecycle.

This resolves the issue where copying a full board set would hang indefinitely on 'loading' if a promise in the hierarchy load chain failed or stalled.


MelissaOneil:
@copilot apply changes based on the comments in [this thread](https://github.com/lingolinq/LingoLinq-AAC/pull/229#pullrequestreview-4201926516)

PR #225, opened 2026-04-27
compliance(coppa): scrub under-13 PII from Bugsnag and NewRelic

swahlquist:
## Summary

Two COPPA Final Rule quick wins (effective 2026-04-22) covering items **5a** and **5b** from the COPPA verification audit. Both are config / initializer scrubs only, no model changes, no feature flag.

Audit doc lives in PR #224 (parent doc PR, will land separately on staging). Stable link to the items:

https://github.com/lingolinq/LingoLinq-AAC/blob/compliance/coppa-final-rule-audit-2026-04-27/docs/legal/COPPA_VERIFICATION_2026-04-26.md

### Item 5a: Bugsnag

`config/initializers/bugsnag.rb` adds a `CoppaBugsnagScrub` module wired through `config.add_on_error`. When the affected user has `settings['coppa']['pending_parent_consent']` set and no `parent_consent_granted_at` (i.e. `User#coppa_parental_consent_pending?` is true at `app/models/user.rb:370`), the callback:

- Clears `event.user` (drops id, email, name)
- Redacts `clientIp`, `body`, `params`, `query_parameters`, `request_parameters`, `form_params`, and `referer` in the `request` meta tab
- Redacts `Cookie`, `Set-Cookie`, `Authorization`, `X-Forwarded-For` (and `Forwarded`, `X-Real-IP` variants) in the `headers` and `cookies` tabs
- Strips query strings from `url` and `path`, and replaces global_id segments with `[REDACTED_ID]`
- Redacts `context` if it contains an `@` or a global_id pattern

Error class, message, stack frames, app version, release stage, and severity all pass through unchanged. Adult users and anonymous reports are not modified.

### Item 5b: NewRelic

`config/newrelic.yml` adds an `attributes:exclude` block under `common: &default_settings`. The NewRelic Ruby agent cannot filter per-user from yaml, so the conservative read is to over-filter for every user. Over-filtering is safe; under-filtering risks the per-violation $53,088 penalty.

Excluded attributes:
- `request.parameters.*` (covers `/api/v1/words/predict` body and any other route parameter)
- `request.headers.cookie`, `authorization`, `referer`, `x-forwarded-for`, `x_forwarded_for`, `forwarded`, `x-real-ip`
- `http.request.headers.cookie`, `authorization`, `referer` (newer agent key namespace)
- `user.*`, `user_id`, `global_id`, `device_id`

The yaml anchor propagates the exclude block to `development`, `test`, `production`, and `staging`.

## Test plan

[...]

PR #246, opened 2026-05-04
fix(coppa): repair CoppaSentryScrub lookup so the COPPA branch actually runs

swahlquist:
## Summary
- The COPPA Sentry scrub introduced in the Sentry adoption PR is dead code in production. `set_sentry_user` writes `SHA-512(remote_ip)` into `Sentry.user.id` (correct, for issue grouping), but `CoppaSentryScrub#lookup_user` then runs `User.where(id: <128-char hex>).first`, which never resolves. `child_user?` returns false, `scrub!` never fires.
- Net effect: child PII has been shipping to Sentry on every event since the Sentry migration, despite the privacy promise documented in PR #225.

## Fix
- Stash `@api_user` in `RequestStore` from `set_sentry_user` (request-scoped, auto-cleared per request via the existing `request_store` gem middleware).
- `lookup_user` reads `RequestStore` first. Numeric-id fallback retained for background-job paths that explicitly use `Sentry.with_scope`.
- Issue grouping in Sentry UI is unchanged; `user.id` is still the IP hash.

## Also addresses (compliance-check WARN #1)
- New `config.before_breadcrumb` strips `access_token`, `token`, `password`, `reset_token`, `confirmation_token`, `tmp_token`, `api_key`, `secret_key`, `auth`, `bearer`, and `email` query params from breadcrumb URLs and message strings, for ALL users regardless of COPPA status. Closes the adult / unauthenticated breadcrumb leakage gap.

## Tests
- New regression case: hex-string `user.id` + child user reachable only via RequestStore -> full scrub fires.
- New negative case: hex-string id + empty RequestStore -> event left alone.
- 5 new `before_breadcrumb` cases.

## Audit reference
- `audit-reports/security-review-2026-05-04.md` finding #1 (CRITICAL ship-blocker)
- `audit-reports/compliance-check-2026-05-04.md` warning #1

## Test plan
- [ ] CI passes
- [ ] Manual: trigger a Sentry event in staging while logged in as a COPPA-pending test account; confirm event payload in Sentry UI shows `[REDACTED]` for user, request body, and ip.
- [ ] Manual: trigger a Sentry event with `?access_token=foo` in the URL; confirm breadcrumb shows `access_token=[REDACTED]`.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -44,11 +44,14 @@ class ApplicationController < ActionController::Base
   
   # Hash the requester IP and attach to the Sentry scope as the synthetic
   # user id so issues group per-requester without exposing raw IPs.
-  # CoppaSentryScrub redacts the user entirely if a logged-in user turns
-  # out to be COPPA-pending.
+  # CoppaSentryScrub redacts the event entirely if a logged-in user turns
+  # out to be COPPA-pending. The actual User reference for that consent
+  # check rides on RequestStore (NOT on the Sentry event itself), because
+  # Sentry.user_hash[:id] is the IP hash and won't resolve to a database id.
   def set_sentry_user
     return unless defined?(Sentry) && Sentry.respond_to?(:initialized?) && Sentry.initialized?
     Sentry.set_user(id: GoSecure.sha512(request.remote_ip, 'user_ip'))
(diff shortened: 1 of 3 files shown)


MelissaOneil (approved):
gtg when checks pass

PR #247, opened 2026-05-04
fix(secrets): un-collapse SECRET_KEY_BASE / COOKIE_KEY encryption domains

swahlquist:
## Summary
Replaces \`config.secret_key_base = ENV["SECRET_KEY_BASE"] || ENV["COOKIE_KEY"]\` with a strict \`ENV.fetch\` so production fails loud at boot if the env var is missing, instead of silently swapping in a different encryption domain.

## Why this matters
\`SECRET_KEY_BASE\` is the root for Rails signed cookies, encrypted credentials, and \`MessageVerifier\` artifacts. \`COOKIE_KEY\` is a legacy env var with no other consumers in the codebase (grep confirms only the two .rb sites and the CI workflow). The OR-fallback collapsed two key identifiers into one, making rotation procedures ambiguous and invalidation modes hard to reason about.

Render prod sets \`SECRET_KEY_BASE\` via \`generateValue: true\`, so the fallback was never actually reached in prod. The audit flagged the template itself, not a live outage.

## Files
- \`config/environments/production.rb\`: \`ENV.fetch("SECRET_KEY_BASE")\`. Boots loudly.
- \`config/initializers/secret_token.rb\`: Same \`ENV.fetch\` form with a non-prod-only fallback to \`COOKIE_KEY\` (CI compatibility) then \`SecureRandom.hex(64)\` for local dev. Prod still raises.

## What's NOT in this PR
- \`render.yaml\` is untouched (changing it can trigger redeploys with new env values; safer to remove \`COOKIE_KEY\` once CI is updated).
- \`.github/workflows/ci.yml\` is untouched. Adding \`SECRET_KEY_BASE: \${{ secrets.SECRET_KEY_BASE }}\` to CI requires the secret to be added in GitHub repo settings first.

## Audit reference
- \`audit-reports/infra-audit-2026-05-04.md\` finding NEW HIGH (Section 2a)

## Test plan
- [ ] CI passes (the secret_token.rb fallback chain still covers \`COOKIE_KEY\` in CI today).
- [ ] On staging deploy: confirm app boots; \`SECRET_KEY_BASE\` is already set in Render.
- [ ] After staging soak, follow-up PR adds \`SECRET_KEY_BASE\` to GitHub Actions secrets and removes the \`COOKIE_KEY\` fallback in secret_token.rb.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -10,7 +10,11 @@ Rails.application.configure do
   # Rake tasks automatically ignore this option for performance.
   config.eager_load = true
 
-  config.secret_key_base = ENV["SECRET_KEY_BASE"] || ENV["COOKIE_KEY"]
+  # Fail loud at boot if SECRET_KEY_BASE is missing. The previous form
+  # (`ENV["SECRET_KEY_BASE"] || ENV["COOKIE_KEY"]`) silently collapsed two
+  # encryption domains and made rotation procedures ambiguous.
+  # Render production sets SECRET_KEY_BASE via `generateValue: true`.
+  config.secret_key_base = ENV.fetch("SECRET_KEY_BASE")
   
   # Full error reports are disabled and caching is turned on.
   config.consider_all_requests_local       = false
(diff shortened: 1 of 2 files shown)


MelissaOneil (approved):
gtg when checks pass

PR #248, opened 2026-05-04
fix(progress): scope idempotency and view permission to the initiating user

swahlquist:
## Summary
Closes the cross-user pooling and view-permission gaps in `Progress` flagged as HIGH by `audit-reports/security-review-2026-05-04.md` finding #2.

Today: `Progress` exposes \`add_permissions('view', ['*']) { true }\` and pools work purely on \`(class, id, method, args)\`. If two users happen to fire the same operation with identical args, the second one is handed back the first one's progress object and can later see its result hash. For \`generate_download\` that result hash contains a presigned S3 \`download_url\`. Existing safety relies on convention (most callers thread \`user_global_id\` into args); the audit's concern is the next caller that omits it.

## What changed
- \`Progress.schedule\` now accepts a \`for_user:\` keyword. The initiating user's global_id is recorded in \`settings['for_user_global_id']\` and participates in the idempotency match alongside class/id/method/args.
- \`'view'\` permission rule rewritten: owner-only when \`for_user_global_id\` is recorded; permissive when nil (backward-compat for legacy / admin / background progresses).
- All security-sensitive call sites updated to pass \`for_user: @api_user\`. Admin/system call sites (global stats, admin trends, anonymous logs, webhook test, background sound suggestions) deliberately left as nil-owner.

## Tests added (\`spec/models/progress_idempotency_spec.rb\`)
- Cross-user pooling is blocked.
- Same-user pooling still works.
- Two nil-owner progresses still pool (legacy / system path).
- nil-owner and for_user-scoped progresses do not pool.
- View permission: owner allowed, non-owner denied, legacy permissive.

## Files
- \`app/models/progress.rb\`
- \`app/controllers/api/{boards,logs,users,sounds,button_sets,purchasing}_controller.rb\`
- \`spec/models/progress_idempotency_spec.rb\`

## What's NOT in this PR
- The remaining benign call sites (admin/background) keep the nil-owner shape; in-flight records also keep it. Once all in-flight records age out (4-hour window), a follow-up PR can remove the legacy \`owner.nil? -> true\` branch and require ownership unconditionally.

## Audit reference
- \`audit-reports/security-review-2026-05-04.md\` finding #2 (HIGH)

## Test plan
- [ ] CI passes.
- [ ] Manual on staging: User A scheduling \`generate_download\` while User B simultaneously schedules the same board+type yields two distinct progress IDs.
- [ ] Manual on staging: User B polling User A's progress ID returns 401 / not found.

### DIFF
--- a/app/controllers/api/boards_controller.rb
+++ b/app/controllers/api/boards_controller.rb
@@ -286,7 +286,7 @@ class Api::BoardsController < ApplicationController
       if !params['q'].blank? && !params['public']
         limited_boards = boards
         if params['allow_job'] #&& boards.limit(26).select('id').length > 25
-          progress = Progress.schedule(Board, :long_query, params['q'], params['locale'], boards.select('id, board_content_id').map(&:global_id) + (other_searchable_board_ids || []))
+          progress = Progress.schedule(Board, :long_query, params['q'], params['locale'], boards.select('id, board_content_id').map(&:global_id) + (other_searchable_board_ids || []), for_user: @api_user)
           boards = []
         else
           # For private user searches this will limit to the user's first 25 boards
@@ -650,7 +650,7 @@ class Api::BoardsController < ApplicationController
     board = Board.find_by_path(params['board_id'])
     return unless exists?(board, params['board_id'])
     return unless allowed?(board, 'edit')
-    progress = Progress.schedule(board, :slice_locales, params['locales'], params['ids_to_update'], (@api_user && @api_user.global_id))
+    progress = Progress.schedule(board, :slice_locales, params['locales'], params['ids_to_update'], (@api_user && @api_user.global_id), for_user: @api_user)
     render json: JsonApi::Progress.as_json(progress, :wrapper => true).to_json
   end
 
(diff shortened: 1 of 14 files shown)


swahlquist:
Thanks for the review @copilot-pull-request-reviewer. Both findings were correct and have been pushed in 8627ae189:

## 1. Anonymous confirmation-code subscribe (production bug)
You called this exactly right. \`for_user: @api_user || user\` collapsed to \`for_user: user\` for the anonymous flow, scoping the progress to the target user, which caused \`Api::ProgressController#progress\` to 401 anonymous polls.

Fix: changed to \`for_user: @api_user\` on all four \`users_controller\` sites (subscribe gift-code, subscribe token, unsubscribe, verify_receipt). When \`@api_user\` is nil (anonymous confirmation flow), the progress falls back to legacy permissive view, gated only by the \`protect_global_id\` nonce on \`Progress\`. When \`@api_user\` is set, owner scoping applies as designed.

The anon-confirm spec at \`users_controller_spec.rb:2016\` was tightened to explicitly assert \`for_user: nil\`, so a future change cannot silently re-break this path.

## 2. Existing-spec breakage
Confirmed: 11 of the 13 CI failures were exactly this. Updated the call-signature \`.with(...)\` matchers and the exact \`settings.eq({...})\` hashes in:
- \`boards_controller_spec.rb\` (import)
- \`purchasing_controller_spec.rb\` x3
- \`sounds_controller_spec.rb\` (import)
- \`users_controller_spec.rb\` x4 (subscribe variants, unsubscribe, reset_eval settings hash)
- \`progress_spec.rb\` (settings hash now includes \`for_user_global_id => nil\`)
- \`throttling_spec.rb\` (purchase_gift)

## Pre-existing flake, not addressed
\`spec/models/concerns/relinking_spec.rb:1519\` ('should wrap them all together if created within 30 seconds of each other'): \`assert_copy_id\` picked \`bs[1].global_id\` instead of \`bs[0].global_id\`. The test creates 3 boards in a tight loop with no explicit timestamp ordering, so the failure mode looks order-dependent on creation timing. None of this PR's changes touch that code path. Tracking as a separate flake.

CI re-running now.

MelissaOneil (approved):
gtg

PR #257, opened 2026-05-09
fix(copy): race buttonset hierarchy load against live-links fallback

swahlquist:
## Summary

The Copy Board modal currently waits up to **60 seconds** for a stale or missing buttonset before falling back to the live-links walk that PR #255 added. On staging, that's exactly what happens after a 97-board bulk copy: the freshly-copied root has no buttonset yet (the `:slow` Resque queue from PR #230 hasn't drained the deferred `update_for` jobs), and the user sees the full master-timeout from PR #237 before the fallback runs.

This PR races the two paths instead of cascading them.

## What changed

- New internal helper `loadHierarchyForCopyModal(board, opts)` in `app/frontend/app/controllers/copying-board.js`:
  - Starts the buttonset hierarchy load immediately.
  - Schedules a delayed live-links walk after `EARLY_LIVE_LINKS_DELAY_MS` (default **6000ms**) — if buttonset hasn't returned by then, kick off live-links in parallel and accept whichever resolves a usable hierarchy first.
  - On buttonset **rejection** (timeout / `generation_stalled` / etc.), runs live-links immediately rather than waiting for the early-fire timer. The previous behavior is preserved.
  - Cancels the early-fire timer once one path settles, so we don't leak runloop work.
- `opening()` refactored to use the helper. `isDestroyed`/`isDestroying` checks, error-flag preservation, and the `hierarchyRootOnlyWarning` logic are unchanged.

## Why this is the right surface to fix

The race is at the modal-opening layer, not in `BoardHierarchy` itself, because both `load_with_button_set` and `load_from_live_links` are useful primitives on their own. Putting the orchestration in the controller keeps the utility module narrow and avoids surprising any other callers.

## Tradeoff

Live-links produces a flatter hierarchy than the buttonset (root + immediate children, not the full graph). When live-links wins the race we still set `hierarchyRootOnlyWarning=true` so the user is informed. Existing fallback path already did this.

## Worst-case behavior

If buttonset is healthy and fast (<6s), live-links never runs. No extra work.

If buttonset is hanging, live-links kicks off at 6s and wins. User sees the modal in ~6s instead of ~60s.

If both fail, the existing error UI (`hierarchyLoadFailed=true` + `isTimeoutError=true`) still surfaces.

## Test plan

- [x] New unit test: live-links wins after early-fire (5 assertions)
- [x] New unit test: buttonset wins fast, live-links never invoked (4 assertions)
- [x] New unit test: both paths fail → timeout error UI (4 assertions)
- [ ] Staging soak: trigger bulk copy of `vocal-flair-112` (~97 boards) on a fresh source whose buttonset is stale, confirm modal opens in ~6s
- [ ] Verify existing test "users can select live linked boards when hierarchy loading fails" still passes (live-links is still triggered immediately on buttonset rejection)

## Companion PR

Branch B (`chore/staging-slow-queue-capacity`) addresses the underlying queue-drain time on staging. This PR is the user-visible fix; that one is the infrastructure tuning.


MelissaOneil (approved):
ready to merge if checks pass

PR #283, opened 2026-05-23
Omer/fix/speak mode pin

MelissaOneil:
Speak Mode PIN Fix
Title: fix: speak-mode-pin modal — fix digit input and add submit button
Description:
The speak mode PIN modal was broken in two ways: number buttons used on="select" which does not fire on click, and there was no submit button so users had no way to confirm their PIN.
Changes:

Removed on="select" from all digit buttons in speak-mode-pin.hbs so tapping numbers correctly enters digits
Added a backspace button so users can correct mistakes
Added a Submit button wired to a new submit_pin action
Added submit_pin and delete_digit actions to both speak-mode-pin.js component and controller
Users can now enter their PIN and exit speak mode to either edit mode or home depending on context

How to test:

Go to preferences → set require pin to exit speak mode → set a PIN → save
Enter speak mode
Try to exit — PIN modal should appear
Tap number buttons — dots should appear in the field
Tap Submit — if PIN is correct, speak mode exits
If PIN is wrong, error message appears and field clears

Copy status: N/A — no new copy


MelissaOneil:
Re: “Reveal PIN” in the Speak Mode PIN modal

Yes — the “Reveal PIN” link is intentional. It exists for the case where someone is locked in Speak Mode and has forgotten the PIN. Tapping the link shows the PIN in the modal so they can enter it and exit.

This can be turned off per user in Settings → Preferences: the “Don’t show hint for discovering forgotten PIN” checkbox (hide_pin_hint). When that preference is enabled, the reveal link is not shown in the PIN modal. The settings UI also warns that hiding the hint means you won’t be able to exit Speak Mode on that device unless you remember the PIN or look it up elsewhere. 
#286

Issue #286, opened 2026-05-24
Adversary findings + Sentry alerts remediation (post-meeting 2026-05-22)


swahlquist:
## Summary

Consolidated tracking issue for adversary findings + Sentry alerts surfaced between 2026-05-18 and 2026-05-24. Driven by the 2026-05-22 DevOps meeting and a post-meeting adversary review of PRs #280-#284, a parallel review of PR #174, and Sentry alerts in the same window.

**Governance posture (v3 plan):** Soft freeze. Critical + High findings block merges of PRs that touch the SAME files or features. Other parallel work continues. Phase 1 dual-reviewer (/review-pr + /adversary-review) is mandatory for any PR touching security, AI generation, user data flows, feature flags, mailers, or the #tree/bulk/global_id/board/board_content/SlowWorker paths. Phase 2 expansion to all PRs after we have run Phase 1 on ~5 PRs.

**Companion artifacts:**
- MVP punchlist: `~/ai-company-brain/outputs/docs/2026-05-24-lingolinq-mvp-punchlist.md` (in-repo mirror at `docs/ops/MVP_PUNCHLIST_2026-05-22.md` pending `fix/scot-slowworker-cache-leak` landing)
- Plan: `~/.claude/plans/c-users-scotw-downloads-devops-sync-mee-lucky-mist.md` (v3)

## Section 1: Adversary findings from post-meeting merges (PRs #280-#284)

Each row: severity / finding / freeze scope / owner / status.

- [x] **Critical** - PIN modal reveal link prints `actual_pin` plaintext after user taps "touch here to show the PIN". `app/frontend/app/templates/components/speak-mode-pin.hbs:58-60` (PR #283 branch). **Freeze:** PR #283 PIN-touching half. **Owner:** Melissa or Omer.
- [ ] **Critical** - Customize Menu (PR #284) ships enabled-for-all with NO feature flag. Violates CLAUDE.md hard rule because AAC users are disruption-sensitive. `lib/feature_flags.rb:3-27` has no entry; `app/frontend/app/templates/user/board-detail.hbs:2878-2945` + `controllers/user/board-detail.js:6186-6202` use it unconditionally. **Freeze:** future PRs touching Customize Menu surface or speak-mode dropdown until flagged. **Owner:** Traci.
[...]

PR #287, opened 2026-05-24
fix: clear request-scoped caches on slow-queue jobs

swahlquist:
## Problem

PR #174 added request/job-scoped `Thread.current` caches (`board_content_cache`, `word_inflection_cache`, later `bulk_copy_in_progress`) and cleared them in `Worker.perform` and `ApplicationController`. The clearing never ran for **slow-queue jobs**:

- Slow jobs are enqueued under `SlowWorker` (`boy_band.rb:82`).
- At execution, Resque calls `SlowWorker.perform`, which delegates to `Worker.perform_at(:slow, ...)`.
- `Worker.perform_at` (`boy_band.rb:134`) runs the job via `klass.send(...)` directly, so the overridden `Worker.perform` (which held the cache-clearing `ensure`) is never on the stack.

The slow queue is where the heavy work runs: board copies, OBZ/PDF exports, and `BoardDownstreamButtonSet.update_for` (scheduled `:slow`, runs frequently). On a long-lived Resque worker, `board_content_cache` accumulates full board-content blobs across every slow job for the life of the process. That is an unbounded `Thread.current` memory leak on the worker, the opposite of #174's goal, and a credible contributor to worker OOM / restart-and-retry churn / timeouts.

Separately, `ApplicationController#with_request_caching` cleared only two of the three keys, so `bulk_copy_in_progress` was also leaking across web requests on the same Puma thread.

## Fix

Centralize the key list into a single source of truth, `Worker.clear_request_thread_caches`, called from all three sites (`Worker.perform`, `SlowWorker.perform`, `ApplicationController`). The root cause of this bug was the duplicated key list drifting apart; centralizing prevents recurrence (a new cache key is added in one place).

- `lib/worker.rb`: extract `Worker.clear_request_thread_caches`; `Worker.perform` calls it.
- `lib/slow_worker.rb`: add `ensure` calling `Worker.clear_request_thread_caches` (closes the slow-path leak).
- `app/controllers/application_controller.rb`: `with_request_caching` calls the shared method (also fixes the `bulk_copy_in_progress` web-request leak).
- `Worker.clear_request_thread_caches` also resets the **PiiScrubber blocklist** (`PiiScrubber.reset_blocklist!`). See the PII / FERPA note below.

### PII / FERPA scope note

This PR also changes PII-redaction lifecycle, which is broader than the title implies, so flagging it explicitly. The `PiiScrubber` per-user blocklist lives in `Thread.current` and is configured by `AiBoardGenerator` / `AiWordPredictor` before an AI call. `AiApiLog#scrub_summary_columns` (a `before_validation` hook) invokes `PiiScrubber.redact_for_ai` without re-configuring the blocklist. So on a shared Resque/Puma thread, a leftover blocklist from User A's job would scrub User B's `AiApiLog` against User A's names, leaving User B's own PII unredacted. Clearing the blocklist in the shared teardown closes that leak. It is a distinct, and more severe, instance of the same `Thread.current` leak class as the memory bug, and it is FERPA/HIPAA-relevant. Reviewers touching AI generation or `AiApiLog` should know redaction state is now reset on every request and job teardown.

## Tests

`spec/lib/worker_spec.rb` gains a `clear_request_thread_caches` block with four tests: direct method call, normal job via `Worker.perform`, slow job via `SlowWorker.perform`, and an end-to-end `schedule_for('slow')` then `process_queues` path that exercises the real enqueue/pop/`SlowWorker.perform` flow a production OOM leak would travel. Each asserts the board-content, word-inflection, and bulk-copy caches plus the PiiScrubber blocklist are cleared. The slow-path tests fail on the pre-fix code and pass after.

> Not run in CI-equivalent locally due to a missing `SECURE_ENCRYPTION_KEY` in the dev shell; syntax verified with `ruby -c` on all four files. Will run in CI.

## Scope

No user-facing behavior change, so no feature flag. Backend-only. Related to but does not revert PR #174.


swahlquist:
# Dual-Reviewer Pass on PR #287

**Reviewer:** Phase 1 dual-reviewer (`/review-pr` senior-dev pass + `/adversary-review` red-team pass)
**Diff base:** `staging`...`fix/scot-slowworker-cache-leak` (4 files, +59/-3)
**Verdict: REQUEST CHANGES (Block)**

Single blocking reason: the centralized `Worker.clear_request_thread_caches` method introduced by this PR misses two `Thread.current` keys that share the exact same lifecycle problem the fix is targeting, and one of them is the PiiScrubber blocklist. This is the cheapest possible time to close that gap.

The PR is otherwise tight, well-documented, and the centralization pattern is correct. Once the two PiiScrubber keys are added (or the existing `PiiScrubber.reset_blocklist!` is invoked) and one regression spec covers them, this becomes a clean ship.

---

## Critical (must fix before merge)

### PiiScrubber blocklist leaks across users on the same thread

**File:** `lib/worker.rb:20-24` (the new `clear_request_thread_caches` method)

**What:** `PiiScrubber` stores a per-user name blocklist in `Thread.current[:pii_scrubber_blocklist]` and `Thread.current[:pii_scrubber_blocklist_pattern]` (`lib/pii_scrubber.rb:241-242`). The PR's new clear method does NOT clear them. The entire motivation of this PR (Thread.current state surviving across slow-queue jobs on long-lived workers) applies identically to those keys.

**Trigger:**
1. `AiBoardGenerator` for User A runs on the slow queue and calls `PiiScrubber.configure_blocklist(names_for_A)` at `lib/ai_board_generator.rb:50`.
2. The same Resque worker thread later runs an `AiApiLog` save for User B.
3. `AiApiLog#scrub_summary_columns` (a `before_validation` hook at `app/models/ai_api_log.rb:20-24`) calls `PiiScrubber.redact_for_ai` WITHOUT calling `configure_blocklist` first.
4. The scrubber uses User A's stale blocklist. User B's family/student names are NOT redacted from the persisted `request_summary` / `response_summary`.

`AiWordPredictor` has the same `configure_blocklist` pattern at `lib/ai_word_predictor.rb:49` and runs on the same queue lifecycle.

**Impact:** FERPA / HIPAA / COPPA-relevant. `AiApiLog` is the privacy chokepoint per the existing audit; persisting user records with someone else's blocklist applied defeats the redaction guarantee. The fact that AiBoardGenerator and AiWordPredictor are slow-queue jobs is exactly the path this PR fixes for other caches but skips for the more sensitive one.

**Recommended fix:** `PiiScrubber.reset_blocklist!` already exists at `lib/pii_scrubber.rb:253-256`. Call it from `clear_request_thread_caches`:

```ruby
def self.clear_request_thread_caches
  Thread.current[:board_content_cache] = nil
  Thread.current[:word_inflection_cache] = nil
  Thread.current[:bulk_copy_in_progress] = nil
  PiiScrubber.reset_blocklist! if defined?(PiiScrubber)
end
```

Add a spec asserting `PiiScrubber.blocklist == []` after `Worker.perform`, `SlowWorker.perform`, and the controller `with_request_caching` ensure.

---

## High (fix before merge unless explicitly accepted)

### Spec does not exercise the real Resque job path

**File:** `spec/lib/worker_spec.rb:115-136`

**What:** The new specs call `Worker.perform('User', 'bacon')` and `SlowWorker.perform('User', 'bacon')` with `expect(User).to receive(:bacon)`. Production Resque calls these with chain-tracking arg suffixes appended by `Worker.schedule_for` / `boy_band.rb`. The specs bypass `Resque.enqueue` / `Resque.pop` entirely. The fix is correct, but the regression test only proves the ensure block runs on directly-invoked methods, not on a real enqueued slow job traveling through the full boy_band path.

**Trigger:** A future change to chain-arg stripping in `boy_band.perform_at` that raises BEFORE the inner `klass.send` runs. The slow-path `ensure` would still fire (good), but the spec wouldn't catch a regression where someone moved the ensure into a rescue branch.

**Impact:** False confidence. The PR body flags "specs not run locally" due to missing `SECURE_ENCRYPTION_KEY`; even when CI runs them, they don't exercise the path that motivated the fix.

**Recommended fix:** Mirror the pattern at `spec/lib/worker_spec.rb:19-25, 28-32`, which already enqueues and processes through the real queue path. One additional spec that enqueues via `Worker.schedule_for(:slow, ...)` and processes via `Worker.process_queues` would catch the regression class this fix is closing.

---

## Medium

### "Specs not run locally" caveat is a real verification gap

**File:** PR body, "Tests" section

**What:** The PR documents that specs were verified with `ruby -c` only because `SECURE_ENCRYPTION_KEY` is missing in the dev shell. The new specs touch `Thread.current` global state. If `worker_spec.rb` errors in CI before reaching the new describe block (e.g. `Resque.redis.del` at `boy_band.rb:147` raises without a stub), the regression tests silently never run.

**Recommended fix:** Block merge until CI shows the three new `describe "clear_request_thread_caches"` specs as `passed` (not skipped or errored). Look for them by name in the run log.

---

## Low / nits

### `bulk_copy_in_progress` clear vs `BoardSetCopier`'s own ensure

**File:** `app/services/board_set_copier.rb:26-27, 98, 105-106, 156` and `lib/worker.rb:23`

`BoardSetCopier` saves `prior_bulk_copy = Thread.current[:bulk_copy_in_progress]` and restores it in its own ensure. The worker-level clear fires AFTER that ensure per Ruby semantics, so on the happy path it's redundant but harmless. Latent risk only if a future code path calls `Worker.perform` synchronously from inside another in-flight job. Worth a `# Do not call mid-execution; teardown only` comment on `clear_request_thread_caches`.

### Trailing-newline missing on `lib/slow_worker.rb`

`\ No newline at end of file` shows in the diff. Cosmetic, but POSIX-correct.

---

## What was done well

- The "single source of truth" framing in the method comment is the right durable lesson from how PR #174's bug originated (duplicated key list drifting apart). Future cache keys land in one place by construction.
- The `ensure` placement in `SlowWorker.perform` is exactly right: it fires whether `Worker.perform_at` returns normally or raises.
- The PR body precisely names the production failure mode (board copies, OBZ/PDF exports, `BoardDownstreamButtonSet.update_for`), which lets the reviewer reason about scope without re-deriving it.
- Web path (`ApplicationController#with_request_caching`) confirmed correct: no `skip_around_action :with_request_caching` exists in any controller; the `around_action` ensure fires on exceptions and on normal returns.
- Other `Worker.perform_at` call sites checked: only `SlowWorker.perform` (this fix) and `boy_band.rb` test-only `process_queues`. No production bypass paths remain.

---

## Coverage and conventions pass

- **i18n:** N/A. Backend-only, no user-facing strings.
- **Feature flag:** N/A per PR body ("no user-facing behavior change"). Confirmed correct; this is internal cache hygiene.
- **Branch naming:** `fix/scot-slowworker-cache-leak`. Matches `<type>/scot-<kebab>` per CLAUDE.md.
- **Em dashes:** None in commit message or PR body destined for external publication.
- **Em dash hook noise:** N/A.
- **Branch base:** Branched from `staging`, targets `staging`. Correct.
- **CLAUDE.md hard rules:** All respected. No AI call sites touched, no secure_serialize fields, no AuditEvent gap.

---

## Summary

Land the PiiScrubber fix (one line + one spec) and the real-enqueue spec, re-run CI, and this ships. The architecture is right; the gap is one Thread.current key that wasn't in the original PR #174 inventory because PiiScrubber's blocklist API arrived later. Adding it here closes the same class of bug for the more privacy-sensitive surface, which is the exact PR to do it in.

swahlquist:
## Critical addressed in `a8818d48d`

`Worker.clear_request_thread_caches` now also calls `PiiScrubber.reset_blocklist!`. The teardown-only contract is documented on the method so nested callers do not accidentally invoke it mid-job.

**Specs:** the existing three `describe "clear_request_thread_caches"` cases each now configure a blocklist via `PiiScrubber.configure_blocklist(['Alice'])` (etc.) inside the cache-population step, then assert `PiiScrubber.blocklist == []` after the teardown path runs. Coverage spans:
- Direct invocation of `Worker.clear_request_thread_caches`
- Full `Worker.perform` ensure path
- Full `SlowWorker.perform` ensure path

Spec uses the public PiiScrubber API (`configure_blocklist` + `blocklist`) so the assertion does not couple to the underlying Thread.current keys.

## What is NOT in this commit

**[High] Spec does not exercise the real Resque job path.** Still mocks `User.bacon` and bypasses `Resque.enqueue` / `Worker.process_queues`. Left for a follow-up decision rather than rolled in here. The pattern at `spec/lib/worker_spec.rb:19-25` is the right shape if you want me to add it.

**[Medium] Specs-not-run-locally caveat.** Same caveat as the PR body; the new lines verify with `ruby -c`. Will need CI to confirm the three describe blocks resolve as `passed`, not `errored` or `skipped`.

**[Low] `bulk_copy_in_progress` mid-execution-clear note.** The new method comment now says "Teardown-only; do not call mid-job because nested callers may still need the state."

**[Low] Trailing newline on `slow_worker.rb`.** Not addressed in this commit.

Verdict shifts from REQUEST CHANGES to APPROVE WITH COMMENTS once CI confirms the three new specs pass.

swahlquist:
## High addressed in `3b9ae5b9e`

Added an end-to-end `it` block to the `clear_request_thread_caches` describe that schedules via `Worker.schedule_for('slow', User, :bacon)` and processes via `Worker.process_queues`. Mirrors the existing enqueue+process pattern at `spec/lib/worker_spec.rb:19-25`.

The new spec:
1. Schedules a job onto the slow queue.
2. Stubs `User.bacon` to populate `Thread.current[:board_content_cache]` and call `PiiScrubber.configure_blocklist(['Eve'])`.
3. Calls `Worker.process_queues`, which pops the job and invokes `SlowWorker.perform` with the chain-tracking arg suffix appended by boy_band.
4. Asserts the cache and PiiScrubber blocklist are cleared post-teardown.

This catches regressions in the actual production path: any change to boy_band's chain-arg handling in `perform_at`, or any reordering of the `SlowWorker.perform` ensure into a rescue branch, surfaces here rather than in production.

## Verdict shift

- Critical (PiiScrubber leak): closed in `a8818d48d`
- High (real-enqueue spec): closed in `3b9ae5b9e`
- Medium (specs-not-run-locally): still pending CI confirmation
- Low x2: nits, fine to defer or roll into a follow-up

Once CI confirms all four `describe "clear_request_thread_caches"` specs pass, verdict moves to APPROVE.

MelissaOneil (approved):

swahlquist (commented):
## Independent confirmation pass (concur with approval, one gate remaining)

Re-reviewed the full branch (3 commits) against the earlier dual-reviewer block and Melissa's approval. The two blocking findings from the dual-reviewer pass are resolved, and I independently verified a detail the prior pass left implicit. Concurring with the approval; the only open item is CI.

### Prior block findings: resolved
- **Critical (PiiScrubber blocklist not cleared)** resolved in `a8818d48d`. I verified the FERPA path end to end: `AiApiLog#scrub_summary_columns` (a `before_validation` hook) calls `PiiScrubber.redact_for_ai`, which reads the `Thread.current` blocklist via `scan_blocklist` and never re-configures it. A leftover per-user blocklist on a shared Resque/Puma thread would scrub User B's `AiApiLog` against User A's names and leave User B's own PII unredacted. Resetting in the shared teardown is the correct close.
- **High (spec did not hit the real Resque path)** resolved in `3b9ae5b9e`. New detail to record: `Worker.process_queues` (boy_band.rb:334) drains every queue and routes slow jobs explicitly through `SlowWorker.perform` (boy_band.rb:344), so the `schedule_for('slow')` then `process_queues` test exercises the exact enqueue/pop/perform path a production OOM leak travels. This is the strongest of the four specs.
- **Low (trailing newline on `lib/slow_worker.rb`)** fixed in a follow-up commit.
- **Low (nested-job teardown)** the "teardown-only; do not call mid-job" comment is present on `clear_request_thread_caches`. Pre-existing, acceptable.

### Open gate (the prior Medium)
- CI must show the four `describe "clear_request_thread_caches"` specs as passed, not skipped or errored. The local run was blocked by a missing `SECURE_ENCRYPTION_KEY` in the dev shell, so the specs are verified by reading, not execution. Do not merge until the run log shows them green by name.

### Scope reminder for the record
The PiiScrubber change alters PII-redaction lifecycle on every request and job teardown, which is broader and more privacy-sensitive than the PR title implies. Captured in the PR description and in tracking issue #286 Section 2 so it is not lost.

Architecture is right (single source of truth removes the duplicated key list that caused PR #174's original drift). Ship once CI is green.

MelissaOneil (approved):

Issue #293, opened 2026-05-27
Frontend: refetch board on share/details modal open when lite-prefetched (fast-follow to #286)


swahlquist:
Fast-follow to the server-side lite serialization fix in #286 (PR adding `as_lite` to `#tree`/`#bulk`).

## Problem
The lite serialization for `#tree`/`#bulk` intentionally omits `shared_users`, `copies`/`copy`, and `parent_board_id` from the prefetch payload to kill the per-board N+1 (`JsonApi::Board.as_json` fan-out). The frontend serves a board first seen via the `/tree` prefetch from `boardDetailCache` on navigation (a cache hit, no `/show` reload) for the 5-minute TTL. Within that window:

- `share-board.hbs:80` reads `board.shared_users` -> a genuinely shared board renders an **empty share list**, which an editor could misread as "shared with nobody".
- `models/board.js` `multiple_copies` (from `copies`) and `parent_board_id` ("back to original" affordance) are also absent.

Bounded by: edit-permission gating, the 5-min TTL, and self-heal on save/edit (which invalidates the cache) or manual reload. `store.push` merges, so a previously-full record is NOT clobbered; the gap only affects boards first materialized from a lite prefetch. Surfaced by adversary review (2026-05-26), rated Medium.

## Fix options
1. Have the board-details / share modals call `model.reload()` on open when the model lacks `shared_users` / `copies` (mirrors the existing `batch_recording` reload in `components/board-details.js:98`).
2. Mark lite-sourced cache entries so navigating into one schedules a background `/show` refresh before edit/share affordances render.

## Test
Add a regression test on the share modal against a lite-cached, genuinely-shared board.

## Notes
- Server kill-switch exists if this proves disruptive before the frontend lands: `Setting.set('tree_lite_serialization', 'false')` reverts to full serialization without a deploy.
- Node 20 only; i18n any new strings.

swahlquist:
Resolved in PR #313 (squash-merged to staging as b4db9e4d7, auto-closed on merge) and manually validated on staging (lingolinq-staging.onrender.com).

Board#reload_if_lite now refetches lite-sourced boards (from the #tree/#bulk as_lite prefetch) when the share or board-details modal opens, restoring the "Shared with" list and the "Copied From" link. Includes a sharing_locked() guard on share_with_user / unshare / make_public to narrow the in-flight-reload overlap window.

[...]

PR #294, opened 2026-05-27
fix(boards): lite serialization for #tree and #bulk to kill as_json N+1 (#286)

swahlquist:
Implements remediation #1 from the #tree Rack::Timeout RCA (`outputs/docs/2026-05-24-tree-bulk-rca.md`, issue #286). Freeze-scoped.

## Problem
`#tree` and `#bulk` serialized every board through the full `JsonApi::Board.as_json` path used by `#show`. Across `MAX_TREE=500` descendants that path fans out into thousands of unindexed per-board queries: `parent_board` (x2), `find_copies_by` (`board.rb:620`, a `board_content` join + limit-15 sort), `shared_users`, the delete/admin `using_user_names` `UserBoardConnection`+`User` lookup, and a per-image `ButtonImage.find_by_global_id` skin lookup. PR #281's login prefetch started firing `#tree` at session-start scale, which surfaced as `Rack::Timeout` in Sentry.

## Fix
Add an `:as_lite` arg to `JsonApi::Board` honored by `build_json` and `extra_includes`. When set it skips that enrichment while keeping board identity, `permissions`, grid content, and `image_urls` (so prefetched thumbnails still render). `#tree` and `#bulk` pass it; the full `#show` / `#index` path is untouched (guards are truthy-only).

Validated against staging head `148839e89` before implementing: all RCA findings still hold (with line drift); `parent_board` actually fires twice per board, not once.

### Measured impact
On image-bearing boards: **5 queries/board -> 1 query/board** (5x). Not O(1): the surviving 1/board is `images_and_sounds_for` (`known_button_images`), which is remediation #3 (request-scoped image cache), tracked separately in #286.

### Kill-switch (deploy-free)
Lite is ON by default. Revert without a deploy from a console:
`Setting.set('tree_lite_serialization', 'false')` (busts the Redis cache; effective next request).

## Tests
`spec/controllers/api/boards_controller_spec.rb`: contract specs (404, permission filtering, `MAX_TREE`/`MAX_BULK` caps, lite payload shape), a threshold-free behavioral guard (`find_copies_by` / `shared_users` / `ButtonImage.find_by_global_id` never fire), bounded per-descendant query specs, and a kill-switch fallback spec. Full file: 195 examples, 0 failures (1 pre-existing pending). Verified the guard fails when `as_lite` is removed.

## Dual-reviewer pass
- **Adversary:** found the original query specs were tautological (empty-grid boards hid the surviving image cost and falsely claimed O(1)) -> **fixed** here (image-bearing fixtures + honest reframing + behavioral guard). Also flagged a Medium client-contract gap (lite-cached boards show empty share lists in the Ember modals on a cache hit) -> filed as #293 with the kill-switch as the interim lever.
- **Senior-dev pass:** Codex CLI was unavailable in the authoring environment, so this was a Claude senior-dev checklist pass, not a cross-model second opinion. Flagged the missing kill-switch (added here) and confirmed the change is a net PII reduction (drops supervisor identities from the prefetch payload).

## Index overlap
`#index`/`paginate` shares the same `build_json`/`extra_includes`. Technical interaction with Traci's `#index` N+1 work is low (index passes no `:permissions`), but merge-collision probability on these lines is high. Whoever lands second should rebase.

Addresses #286 (remediation #1 only; #2 batch-preload and #3 image cache remain). Fast-follow: #293.

### DIFF
--- a/app/controllers/api/boards_controller.rb
+++ b/app/controllers/api/boards_controller.rb
@@ -420,12 +420,27 @@ class Api::BoardsController < ApplicationController
     # as_lite drops the per-board N+1 enrichment (parent_board, find_copies_by,
     # shared_users, per-image ButtonImage lookups) that made this endpoint
     # Rack::Timeout under MAX_TREE fan-out. See RCA 2026-05-24, issue #286.
-    root_json = JsonApi::Board.as_json(root, wrapper: true, permissions: @api_user, skip_subs: true, as_lite: true)
+    # Gated by a deploy-free kill-switch (see lite_board_serialization?).
+    lite = lite_board_serialization?
+    root_json = JsonApi::Board.as_json(root, wrapper: true, permissions: @api_user, skip_subs: true, as_lite: lite)
     descendants_json = descendants.map do |b|
-      JsonApi::Board.as_json(b, wrapper: true, permissions: @api_user, skip_subs: true, as_lite: true)
+      JsonApi::Board.as_json(b, wrapper: true, permissions: @api_user, skip_subs: true, as_lite: lite)
     end
     render json: { root: root_json, descendants: descendants_json }
   end
+
+  # Kill-switch for the #tree / #bulk lite serialization (RCA 2026-05-24,
+  # issue #286). Lite is ON by default. Ops can revert to the full as_json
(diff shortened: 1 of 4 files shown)


MelissaOneil (approved):
gtg

PR #301, opened 2026-05-28
fix(security): route admin database_contents through model allowlist (#286)

swahlquist:
## Summary

Closes the Section 1 finding in #286. `GET /api/v1/database_contents` read rows via raw `exec_query("SELECT *")`, serializing the underlying column bytes and bypassing the `secure_serialize` (go_secure) decryption/abstraction layer. That exposed encrypted FERPA/HIPAA blobs (e.g. `users.settings`, `log_sessions.data`, `contact_messages.settings`) and every other column of any public table to admin / admin_support_actions users, with no field-level control. Legacy `**`-prefixed rows would have leaked plaintext.

## What changed

- **Deny-by-default `(table -> model)` allowlist** (`ALLOWED_MODELS`). Any table not listed returns 404. `users`, `log_sessions`, `contact_messages`, `devices`, `developer_keys`, and all log/audit tables are no longer browsable.
- **Reads route through the ActiveRecord model**, never raw SQL. `exposable_columns` strips the model's `secure_serialize` column automatically (fail-safe for any future encrypted column) plus an explicit `SENSITIVE_COLUMNS` denylist for plaintext landmines.
- `safe_constantize` so a misconfigured allowlist entry returns 404 instead of a 500.
- **`AuditEvent` logged per successful read** (requester, table, limit, offset, rows returned) for FERPA/HIPAA accounting-of-disclosures. Best-effort / fail-open: an audit-write error never breaks an authorized read. 404/403 paths are not logged (no disclosure).
- `model.count` guarded against never-analyzed large tables (`pg_class.reltuples = -1`).

## Dual-reviewer pass

Ran the senior-dev (`/review-pr`) then red-team (`/adversary-review`) passes. The adversary caught three real plaintext leaks the first two passes missed, all verified against the code before fixing:

- **High** `boards.search_string`: plaintext denormalized board content (button labels = AAC user vocabulary). Now stripped via `SENSITIVE_COLUMNS`.
- **High** `board_locales.search_string` + `tsv_search_string`: same data plus the tsvector. `board_locales` removed from the allowlist entirely (search table, no admin-browse value).
- **Medium** `licenses.metadata` + `external_reference`: plaintext (`License` has no `secure_serialize`). Now stripped.

Audience was verified to be managers of the global admin org (`Organization.admin_manager?`), i.e. LingoLinq staff, so the cross-tenant support view is intended; no audience change was needed.

## Tests

`spec/controllers/api/database_contents_controller_spec.rb`: 27 examples, 0 failures. Covers auth gates, 404 for each sensitive/credential table, proof that decrypted secrets and plaintext sensitive columns never appear in the body, the audit-event write, and column/row shape.

## Follow-up (not in this PR)

The sibling `GET /api/v1/database_schema` (metadata only, lower stakes) has the same no-audit gap. Worth a separate issue rather than widening this PR.

### DIFF
--- a/app/controllers/api/database_contents_controller.rb
+++ b/app/controllers/api/database_contents_controller.rb
@@ -5,11 +5,50 @@ class Api::DatabaseContentsController < ApplicationController
   DEFAULT_LIMIT = 50
   MAX_LIMIT = 500
 
+  # Deny-by-default allowlist for the admin schema explorer. Only the tables
+  # listed here may be browsed; every other table returns 404. Each entry maps a
+  # table to its ActiveRecord model so rows are read through the model layer
+  # (never via raw `SELECT *`), and the model's secure_serialize column is
+  # stripped from the response automatically (see #exposable_columns).
+  #
+  # Tables that hold regulated PII/PHI in non-encrypted columns (users,
+  # log_sessions, contact_messages, devices, ...) or plaintext credentials
+  # (developer_keys) are deliberately omitted. board_locales is also omitted:
+  # it is a search/tsvector table whose search_string and tsv_search_string are
+  # a plaintext, denormalized copy of board content (button labels), so it
+  # carries the same FERPA/HIPAA data that secure_serialize protects with no
+  # admin-browse value. Widen this map only after a privacy review of every
+  # non-encrypted column the candidate table would surface.
+  ALLOWED_MODELS = {
+    'organizations'          => 'Organization',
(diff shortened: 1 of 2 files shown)


swahlquist:
Follow-up filed for the sibling `database_schema` endpoint (audit trail + allowlist parity): #302. Deliberately out of scope for this PR to keep it focused on the row-dump leak.

MelissaOneil (approved):

PR #304, opened 2026-05-28
fix(security): redact secure_serialize columns in /database admin dump

swahlquist:
## Summary

The `/api/v1/database_contents` admin schema explorer ran `SELECT *` via `exec_query` and serialized every cell raw (`app/controllers/api/database_contents_controller.rb`), **bypassing the `secure_serialize` (go_secure) decryption layer**. That emitted the encrypted-at-rest blobs for ~30 models (`users.settings`, `log_sessions.data`, `contact_messages.settings`, `boards.settings`, `organizations.settings`, `api_calls.data`, etc.) in the clear to an admin API. FERPA/HIPAA-implicated.

This is the **Scot-owned High** finding in Section 1 of #286.

## Fix

- **Redact, do not decrypt.** Decrypting would worsen the leak. Instead, introspect `Model.secure_column` (go_secure sets a class-level accessor) to build a memoized `table_name => [secured columns]` map, and replace those cells with `[redacted: secured]`.
- Add `redacted_columns` to the payload (additive; `rows` shape unchanged, so the frontend schema explorer is unaffected).
- Non-secured columns are untouched, so the explorer remains useful.

## Why no feature flag

The CLAUDE.md feature-flag rule exists because AAC users are disruption-sensitive. This hardens an **admin-only internal API**, not an AAC-user-facing surface, so no flag is warranted.

## Residual (out of scope for this finding)

Plaintext PII columns that are sensitive-by-design but not `secure_serialize`'d (e.g. emails) remain visible to admins. That is pre-existing admin visibility; a column allow-list is a separate follow-up.

## Test plan

- [x] `bundle exec rspec spec/controllers/api/database_contents_controller_spec.rb` (8 examples, 0 failures)
- [x] New test proves: `redacted_columns` includes `settings`, every `settings` cell == `[redacted: secured]`, the raw ciphertext never appears in the response body, and a non-secured column (`user_name`) still shows its real value
- [x] New test proves a table with no secured column returns `redacted_columns: []`

## Reviewer note

Dual-reviewer (`/review-pr` Codex + `/adversary-review` Claude) per #286 Phase 1 governance, since this touches security + user-data flows. Do not merge until both pass. Refs #286.

### DIFF
--- a/app/controllers/api/database_contents_controller.rb
+++ b/app/controllers/api/database_contents_controller.rb
@@ -9,6 +9,25 @@ class Api::DatabaseContentsController < ApplicationController
   # never leaves the model layer through this raw SELECT * dump.
   REDACTED_PLACEHOLDER = '[redacted: secured]'
 
+  # PaperTrail snapshot columns embed the raw ciphertext of a secured column for
+  # any paper-trailed secured model (e.g. User/Board settings), so the versions
+  # audit table re-exposes the same encrypted blobs even though it has no
+  # secure_serialize of its own. Redact those snapshot columns by table name.
+  SNAPSHOT_REDACTED_COLUMNS = { 'versions' => %w[object object_changes] }.freeze
+
+  # Plaintext columns holding live credentials or impersonation tokens that are
+  # NOT secure_serialize'd, so model introspection cannot find them. Redacted by
+  # table name. A broader sensitive-plaintext-column audit is tracked as a
+  # follow-up to #286.
+  SENSITIVE_PLAINTEXT_COLUMNS = {
+    'developer_keys' => %w[secret key],
+    'purchase_tokens' => %w[token],
+    'organizations' => %w[external_auth_key],
+    'supervisor_relationships' => %w[consent_response_token],
+    'beta_feedback_recordings' => %w[token],
+    'external_nonces' => %w[nonce]
+  }.freeze
+
   # GET /api/v1/database_contents?table=NAME&limit=50&offset=0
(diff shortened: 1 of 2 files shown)


swahlquist:
Closing as superseded by #301 (merged to staging 2026-05-28), which fixed the same #286 `/api/v1/database_contents` leak with a stronger architecture.

This branch was cut 2026-05-27 off `Omer/feature/wordprediction (#300)` and never incorporated #301 (`git merge-base --is-ancestor` confirms #301 is not in this branch). The two take incompatible approaches to the same controller:

- **#304 (this PR):** `SELECT *` over any public-schema table, redacting known secured/sensitive columns by name. `allowed_tables` returns every table in the public schema, so this **fails open** (a new sensitive table or column leaks until someone denylists it).
- **#301 (now on staging):** deny-by-default `ALLOWED_MODELS` allowlist of 6 vetted tables, read through the model layer with `secure_column` + `SENSITIVE_COLUMNS` stripping. Every other table 404s, so it **fails closed**.

#301 also already redacts `organizations.external_auth_shortcut` and `boards.search_string`, both of which this PR's denylist misses, so #304 is a net coverage regression on top of the architectural one.

Credit retained: the `secure_column` introspection idea here is sound, and #301's `exposable_columns` independently uses the same primitive. The idea was effectively adopted; this delivery vehicle is obsolete.

Follow-up #305 has been re-scoped against the staging implementation. Refs #286, #301, #305.

Reviewed by Claude Code adversary pass for Scot.

PR #312, opened 2026-05-30
fix(boards): drop eager-load before id-only selects in #index (LINGOLINQ-RAILS-J)

swahlquist:
## What

Fixes the live staging 500 on `GET /api/v1/boards` (Sentry **LINGOLINQ-RAILS-J**): `ActiveModel::MissingAttributeError: missing attribute 'parent_board_id' for Board`, first seen 2026-05-29 22:58 UTC.

## User impact

Board search fires two queries. The public/online catalog query was fine; the **"Your Boards" personal query** (`user_id: 'self', allow_job: true` in `controllers/search.js:80`) routed into the broken branch and 500'd. So a logged-in user searching boards got online results but their own matching boards failed to load. The board picker used when linking a board to a button (`content-grabbers.js` also sends `allow_job`) was affected the same way.

## Root cause

PR #303 added `:parent_board` to the index eager-load (`boards.includes(:board_content, :parent_board)`) to fix a real N+1 in `lib/json_api/board.rb` (`build_json` reads `board.parent_board` per result). Correct fix for that problem, but two branches in the same action harvest IDs with a partial `select('id, board_content_id')` that omits `parent_board_id`. With `:parent_board` now eager-loaded, ActiveRecord's `belongs_to` preloader reads the `parent_board_id` foreign key off those partial records and raises, because the column was never selected. `board_content` survived only because `board_content_id` was in the select.

Both broken branches are gated behind a search query (`q` present), which is why empty browse lists never hit it.

## Fix

`.except(:includes)` on the two id-only extraction sites so the preloader is not invoked on records that are immediately discarded (also drops a wasted preload over up to 500 throwaway rows). The full serialization path keeps `includes(:parent_board)`, so PR #303's N+1 fix is preserved. A guard comment at the eager-load site warns against a future `references`/`eager_load` refactor reopening this.

Not related to lite serialization (PR #294): `#index` never passes `as_lite`, and the `tree_lite_serialization` kill switch does not touch this path.

## Tests

- Two regression specs (the `allow_job` private path and the public+`user_id` twin), each verified to reproduce the exact `MissingAttributeError` against the unfixed line and pass with the fix.
- Full `#index` suite green (37 examples, 0 failures).

## Review

Cleared the dual-reviewer gate: `/review-pr` (Approve) and `/adversary-review` (Ship, all findings Low).

Fixes LINGOLINQ-RAILS-J

### DIFF
--- a/app/controllers/api/boards_controller.rb
+++ b/app/controllers/api/boards_controller.rb
@@ -43,6 +43,11 @@ class Api::BoardsController < ApplicationController
     # extra SELECT per board (~25 at default per_page). Verified by the
     # regression spec in spec/controllers/api/boards_controller_spec.rb.
     # See docs/task-management/2026-05-27-boards-index-n-plus-one.md.
+    # NOTE: id-only partial selects off this relation (the two
+    # `.except(:includes)` sites below, ~line 173 and ~307) drop this
+    # eager-load so the :parent_board preloader does not read an unselected
+    # parent_board_id FK (LINGOLINQ-RAILS-J). If you ever switch this to a
+    # `references`/`eager_load`-backed join, update those two sites too.
     boards = boards.includes(:board_content, :parent_board)
     other_boards = nil
     other_searchable_board_ids = nil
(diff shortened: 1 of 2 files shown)


MelissaOneil (approved):
This PR was automatically reviewed and sent for approval by Scot's BFF, Claude Code automation.

PR #312
fix(boards): drop eager-load before id-only selects in #index (LINGOLINQ-RAILS-J)
🔗 https://github.com/lingolinq/LingoLinq-AAC/pull/312
staging ← scot/fix/boards-index-partial-select-preload · +54 / -3 · 2 files

The issue (Sentry LINGOLINQ-RAILS-J)
Live staging 500 on GET /api/v1/boards: ActiveModel::MissingAttributeError: missing attribute 'parent_board_id', first seen 2026-05-29. A logged-in user searching boards got online results but their own "Your Boards" matches 500'd. The board picker used when linking a board to a button was hit the same way.

Root cause
PR #303 added :parent_board to the index eager-load to fix a real N+1. But two branches harvest IDs with a partial select('id, board_content_id') that omits parent_board_id; the belongs_to preloader then reads that missing foreign key and raises. Only triggered when a search query is present, which is why empty browse never hit it.

What it does
.except(:includes) on the two id-only extraction sites so the preloader is not invoked on throwaway rows (also drops a wasted preload over up to 500 rows). The full serialization path keeps includes(:parent_board), so #303's N+1 fix is preserved. A guard comment warns against a future refactor reopening it.

Tests
Two regression specs (private allow_job path + public/user_id twin), each verified to reproduce the exact error pre-fix and pass after. Full #index suite green (37 examples).

Review: already cleared ✅
Dual-reviewer gate passed and recorded in the PR body: /review-pr → Approve, /adversary-review → Ship, all findings Low.

Status: ready to merge ✅
MERGEABLE, CI green (rspec, build-and-test, security-scan, secret-detection). No rebase needed. This one is good to go as-is.

PR #313, opened 2026-05-30
fix(boards): refetch lite-sourced boards in share/details modals (#293)

### LINKED ISSUE #293: Frontend: refetch board on share/details modal open when lite-prefetched (fast-follow to #286)

swahlquist:
## What this does

Fixes #293: boards first materialized from a `#tree`/`#bulk` **lite** prefetch (server-side `as_lite`, PR #294) omit `parent_board_id`, `copies`/`copy`, and the edit-gated `shared_users`. As a result:

- the **share modal** rendered a genuinely-shared board with an empty "Shared with" list, which an editor misreads as "shared with nobody";
- **board-details** dropped the "Copied From" link.

Adds `Board#reload_if_lite`, called by both modals on open. It detects a lite-sourced record by the absence of `parent_board_id` (a full `/show` always serializes that key, value may be `null`) and refetches once, guarding against new/unsaved records and in-flight reloads. On `reload()` failure it degrades to the lite view rather than throwing.

## Changes

- `app/frontend/app/models/board.js`: `reload_if_lite` with single-flight guard and graceful failure.
- `app/frontend/app/components/board-details.js`: calls `reload_if_lite` on open.
- `app/frontend/app/components/share-board.js`: calls `reload_if_lite` on open; blocks `share_with_user`/`unshare` while `reloading_detail` is set (see review note below).
- `app/frontend/app/templates/components/share-board.hbs`: transient "Loading..." hint (existing `loading` i18n key, `aria-live="polite"`); share/unshare buttons disabled while reloading.
- `lib/json_api/board.rb`: comment only, documenting the contract that the non-lite path must always emit `parent_board_id` (the client lite-detector depends on it).
- Tests: `tests/unit/models/board-reload-if-lite-test.js` (lite-detect, full-skip, new-skip, in-flight dedup, reload-rejection, re-dispatch-after-failure, mid-flight `sharing_key` survival) and `tests/unit/components/share-board-guard-test.js` (action guard).

## Dual-reviewer pass

Reviewed with `/review-pr` (senior-dev) and `/adversary-review` before this PR.

- **Adversary High**: an in-flight `reload()` could overlap a `sharing_key` set+save in the share/unshare actions, risking a silently-dropped sharing change (access-control mutation on student board data). Resolved: `sharing_locked()` blocks both actions (with a wait notice) and the buttons are disabled while `reloading_detail` is set.
- **Medium**: detector depends on the #294 serializer always emitting `parent_board_id`. Verified against `lib/json_api/board.rb` and documented inline.
- **Medium**: added the missing tests (rejection, re-dispatch, mid-flight survival, guard).
- Feature flag: exempt. This is a regression fix restoring pre-#286 behavior. The only new UI is the transient loading hint.

## Test execution caveat (please read)

The new unit tests are **lint-clean and load without error, but their assertions do not execute** under `ember test`. This repo's harness currently runs only about 3 behavioral tests (1 smoke + 2 skipped acceptance) out of 1369; the rest are ESLint/TemplateLint. **No unit behavioral tests run at all**, including staging's pre-existing `copying-board`, `telemetry`, and `eval-*` specs. This is a **pre-existing AMD test-loader defect**, documented in `app/frontend/tests/test-helper.js:21-26` and `app/frontend/tests/README-test-loader.md`, not introduced by this PR. It is tracked separately (see linked issue). The implementation here was verified by tracing the code paths and the serializer contract.

## Scope

Frontend board-details/share modal path, plus a comment-only note in the serializer. No backend serialization logic or #286 work was modified.

Refs #293, #294.


swahlquist:
Test-loader caveat referenced in the description is tracked in #314 (pre-existing harness defect, not caused by this PR).

PR #316, opened 2026-05-31
Add phased background board prefetch for online navigation and offline sync

MelissaOneil:
## Summary
- Adds a shared `board_prefetch_planner` and phased pipeline that prefetches board trees in priority order: home → liked → owned roots → public (catalog + global search).
- Session navigation uses background `/tree` fetches (fire-and-forget, rate-limited, pauses when tab hidden); offline sync seeds the same roots into the existing BFS queue in both persistence modules.
- Enables `background_board_prefetch` globally so extended prefetch is default behavior for all users.

## Test plan
- [ ] Log in and confirm dashboard loads normally with no blocking wait for prefetch
- [ ] Open home board and sub-boards — navigation should be instant once prefetch completes (Network tab shows background `/tree` calls)
- [ ] Like a board, wait for background prefetch, open it — should cache-hit without extra round-trip
- [ ] Trigger offline sync while online — owned/public roots should appear in sync progress over time
- [ ] Confirm prefetch pauses when tab is backgrounded and resumes on reconnect (`persistence.online`)

### DIFF
--- a/app/frontend/app/utils/board_detail_cache.js
+++ b/app/frontend/app/utils/board_detail_cache.js
@@ -18,6 +18,7 @@
 // and on entering edit mode so editors never read stale state.
 
 import persistence from './persistence';
+import boardPrefetchPlanner from './board_prefetch_planner';
 import { later as runLater } from '@ember/runloop';
 import RSVP from 'rsvp';
 import LingoLinq from '../app';
@@ -26,10 +27,7 @@ var TTL_MS = 5 * 60 * 1000;
 var MAX_PREFETCH = 20;
 var WARM_BATCH = 20;
 var WARM_BATCH_GAP_MS = 80;
-var CATALOG_TREE_GAP_MS = 400;
-var CATALOG_ROOTS_PER_PAGE = 50;
-var CATALOG_ROOT_CAP = 100;
-var CATALOG_ACCOUNT = 'lingolinq';
+var TREE_GAP_MS = 400;
 
 // key (e.g. "user_name/boardname") → entry
 var _by_key = {};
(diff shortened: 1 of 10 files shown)


app/frontend/app/utils/board_detail_cache.js
```diff
@@ -451,150 +496,197 @@ export default {
...
+    if (!phaseDone.phase1) {
+      chain = chain.then(function() {
+        if (_document_hidden() || !_is_online()) { return RSVP.resolve(); }
+        var lookups = boardPrefetchPlanner.collectHomeLookups(user);
+        if (!lookups.length) {
+          phaseDone.phase1 = true;
+          return RSVP.resolve();
+        }
+        return _process_roots_sequentially(_this, lookups, warm_opts, TREE_GAP_MS).then(function(completed) {
+          _complete_phase_if_done(phaseDone, 'phase1', completed);
```
MelissaOneil, line 521: **Item 1 — phase completion (fixed in 3df7caa)**

Previously `_process_roots_sequentially` resolved successfully even when interrupted mid-phase (tab hidden / offline), and callers always set `phaseDone.phaseN = true`. That caused later phases to be skipped for the rest of the session.

**Fix:** return `true` only when every root in the phase was processed or already fresh; return `false` on interruption. `_complete_phase_if_done` only marks the phase when `completed === true`.

app/frontend/app/utils/board_detail_cache.js
```diff
@@ -451,150 +496,197 @@ export default {
...
     var user_id = user.get('id');
     if (!user_id) { return; }
-    this._prefetched_user_ids = this._prefetched_user_ids || {};
-    if (this._prefetched_user_ids[user_id]) { return; }
-    this._prefetched_user_ids[user_id] = true;
-    var home_key = user.get('preferences.home_board.key');
-    var home_id = user.get('preferences.home_board.id');
-    var lookup = home_key || home_id;
+    if (!_is_online()) { return; }
+    if (_document_hidden()) { return; }
```
MelissaOneil, line 627: **Item 2 — no resume when tab becomes visible (follow-up)**

`prefetch_for_user` returns early when `document.hidden` at call time. Reconnect via `on_online_board_prefetch` resumes incomplete phases, but switching back to a visible tab does not re-trigger prefetch.

**Suggested fix:** add a `visibilitychange` listener (or retry from existing observers) that calls `prefetch_for_user` when `!document.hidden && persistence.online`. Without this, login with the tab in the background may never start prefetch until an offline/online flip.

app/frontend/app/utils/board_prefetch_planner.js
```diff
@@ -0,0 +1,247 @@
...
+  var lookup = hb.key || hb.id;
+  return lookup ? [lookup] : [];
+}
+
+export function collectLikedLookups(user, seen) {
+  if (!user || !user.get) { return []; }
+  seen = seen || {};
+  var refs = user.get('stats.starred_board_refs') || [];
+  var result = [];
+  refs.forEach(function(ref) {
```
MelissaOneil, line 42: **Item 3 — suggested liked boards behavior change (confirm intent)**

Legacy offline sync filtered starred refs with `sync_all || !ref.suggested`. `collectLikedLookups` includes **all** refs with no `suggested` filter.

With `background_board_prefetch` globally enabled, users may sync/prefetch more boards than before (including home-board picker suggestions).

**Question:** Is this intentional? If not, restore the `!ref.suggested` filter unless `preferences.sync_starred_boards === true` (matching old `sync_boards` behavior).

MelissaOneil (commented):
Senior dev review follow-up (items 1–3). Item 1 is fixed in this commit; items 2–3 are follow-ups.

app/frontend/app/utils/board_detail_cache.js
MelissaOneil: Resolved in 3df7caa — `_complete_phase_if_done` only marks a phase when every root in that phase was processed.

app/frontend/app/utils/board_detail_cache.js
MelissaOneil: Resolved in d6339ac — `visibilitychange` in app-state init re-triggers `prefetch_for_user` when the tab becomes visible (online + logged in).

app/frontend/app/utils/board_prefetch_planner.js
MelissaOneil: Resolved in d6339ac — `collectLikedLookups` skips `ref.suggested` unless `sync_starred_boards === true`, matching legacy sync.

PR #335, opened 2026-06-02
Add supervisor caseload board prefetch

MelissaOneil:


MelissaOneil:
Senior review finding:

The caseload prefetch marks a supervisor/supervisee pair as prefetched as soon as `_run_prefetch_pipeline` resolves, but that pipeline resolves even when it intentionally stops early because the tab became hidden or the app went offline. `_process_roots_sequentially` returns `false` in those cases and `_run_prefetch_pipeline` leaves the phase incomplete so later visibility/online triggers can resume. The new `_prefetched_caseload_supervisee_ids[key] = true` bypasses that retry path, so a supervisor who backgrounds the tab or loses connection mid-prefetch can permanently skip that supervisee for the rest of the session.

Please only mark the caseload pair complete when the relevant per-user phases are actually complete, or remove the additional per-supervisee completed flag and rely on `_prefetch_phase_done`/`_prefetch_pipeline_running` to dedupe and resume.

PR #338, opened 2026-06-03
Fix email delivery configuration

MelissaOneil:
## Summary
- Fix SES email delivery by allowing the mailer config to use the standard AWS region environment variable.
- Update shared email signatures to say "The Lingolinq Team" instead of using the placeholder company name.
- Document the SES region gotcha for future email delivery investigations.

## Test plan
- `ruby -c config/initializers/amazon_ses.rb`
- `ruby -c app/mailers/mailer_helper.rb`
- ReadLints on the changed Ruby helper and docs files
- Read-only SES `get_send_quota` check succeeded locally after applying the region fallback

### DIFF
--- a/app/frontend/app/index.html
+++ b/app/frontend/app/index.html
@@ -414,7 +414,7 @@
         if (typeof window.domain_settings === 'undefined') {
           window.domain_settings = {
             app_name: 'LingoLinq',
-            company_name: 'Someone',
+            company_name: 'Lingolinq',
             logo_url: '/images/LL_Logo_Line_Bright.png',
             full_domain: true,
             support_url: 'https://lingolinq.zendesk.com',
--- a/app/mailers/mailer_helper.rb
+++ b/app/mailers/mailer_helper.rb
@@ -5,7 +5,7 @@ module MailerHelper
   end
 
   def email_signature
-    "-The #{JsonApi::Json.current_domain['settings']['company_name']} Team"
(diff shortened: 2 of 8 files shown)


MelissaOneil (commented):
Senior review finding before follow-up fixes:

- Important: `app/mailers/mailer_helper.rb` hardcodes every outbound email signature to `The Lingolinq Team`, which removes existing custom-domain/company branding and likely breaks existing mailer specs that expect domain-specific signatures. The better fix is to keep the shared signature using the configured company name, and update the default company-name source from the placeholder `Someone` to `Lingolinq`.
- Suggestion: add/update focused mailer coverage for the default signature and custom-domain signature so this behavior stays intentional.

MelissaOneil:
Follow-up pushed in `f32bfe1b6`:

- Restored `email_signature` to use the configured `company_name`, so custom-domain/company branding still works.
- Changed the default server-side company name source from `Someone` to `Lingolinq` in `JsonApi::Json.default_domain` and domain fallback settings.
- Updated the Ember dev `domain_settings` stub to match `Lingolinq`.
- Updated focused specs for default and custom-domain signatures/domain settings.

Focused check passed:
`DB_USER=melis bundle exec rspec spec/mailers/user_mailer_spec.rb:47 spec/lib/json_api/json_spec.rb:160 spec/controllers/application_controller_spec.rb:502`

PR #358, opened 2026-06-09
compliance(signup): beta notice + privacy link in consent flows

swahlquist:
## What this does

Adds the Privacy Policy to the places it was missing in the consent flow, and adds a time-boxed Beta Program Notice to the live privacy policy. Part of the beta-program compliance work (Action items #5 and #6 from the beta docs review).

1. **Beta Program Notice** on the live privacy policy (`privacy.hbs`): pre-release "as is" status, keep-your-backup-voice safety note for AAC users, opt-in anonymized logging, under-13 consent, clinician BAA expectation, and a pointer to the retention schedule. "Last Updated" bumped to June 9, 2026.
2. **Privacy Policy acknowledgment** beneath all three signup consent checkboxes (Google signup, under-13 COPPA, standard 13+), as a sibling line so it stays out of the contractual Terms checkbox.
3. **COPPA direct-notice:** Privacy Policy link added to the parental consent request email (HTML + text), with `@privacy_url` exposed from the mailer.
4. **i18n:** new strings + the bumped date registered in `public/locales/en.json` (the resolver prefers the locale file over inline `{{t}}` defaults, so this is required for the change to render in English).

## Why the design choices

- **Privacy link is a sibling, not inside the checkbox label.** A Terms of Service is a contract you *agree to* (the checkbox); a Privacy Policy is a notice you *acknowledge*. Keeping the privacy link out of the `<label>` also avoids an `ember-template-lint` `no-nested-interactive` violation (two interactive elements in one label).
- **Notice lives only on the live Ember page, not the `_privacy.html.erb` server partial.** `/privacy` routes to `boards#privacy`, which does `render :index` (the SPA), so that partial is dead code; editing it would create a divergent duplicate of legal copy.

## Reviewed

Ran both reviewers before opening this:
- `/review-pr` (Codex senior-dev pass)
- `/adversary-review` (Claude red-team pass)

Both flagged the same blocking issue: the date/keys were missing from `en.json`, so English users would have kept seeing the old date. Fixed. The dead-code partial duplication was also flagged and resolved.

## Reviewer call-outs (need a human / legal decision)

These are intentional and left for review, not oversights:

1. **BAA wording is a policy statement, not an enforced control.** The notice says clinicians "must have a signed BAA before entering patient information," but nothing in the code blocks PHI entry without a BAA on file. Given HIPAA/BAA is now in scope for the cloud migration, please confirm whether to keep this as a policy assertion or soften to contractual-obligation framing ("are required by their agreement to...").
2. **COPPA "direct notice before consent" rests on the email.** Consent is granted one-click on the `/parental_consent/complete` page load, with no privacy content on that page, so the email is the only pre-consent notice. Please confirm the email-link pattern meets the FTC "direct notice" bar, or we can surface a short privacy summary on the consent-completion page.

## Follow-ups (deferred, not blocking)

- Run `i18n_generator.rb` to propagate the new keys into the other `public/locales/*.json` for translation (non-English users currently get the English fallback).
- `target="_blank"` links have no "opens in new tab" cue (a11y nit; matches the existing terms-link pattern, fix uniformly if desired).
- `@privacy_url` inherits the same custom-domain host dependency as the existing `@consent_url` in the mailer (pre-existing, Low).

## Test plan

- [ ] Visit `/privacy` in English and confirm "Last Updated: June 9, 2026" and the Beta Program Notice render.
- [ ] Start signup (standard, under-13, and Google paths) and confirm the Privacy Policy link appears and resolves to `/privacy`.
- [ ] Trigger a parental consent email and confirm the Privacy Policy link is present in both HTML and text versions.


swahlquist:
## Three-model review + fixes applied (`38abf3c3d`)

Ran the dual-reviewer pattern plus a third independent model on this PR:

| Pass | Model | Headline |
|---|---|---|
| Senior-dev (`/review-pr`) | Claude | Approve with comments |
| Adversary (`/adversary-review`) | Claude red-team | 2 High on legal copy |
| `codex review` | DeepSeek v3.2 (OpenRouter) | P1 unstyled CSS classes |

**Convergent findings → fixed in `38abf3c3d`:**
- **Beta notice over-promised.** It claimed we review "communication logs ... stripped of names and personal profiles," which (a) contradicted the Private Thoughts Guarantee in Section 1 and (b) asserted a scrubbing control that isn't implemented. Reworded to reference anonymized **diagnostic usage data** only, and to reaffirm we do not review the words a communicator generates.
- **BAA sentence** reworded to put the obligation on the clinician's organization without implying platform enforcement (points to contacting us to execute one).
- **Section 7 cross-reference** corrected to "retention windows described in Section 7" (there is no "Data Retention Schedule" heading in that section).
- **Signup acknowledgment** changed from "I also acknowledge ..." (implied an affirmative act that was never captured) to passive notice tied to account creation. Decision: a privacy policy is **notice, not a contracted term**, so binding consent stays on the existing Terms checkbox; we did not add a second required checkbox (signup friction for this audience). 
- **Added SCSS** for `.la-privacy-beta-notice` and `.register-privacy-ack`, which were rendering unstyled (caught by the DeepSeek pass).

`en.json` body is byte-identical to the `privacy.hbs` inline default, so English and non-English render the same text.

**Still wants a human/counsel glance (not code-blocking):**
- Final BAA wording and the comms-data language are conservative and honest now, but are policy statements, not enforced controls. If/when clinical features gate on BAA status, tighten further.
- The new strings exist in `en.json` / `en.yml` only; other locales fall back to English until `i18n_generator.rb` propagates them. Mechanical follow-up.

swahlquist:
**Update (826d98c58): passive notice → captured consent checkbox.**

Per follow-up review, the signup privacy acknowledgment was changed from a passive notice line ("By creating an account, you acknowledge…") to a **captured affirmative agreement**.

**What changed**
- The Privacy Policy (linked) is now folded into the **existing** signup consent checkbox, so one required tick covers **both** the Terms of Use and the Privacy Policy. The box already gates submit and mirrors the 13+ age attestation, so no new submit logic was needed.
- On the backend, `User#process_params` now records a versioned `settings['privacy_consent']` (`agreed_at` UTC ISO-8601 + `policy_version`) whenever `terms_agree` is set, alongside the existing `terms_agreed` timestamp.
- `policy_version` is stamped from `User::PRIVACY_POLICY_VERSION` (`2026-06-09`), kept in sync with the Privacy Policy "Last Updated" date.
- Both signup paths are covered: standard (sends `terms_agree`) and Google (forces `terms_agree=true` through the same `process_params`).
- Removed the now-unused passive-notice markup, its SCSS rule, and the `register_consent_privacy_ack_*` i18n keys; added a `register_consent_and` key; updated the consent-required error copy to name the Privacy Policy.
- Added a `process_params` spec asserting the consent record is stored with version + timestamp (passes; the unrelated pre-existing `user_spec.rb:916` LogSession failure was confirmed present on the clean branch too).

**Design note:** a Privacy Policy is disclosure, so bundling its acknowledgment with the Terms checkbox is standard practice; the one place that legally needs a *recorded* affirmative act (under-13) is already handled separately by the parental-consent email flow, not this checkbox.

**Still for human/counsel eyes:** whether `policy_version` should be sourced from a single shared config rather than a hand-bumped constant; locale propagation of the new/changed i18n keys via `i18n_generator.rb`.

swahlquist:

## Automated PR Review

| Pass | Verdict | Findings |
| --- | --- | --- |
| Senior-dev (Claude Sonnet 4.6) | `REQUEST_CHANGES` | 10 |

### Senior-dev review
The PR correctly improves COPPA/privacy compliance and i18n handling, but has several blocking issues: new signup behavior (age-attestation gate, combined-consent checkbox) is shipped without a feature flag, the Google-flow COPPA branch logic is inverted, the under-13 consent form records privacy acknowledgment without parental consent, and the combined_consent computed property silently clears te...

- **[HIGH]** New signup gate ships without a feature flag - The age_attested / combined_consent / ageAttestationRequired logic and the changed checkbox structure in register.hbs are new user-facing behavior gated only on showCoppaConsent, which already existed. AAC users completing a signup mid-session after a deploy will hit a structurally different form wi...
- **[HIGH]** Google-flow COPPA branch logic is inverted - In register.hbs the Google signup block renders the COPPA-parent form when {{#if this.showCoppaConsent}} and the 13+ combined form when {{else}}. This is the opposite of the standard signup path at line 290 and the opposite of what the PR description states. Under-13 Google signups will see the adul...
- **[HIGH]** Under-13 path records privacy_policy_acknowledged then deletes it, but the deletion guard is only in process_params - In user.rb, process_params stamps privacy_policy_acknowledged when terms_agree is true, then deletes it in the COPPA block. However grant_parental_consent! re-stamps it. If a developer calls process_params again after consent (e.g. during profile update with terms_agree still present in params), the...
- **[HIGH]** combined_consent setter clears model.terms_agree when unchecked mid-flow - The combined_consent computed setter does this.set('model.terms_agree', v) where v = !!value. If a user checks then unchecks the combined box, model.terms_agree is set to false. On a subsequent standard-path save (non-Google), the server receives terms_agree: false and process_boolean correctly igno...
- **[MED]** COPPA one-click consent page lacks any privacy notice (flagged by PR author, needs explicit resolution) - The PR description acknowledges that /parental_consent/complete grants consent on page load with no privacy content on the page, leaving the email link as the sole FTC 'direct notice.' This is a compliance gap under 16 C.F.R. § 312.5(b)(1), which requires notice 'directly to parents' before collecti...
- **[MED]** BAA enforcement gap ships as a policy assertion without blocking control - The beta notice states 'your organization should have a Business Associate Agreement (BAA) in place with us first; contact us to execute one.' No code blocks PHI entry without a BAA. Shipping this as-is means the app affirmatively represents a safeguard it does not enforce. The PR description asks f...
- **[MED]** privacy_beta_notice_body is a single monolithic i18n string containing legal copy with embedded logic - The 200+ word body paragraph is registered as one flat string in en.json and all locale files. This means (1) partial translation is impossible—translators must translate the entire paragraph atomically or not at all, (2) the BAA/COPPA/retention sub-clauses cannot be individually toggled or updated ...
- **[MED]** New i18n keys missing from en.json: register_consent_combined_prefix and register_consent_combined_suffix used in template but not added in this diff - Both register_consent_combined_prefix and register_consent_combined_suffix are referenced in register.hbs (lines ~302 and ~308 in the diff) and appear in en.json at lines 6603-6610, but they were present before this PR. However, the diff adds the *template references* to those keys in new branches w...
- **[LOW]** PRIVACY_POLICY_VERSION constant is defined in user.rb but not surfaced to the API or admin UI - Consent audit records reference User::PRIVACY_POLICY_VERSION but there is no migration to add a privacy_policy_version column, no API endpoint to expose what version a given user consented to, and no admin view. For GDPR Art. 7 accountability, the version needs to be queryable. Consider a separate p...
- **[LOW]** Spec file placed in spec/templates/ and uses Rails.root.join on frontend assets - spec/templates/signup_consent_contract_spec.rb reads app/frontend/app/templates/register.hbs and public/locales/en.json at test time. This will break if the spec suite runs in CI before the asset pipeline compiles or if the working directory differs. Place these in spec/integration/ or spec/complian...

### Adversary review
Several concerning edge cases around consent validation and version management could lead to incomplete or inconsistent privacy acknowledgments.

- **[MED]** Missing consent validation for Google SSO signup - The controller's googleSignupSubmitDisabled computed property includes age_attested check but doesn't include the googleSignupTerms check for users under 13 (showCoppaConsent true). This creates a path where a user could potentially submit without checking the terms/privacy checkbox if they're under...
- **[MED]** Race condition in parental consent token flow - When a parent completes the consent flow via grant_parental_consent!, the privacy_policy_acknowledged timestamp uses Time.now.utc.iso8601, but if multiple parents attempt to consent simultaneously for the same child (unlikely but possible), the last write wins without proper locking on the user reco...
- **[LOW]** Inconsistent version management for existing users - Existing users who agreed to terms before this PR will not have a privacy_policy_acknowledged record. This creates an inconsistent audit trail where some users have versioned privacy consent and others don't. A migration should be considered to backfill this data.
- **[LOW]** Missing privacy acknowledgment for manual user creation - The process_params method only records privacy_policy_acknowledged when terms_agree is true via form submission. Admin-created users or API-created users won't have this acknowledgment unless explicitly added, creating audit gaps.
- **[LOW]** No mechanism to handle policy version updates - When the privacy policy updates in the future, existing users won't be prompted to re-acknowledge the new version. The system records a one-time acknowledgment without tracking when users last viewed the policy or providing a way to require re-acknowledgment for material changes.

---

swahlquist:
**Dual-model review round (commit 667f3a4f5).** Reran both passes on the consent commit: Codex/DeepSeek senior-dev (via the MCP-stripped wrapper) and the Claude adversary red-team. Both confirmed no security or PII-leak defect (`settings` is `secure_serialize`d; `privacy_consent` is not exposed in `lib/json_api/user.rb`). The real findings were all about consent-record integrity. Dispositions:

**Fixed**
- **Strict boolean gate (Codex P1):** consent was recorded via `if params['terms_agree']`, but in Ruby the string `'false'` is truthy, so a crafted API request that declined would still record `terms_agreed` + `privacy_consent`. Now gated on `process_boolean(...)` (accepts `true`/`'1'`/`'true'` only). Added a `'false'`-string decline spec. The normal UI sends real booleans, so the happy path was never affected.
- **Defer minor consent to the parent (adversary M1):** under-13 signups no longer stamp `privacy_consent` on the child at signup; it is dropped in the COPPA block and re-recorded (`consented_by: 'parent'`) in `grant_parental_consent!` when the parent completes the email-token flow. Added a COPPA defer-to-parent spec.
- **Two contract guards (adversary M2, L):** since `privacy_consent` is recorded purely off `terms_agree`, one spec pins that the signup checkbox label still links the Privacy Policy (so a future label revert can't silently fabricate consent), and another pins `PRIVACY_POLICY_VERSION` to the policy's published "Last Updated" date (no silent version drift).

**Deferred (with rationale)**
- **Field naming - consent vs acknowledgment (adversary M3):** flagged for legal review. If counsel says this is acknowledgment-of-notice rather than GDPR consent, rename `privacy_consent` -> `privacy_policy_acknowledged`. No code change for now.
- **Timestamp format (Codex P2):** `terms_agreed` stays a Unix int (existing readers depend on it); `privacy_consent.agreed_at` stays ISO-8601 (clearer for a new audit record). Intentional.
- **Google terms cast strictness (Codex P3):** pre-existing, and the client only ever sends `true`/`false`. Out of scope.
- **i18n propagation + signup feature flag:** known; English is correct and other locales fall back gracefully.

Tests: contract guards 2/2; `process_params` block green except the unrelated pre-existing `user_spec.rb` LogSession failure (confirmed present on the clean branch); controller `create`/COPPA 34/0.

MelissaOneil:
Summary

This PR updates the registration pages with recent improvements and fixes.

Current Status

These changes have been implemented on this branch but are not yet aligned with the latest updates in traci/styling/styling-updates, which contains newer styling and structural changes affecting the registration flow.

Blocker

This PR is currently blocked until traci/styling/styling-updates is merged into the base branch.

Next Steps

Once the styling updates are merged:

Rebase this branch onto the updated staging branch
Re-validate registration flow end-to-end
Reapply or adjust fixes where needed based on any structural changes
Confirm UI consistency with the new styling system
Risk Consideration

There is a high likelihood of merge conflicts or behavioral changes in the registration flow due to overlapping updates in UI and layout logic. A final integration pass will be required after the dependency branch is merged.

swahlquist:
## Adversary red-team review (delta `667f3a4f5..9ac7362dd`)

Second-pass red-team on the two commits that landed after the last review round: the `privacy_consent` -> `privacy_policy_acknowledged` rename (`ca3376686`) and the locale propagation (`9ac7362dd`). Whole-worktree analysis, not just the diff.

**Verdict: Ship.** Two Low findings only.

### Confirmed clean (the top-risk items)
- **Rename completeness:** no stale `privacy_consent` survives anywhere in `*.rb / *.js / *.hbs / *.json`. `privacy_policy_acknowledged` is written only in `app/models/user.rb` (parent stamp, signup stamp, COPPA delete) and read nowhere else.
- **No JSON-API / log leak:** `privacy_policy_acknowledged` does not appear in `lib/json_api/`. `lib/json_api/user.rb:158` exposes only a derived boolean (`json['terms_agree'] = !!user.settings['terms_agreed']`); the acknowledgment hash with its timestamp + policy version is never serialized. `settings` stays `secure_serialize`d.
- **COPPA ordering:** the signup stamp runs in the `terms_agree` block; the delete runs later in the COPPA block gated by `!self.id && coppa_parental_consent_enabled? && wants_minor`. A non-COPPA user's stamp is never reached by the delete. A minor's stamp is deleted at signup and only re-recorded by `grant_parental_consent!` after token validation, with `acknowledged_by => 'parent'`. `process_boolean` returns true only for `true / '1' / 'true'`, so the Ruby-truthy string `'false'` records nothing.
- **i18n integrity:** all 13 locale files parse; the 3 new keys are present in every locale. `i18n.js:51-52` treats any `*** `-prefixed value as untranslated and falls back to the inline English default, so the `*** ` markers are never shown to users. en.json `register_consent_and` is exactly `" and "`, matching the inline default (no silent override).
- **Drift guard:** `Date.parse("June 9, 2026").iso8601 == "2026-06-09" == User::PRIVACY_POLICY_VERSION`; the contract spec assertion holds.

### [Low] Lint fix must keep `/privacy` inside the consent block
The CI `no-nested-interactive` failure is caused by the Privacy link being nested in the consent `<label>`. The contract spec and the legal coupling both rely on the single `terms_agree` checkbox visually presenting the Privacy Policy. Fix the lint by restructuring the checkbox/label association, **not** by moving the link out of the consent sentence.
> Addressed in `0b4fdd4fe`: checkbox + consent text are now siblings (no `<label>` wrapper) with the name supplied via `aria-labelledby`; both links stay inline. Lint passes; contract spec green.

### [Low] `register_consent_and` is whitespace-fragile
The rendered conjunction depends on the inline default, the en.json value, and the `*** ` fallback all agreeing on `" and "`. The existing `expect(en_json['register_consent_and']).to eq(' and ')` guard is the only thing pinning this against an i18n_generator "normalization." No action needed now; just don't let a regen trim it.

_Read-only review. Posted for the PR trail._

swahlquist:
## Senior-dev review (Codex / DeepSeek via OpenRouter, ZDR) — delta `667f3a4f5..9ac7362dd`

Second model on the same two commits (rename + i18n propagation). Raised two findings; both are context-blind false positives, verified below with evidence rather than waved off.

### P1 (claimed): rename breaks backward compatibility, needs a data migration
> "Users who signed up before this change will have `privacy_consent` in their settings, but the code now only writes/reads `privacy_policy_acknowledged`... A data migration or fallback logic is needed."

**False positive.** `privacy_consent` was never deployed. Evidence:
- `git grep privacy_consent origin/staging` and `origin/main` -> **0 results**; the key exists in neither branch.
- It was first introduced in `826d98c5` (a commit inside this same unmerged PR) and renamed three commits later in `ca337668`, both within PR #358.
- `git merge-base --is-ancestor 826d98c5 origin/staging` -> **not an ancestor** (never shipped).

So zero production users have `privacy_consent` in `settings`. Renaming a field that has never existed in any deployed environment needs no migration. The reviewer assumed it was a pre-existing production field; it is not.

### P2 (claimed): `*** ` placeholder strings will display to users
> "placeholder translations marked with '***'... should be translated rather than left as placeholder '***' strings, which will display as-is to users."

**False positive.** `app/frontend/app/utils/i18n.js:51-52` treats any value whose first 4 chars are `*** ` as untranslated and falls back to the inline English default. Users never see the literal markers. This is the normal `i18n_generator.rb` seed-then-translate workflow; the real translations land later via `WordData.translate_locale_batch(locale)`.

### Net
No actionable findings from this pass. The substantive issue on the PR is the CI `no-nested-interactive` lint failure (not raised by either model since neither runs the linter), now fixed in `0b4fdd4fe`.

_Posted for the PR trail._

PR #366, opened 2026-06-11
fix: close four AI-consent validation gaps from PR #266 (COPPA 1b)

swahlquist:
## What

Closes the four consent-logic gaps GitHub Copilot flagged on PR #266
(`feat/scot-ai-consent-foundation`, the Phase 1 model-layer foundation for AI
data-sharing consent, COPPA Final Rule 1b). They were marked "low confidence"
and merged unaddressed. Each was reproduced against current staging before any
change; **none was a false positive.** Frozen decisions D-02..D-05 are preserved
and raises stay machine tokens (Phase 3 owns user-facing copy).

## The four fixes

1. **Stale-version reactivation after a revoke** (`grant_ai_consent!`). The
   downgrade guard ran only while consent was active (`c['revoked_at'].blank?`).
   A user who granted v2, revoked it, then followed an OLD v1 grant link fell
   through and reactivated consent at v1 against superseded disclosures (the
   grant path deletes `revoked_at`). Now the older-than-prior refusal applies
   whether or not consent is currently revoked; the same-version no-op stays
   gated on active consent so the revoke/re-grant-at-same-version lifecycle still
   reactivates. **This is the compliance keystone.**

2. **Revoke audit trail lost `revoked_by` / `reason`** (`revoke_ai_consent!`).
   Those were written only to `settings`, which is deleted on the next re-grant,
   so the immutable trail permanently lost who revoked consent and why. Added
   `revoked_by` and `revoked_reason` to the revoke AuditEvent payload (keys
   mirror the settings keys, as grant's `granted_by` does).

3. **`nil` / non-numeric `disclosures_version`** (`grant_ai_consent!`). A nil
   version wrote a `granted_at` row `ai_consent_granted?` can never honor and
   that a later valid grant could not repair; the stale check was also raw `<`
   (lexicographic on strings: `"10" < "2"`). Added `ai_consent_normalize_version!`
   (Integer or all-digit String, `>= 1`, else `invalid_disclosures_version`),
   coerced up front and for the stored prior version so the comparison is numeric.

4. **Blank `granted_by`** (`grant_ai_consent!`). A parent-consent record could be
   written with no grantor identity. Now raises `invalid_granted_by` before
   `granted_at` is written.

## Test-infrastructure note (please read)

While verifying, I found the ai_consent spec suite was merged **red**: 7 examples
fail on clean staging because `AuditEvent.create!` (fired inside
`with_lock(requires_new: true)`) commits **outside** RSpec's per-example fixture
transaction, so audit rows accumulate across examples and break every
`expect(AuditEvent.count).to eq(0)` baseline (15 stale rows sat in the test DB at
rest). This is why Copilot's findings sat unverified. Fixed with a `before(:each)
{ AuditEvent.delete_all }` **scoped to the four consent describe blocks** (an
earlier global hook in `spec_helper` perturbed the chronically-red, order-
dependent `log_session_spec`, so I narrowed it). `log_session_spec` is unchanged
from staging (45 failures both before and after).

## Verification

```
DB_USER=scotw RAILS_ENV=test bundle exec rspec spec/models/user_spec.rb \
  -e 'ai_consent' -e 'AI consent'
# => 51 examples, 0 failures
```

New/extended specs cover all four fixes, including the revoked-then-stale-v1
no-op (issue #1) and the revoke audit-payload completeness + survival across a
re-grant (issue #2). Regression-checked the other AuditEvent-touching specs:
`log_session_spec` stays at 45 failures, identical to staging.

## Commits (atomic, each green on its own)

- `test:` clean AuditEvent between examples (later scoped to the consent specs)
- `fix:` reject blank granted_by (#4)
- `fix:` validate/coerce disclosures_version (#3)
- `fix:` hold stale-version no-op across a revoke (#1)
- `fix:` record revoked_by/revoked_reason in the revoke AuditEvent (#2)
- `test:` scope AuditEvent cleanup to the consent specs

Diff: `app/models/user.rb` (+~50), `spec/models/user_spec.rb` (+~130 specs),
`spec/spec_helper.rb` (unchanged net).

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -476,6 +476,9 @@ class User < ApplicationRecord
   # bare ids are normalized to global_id form before the self-grant check.
   def grant_ai_consent!(disclosures_version:, granted_by:, source:, ip: nil, user_agent: nil, granted_by_user_id: nil)
     raise ArgumentError, 'invalid_source' unless AI_CONSENT_SOURCES.include?(source)
+    # A parent-consent record without a grantor identity is not auditable. Reject
+    # blank granted_by before granted_at is written. Machine token; Phase 3 owns copy.
+    raise ArgumentError, 'invalid_granted_by' if granted_by.blank?
     if granted_by_user_id.present?
       granted_by_user_id = normalize_ai_consent_granted_by_user_id!(granted_by_user_id)
       raise ArgumentError, 'self_grant_forbidden' if granted_by_user_id == self.global_id
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -468,8 +468,9 @@ class User < ApplicationRecord
   # against the same user. D-04 / D-05.
   #
   # Raises ArgumentError on `invalid_source` (source not in AI_CONSENT_SOURCES),
-  # `invalid_granted_by_user_id` (malformed granted_by_user_id), and
-  # `self_grant_forbidden` (granted_by_user_id resolves to self.global_id). These
+  # `invalid_granted_by` (blank granted_by), `invalid_disclosures_version` (nil,
+  # non-numeric, or < 1), `invalid_granted_by_user_id` (malformed granted_by_user_id),
+  # and `self_grant_forbidden` (granted_by_user_id resolves to self.global_id). These
   # are stable machine tokens, not English prose - Phase 3 owns user-facing copy.
   #
   # `granted_by_user_id:` must be a global_id ("1_42") or bare numeric db id;
@@ -479,6 +480,11 @@ class User < ApplicationRecord
     # A parent-consent record without a grantor identity is not auditable. Reject
     # blank granted_by before granted_at is written. Machine token; Phase 3 owns copy.
     raise ArgumentError, 'invalid_granted_by' if granted_by.blank?
+    # Coerce to a positive Integer up front: a nil/garbage disclosures_version would
+    # otherwise write a granted_at row that ai_consent_granted? can never honor (and
+    # that a later valid grant can't repair), and a string version would make the
+    # stale-version check lexicographic. Raises 'invalid_disclosures_version'.
+    disclosures_version = ai_consent_normalize_version!(disclosures_version)
     if granted_by_user_id.present?
       granted_by_user_id = normalize_ai_consent_granted_by_user_id!(granted_by_user_id)
       raise ArgumentError, 'self_grant_forbidden' if granted_by_user_id == self.global_id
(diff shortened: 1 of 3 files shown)


swahlquist:

## Automated PR Review

| Pass | Verdict | Findings |
| --- | --- | --- |
| Senior-dev (Claude Sonnet 4.6) | `COMMENT` | 6 |

### Senior-dev review
The four compliance fixes are correct and well-tested. A small set of low-severity issues should be addressed before merging: the new helper is public instead of private, one test assertion makes a fragile AuditEvent.last assumption under concurrency, and the 'reactivate at same version after revoke' path may silently no-op rather than reactivate depending on how the stored version is encoded.

- **[MED]** ai_consent_normalize_version! is exposed as a public method - The helper is defined outside any private block in app/models/user.rb. Like normalize_ai_consent_granted_by_user_id!, it should be placed in the private section. A public mutation helper with a bang that raises ArgumentError is a leaky interface that could be called by controllers or serializers and...
- **[MED]** Stale-stored-string-version edge case may silently no-op a legitimate same-version reactivation - After the coercion block: prior_version = Integer(prior_version) if prior_version.is_a?(String) && prior_version.strip.match?(/\A\d+\z/). If the stored disclosures_version is a non-numeric string (e.g. data written by a buggy previous path), has_prior is set to false (because prior_version.is_a?(Int...
- **[MED]** revoke_ai_consent! passes revoked_by / revoked_reason regardless of nil - The fix unconditionally writes 'revoked_by' => revoked_by, 'revoked_reason' => reason to the AuditEvent payload. Both parameters default to nil. For a system-initiated revoke the keys will be present with nil values in every audit row. This is not wrong but it is inconsistent with how the grant path...
- **[LOW]** Test uses AuditEvent.last which is order-dependent under concurrent test runs - In records revoked_by and revoked_reason in the immutable AuditEvent payload the assertion is ae = AuditEvent.last. If another example or background job inserts an AuditEvent between the revoke call and the .last lookup (possible with requires_new: true subtransactions), the wrong record is inspecte...
- **[LOW]** ai_consent_normalize_version! accepts Float-coercible strings like "1e2" - The regex \A\d+\z correctly rejects floats, but Integer(raw.strip) will raise ArgumentError (not the custom token) for strings like " 1 " after strip because Integer is strict — actually strip is called so this is fine. However, consider that Integer('1.0') raises Ruby's own ArgumentError with a non...
- **[LOW]** No feature flag gating on the new validation behavior - The new invalid_granted_by and invalid_disclosures_version raises are model-layer and machine-token only, so they do not surface UI copy. However, if any existing integration test or production caller passes a nil granted_by (e.g. a system-generated grant with no parent identity), they will now rece...

### Adversary review
PR fixes logical COPPA consent gaps but introduces potential concurrency issues and fails to enforce consistent versioning constraints across the system.

- **[MED]** Potential concurrency gap during version coercion - The ai_consent_normalize_version! method is called before the transaction/with_lock, but the coercion of prior_version from settings happens inside the lock. If settings['ai_consent']['disclosures_version'] is a malformed string that passes the initial validation but fails the Integer conversion ins...
- **[MED]** Version coercion logic may be incomplete - The coercion logic for prior_version (lines 516-517) only handles all-digit strings. If prior_version was previously stored as a non-digit string (which shouldn't happen with new validation, but could exist from old data), the comparison disclosures_version < prior_version will fail (TypeError) when...
- **[LOW]** Missing version upper-bound validation - The method validates disclosures_version >= 1 but doesn't enforce an upper bound or sanity check (e.g., < 1_000_000). An extremely large integer could cause issues in comparisons or storage, and without validation, a typo or malicious input could set consent to version 999999999.
- **[LOW]** Test cleanup approach may mask integration issues - The before(:each) { AuditEvent.delete_all } in consent specs deletes audit events outside transaction boundaries, potentially hiding issues with AuditEvent accumulation in production where such cleanup doesn't occur. The underlying problem (AuditEvent.create! committing outside test transactions) re...
- **[LOW]** Inconsistent version type storage - While disclosures_version is coerced to Integer for comparison, it's stored as whatever type is passed (after coercion). The method ensures it's an Integer, but there's no guarantee other code won't write a string version. Should explicitly store as Integer to ensure consistency.

---

MelissaOneil (approved):

PR #367, opened 2026-06-11
test: isolate log_session_spec from orphaned test-DB rows (45 failures → 0)

swahlquist:
## Problem

`spec/models/log_session_spec.rb` had **45 chronically-failing examples** in a full-file run (`182 examples, 45 failures, 5 pending`). They pass or fail seemingly at random and have been a long-standing test-health drag.

## Root cause (verified with evidence)

Every one of the 45 failures is the **same signature**: a *global* count assertion returning an inflated value, e.g.

```
expect(LogSession.count).to eq(2)
  expected: 2
       got: 110
```

Direct query of `lingolinq-test` showed it carries **orphaned, already-committed rows from a prior interrupted run**:

| table | orphan rows | committed |
|-------|------------:|-----------|
| LogSession | 108 | 2026-05-31 |
| JobStash | 95 | 2026-05-31 |
| RemoteAction | 171 | 2026-05-31 |

Those rows inflate every unscoped `LogSession.count` / `LogMerger.count` assertion in the file (108→110, `needs_remote_push` 77, etc.).

Two facts rule out the "cross-example state pollution" theory:
- The live count held **steady at 108** across two full runs → the transactional fixtures work; there is **no ongoing leak**. The orphans are a one-time historical artifact.
- A counting example (`:3593`) **fails identically in isolation** (`got 110`). Examples that "pass in isolation" simply don't assert a global count (e.g. `:1360` checks reply content).

No product bug is involved — the merge/split/event-ordering assertions always passed; only the ambient row counts were wrong.

## Fix

A **file-scoped** `before(:each)` clears `LogSession`, `LogMerger`, and `JobStash` inside each example's transaction (rows are restored on rollback), giving every example a clean, deterministic baseline. This mirrors the existing patterns already in the suite:
- `spec_helper.rb` → `RemoteAction.delete_all` per example
- `log_session_spec.rb:~1490` → `JobStash.delete_all` in the `process_as_follow_on` block

Scoped to this file, **not** suite-global (a global `AuditEvent.delete_all` was previously shown to perturb this file's counts).

## Result

```
182 examples, 0 failures, 5 pending
```

45 → 0 failures, zero new failures, same 5 pre-existing pending placeholders. No production code changed.

## Note for reviewer

The 108 orphaned rows still sit in local/CI test DBs and could inflate unscoped `LogSession.count` assertions in *other* spec files. This PR makes `log_session_spec` immune. A separate housekeeping pass (`LogSession.delete_all` etc. against the test DB, or a broader teardown convention) could address the rest, but is out of scope here.


swahlquist:

## Automated PR Review

| Pass | Verdict | Findings |
| --- | --- | --- |

### Senior-dev review
Narrowly-scoped test-isolation fix: three delete_all calls inside the per-example RSpec transaction, plus a learnings doc entry. No production code is touched and the approach correctly mirrors the existing RemoteAction.delete_all / JobStash.delete_all patterns already in the suite.

- **[LOW]** LogMerger.delete_all may surprise any future examples that pre-seed LogMerger rows outside this file's before(:each) - The guard is file-scoped and runs inside the transactional rollback, so there is no persistent damage; but if a future test in this file seeds LogMerger rows in a before(:all) (non-transactional) context they would also be wiped. Low risk given the current suite, but worth a brief inline note that b...
- **[LOW]** LEARNINGS.md references a DB username ('scotw') in a diagnostic command - The snippet DB_USER=scotw … in the learnings doc hard-codes what appears to be a developer username. This is internal docs in an AGPLv3-public repo, so it is not a secrets leak, but it may confuse other contributors. Consider replacing with a placeholder like DB_USER=.

### Adversary review
Test fix introduces hidden dependency on transaction rollback to function correctly and fails to address root cause of persistent database pollution, potentially masking concurrency or isolation issues.

- **[MED]** Broken assumption: delete_all in before(:each) always runs inside a test transaction - The fix assumes RSpec always wraps each example in a database transaction and will rollback the deletes, but this may not hold if transactional fixtures are disabled, a different test adapter is used, or nested contexts modify transaction behavior, leading to unintended persistent data loss across t...
- **[LOW]** Incomplete isolation from orphaned rows in other models - The PR acknowledges orphaned rows in RemoteAction table (171 rows) but only clears LogSession, LogMerger, JobStash. Unscoped count assertions on RemoteAction elsewhere could still fail; inconsistency in cleaning approach may cause future confusion.
- **[LOW]** Race condition risk in parallel test execution - If CI or local runs use parallel_rspec, the delete_all in before(:each) may conflict with other threads/processes sharing the same test database, causing flaky failures not addressed by the fix.

---

PR #369, opened 2026-06-12
Traci/styling/styling updates

traciday:
# 🎨 Dashboard Design: Focused & Gentle layouts, drag-to-customize, and persistence

Reworks the home dashboard + the **Dashboard Design** ("Customize your dashboard")
flow into two named, user-selectable layouts, makes the arrangement fully
drag-customizable and **persistent**, and fixes a batch of rendering, responsive,
and i18n issues along the way. Staging is merged in, so this is up to date with
system-settings (#364), the signup consent flow (#358), and PHI scrubbing (#363).

---

## ✨ Two named layouts: `Focused` & `Gentle`

Renamed the internal layout keys so code matches the UI labels (no backward-compat
needed — pre-production):

- `balanced` → **`focused`** ("Focused View") and `dynamic` → **`gentle`** ("Gentle View")
- Applied across the persisted `preferences.dashboard_layout` value (incl. the
  `user.rb` default), the layout engine, ~all `.md-grid--layout-*` / `.md-card--speak-*`
  / `ll-layout-*` CSS, `data-gst-layout`, and the i18n key names.
- The engine branches on `layout === 'focused'`; everything else falls back to Gentle.

## 🧩 Layout engine (`utils/dashboard_sections.js`)

- **Focused View fills the row:** the grid column count now tracks the number of
  *visible* utility cards (Account / Create a Board / Reports / Edit Dashboard), so
  hiding one makes the rest expand to fill the row instead of leaving empty cells.
- **Constrained drag rules** (`reorderForFocused`): utility cards reorder only
  *within* their row; a single card can't be pulled into another row; whole rows
  reposition between rows.
- **Drag-aware responsive order:** the engine emits per-section `--ord-*` reading-order
  custom properties so the small-screen single-column stack can match the large-screen
  (and dragged) arrangement.
- Added `gridLayoutState` outputs: `columns` (inline grid columns for Focused) and
  `orderIndices`.

## 🖥️ Customize Dashboard modal (`getting-started-tour.js`)

- **Drag-to-reorder** in the live preview, honoring the Focused constraints.
- **Right-panel checklist mirrors the preview** — it re-sorts live as you drag.
- **Layout-aware labels:** the Speak item reads **"Let's Communicate"** on Focused
  (where Speak is the hero) and **"Speak Mode"** on Gentle.
- **Empty-state gating fixed:** only layout-*applicable* toggles count, so unchecking
  the visible elements now shows the "Nothing to display" overlay and disables **Done**
  *and* the **✕ close** button.
[...]

PR #370, opened 2026-06-12
Traci/styling/styling updates

traciday:
🎨 Styling continuation + Shepherd-on-Rails build fix
traci/styling/styling-updates → staging  ·  21 files · +682 / −97 (excluding generated assets)

Builds on PR #369 (already in staging) with one infrastructure fix, three new interactions, two leak/UX fixes, and a Focused-View theming refactor.

🚀 Build / Deploy

ember-shepherd is a v2 addon; ember-auto-import was code-splitting its services/tour into a content-hashed chunk that only Ember's index.html loads — so on the deployed app (which concatenates vendor.js + frontend.js through Sprockets) the module was missing → "Could not find module ember-shepherd/services/tour" and the tour service failed to build.

Disabled auto-import code-splitting + pinned a stable filename → everything folds into one predictable auto-import-app.js.
Wired it onto the Sprockets load path (render-build.sh copy + assert_js + //= require in both manifests), with a placeholder fallback so precompile never fails when no build ran.
Verified: the served application.js now contains the Shepherd implementation (was absent) and defines the module, loaded before the app code.
✨ New interactions
Drag-and-drop an image onto a Create-a-Board tile — create-board-new.js (+195), create-board-new.hbs
Drop an image file (or an image dragged from another browser tab) onto a button and it becomes that button's symbol in place. Reuses the board-detail upload pipeline (content_grabbers → save_image_preview), only accepts on labeled cells (with a drop-highlight cue), and bakes the URL (not the private image id) so the symbol renders and caches correctly on save.

New per-user preferences.board_dark_mode, shared between the board-detail dark toggle and the create-board preview. Persisted on toggle (server-whitelisted in user.rb); unset defaults differ by surface — board-detail → dark, create-board → light.

Clicking a row in the pinned board collection dims the panel with the same premium progress chrome the boards page uses. select_board now returns the route Transition so the overlay clears exactly when the board finishes loading (8s safety timeout for the fallback path).

🐛 Fixes
Brand sub-boards no longer leak as top-level tiles — utils/board-brands.js (new filterBrandRoots), board-collection.js, controllers/user/index.js
Brand-family page copies (e.g. CommuniKate "alcohol"/"bodyparts") often have no copy_id, so the copy_id-based filterRootBoards couldn't drop them — they showed as standalone tiles on the Mine & Public tabs and the speak-menu My Boards. New key-pattern classifier (test() matches the brand but key isn't the <brand>-<size> root shape → it's a sub-board) catches them; Public additionally clusters + dedupes by name. Non-brand boards untouched.

Speak was force-pinned to the top, making Speak↔Boards drag a no-op. It now packs at its saved order position like any other full-width block.

Communicator accounts (no supporter_role) get the board-detail options → Buttons section expanded on first open (Find a Button / Focus Words are their primary actions); supporters keep it collapsed; a one-time flag respects later manual collapses.

The board-picker page now surfaces the Cause and Effect tab with a disabled "Coming soon" button instead of hiding it, and skips the board query for it.

Focused-View theming tokens: a new semantic $focus-* palette (hero, chrome accents, glass surfaces, glossy buttons) derived from the brand colors via color-mix(... oklch). Every Focused surface — the "Let's Communicate" Speak hero, pill-nav, navbar trigger, Boards card, action buttons — was migrated off hardcoded literals onto these anchors, so the whole experience re-themes from a few variables.
Board-picker page full height: full-viewport background fix (top-level route, no .index.with_user background) plus the workspace card now filling height at ≤950px instead of collapsing to content.
Supporting styles for the new image-drop cell highlight and the board-collection loading overlay.

⚠️ Reviewer note: the new user-facing strings (coming_soon, opening_your_board, loading_boards_and_resources) ship with inline English defaults — run the i18n generator before release to populate the other locales. The board-picker ≤950px height fix is also pending a final visual confirm.


swahlquist:

## Automated PR Review

| Pass | Verdict | Findings |
| --- | --- | --- |
| Adversary (DeepSeek-V3.2, ZDR) | `CONCERNS` | 6 |

### Senior-dev review

_No findings._

### Adversary review
Several security and data integrity issues found, including potential HTML injection via dragged content, PII exposure in dropped image metadata, and improper image ownership/permissions.

- **[HIGH]** HTML injection via dragged 'text/html' content - _dropped_image_url extracts src attribute via regex from dragged HTML; attacker-controlled HTML could contain malicious scripts that execute when parsed. Should sanitize or use DOMParser.
- **[LOW]** Missing validation for dropped image URLs - _dropped_image_url accepts any URI from text/uri-list; could be non-image URLs or SSRF vectors. Should validate scheme and MIME type before upload.
- **[LOW]** PII leakage in dropped image metadata - save_image_preview's 'suggestion' param includes label text; if label contains PII (e.g., student name), it may be stored in image metadata. Should sanitize or truncate.
- **[LOW]** Insecure image ownership transition - Dropped private image (created by save_image_preview) is referenced only by URL; server's process_client_supplied_images may create a new public image, but original private image remains accessible to uploader. Could leak across districts.
- **[LOW]** Race condition in board dark mode preference save - Multiple rapid toggles of dark mode can trigger overlapping user.save() calls, risking stale preference state. Should debounce or queue.
- **[LOW]** Missing CSRF protection for user preference updates - Ember Data may not include CSRF token for PATCH to user; relies on Rails verify_authenticity_token, but if disabled, could be vulnerable.

---

PR #371, opened 2026-06-12
Traci/styling/styling updates

traciday:
🎨 Styling continuation + Shepherd-on-Rails build fix + security hardening
traci/styling/styling-updates → staging · 26 files · +909 / −100 (excludes generated assets)

Builds on PR #369 (already in staging) with: one deploy-infra fix, three new interactions, two display/UX fixes, a Focused-View theming refactor, and three security hardenings (two from a PR security review, one it missed).

🚀 Build / Deploy

ember-shepherd is a v2 addon; ember-auto-import was code-splitting its services/tour into a content-hashed chunk that only Ember's index.html loads. The Rails app concatenates vendor.js + frontend.js via Sprockets and never loaded that chunk → "Could not find module ember-shepherd/services/tour" and the tour service failed to build. Fix: disable code-splitting + pin a stable filename so everything folds into one auto-import-app.js, wire it onto the Sprockets load path (manifest require + render-build.sh copy + assert_js fallback). Verified the served application.js now contains the implementation.

✨ New interactions
Drag-and-drop an image onto a Create-a-Board tile — create-board-new.js (+195), create-board-new.hbs

New per-user preferences.board_dark_mode, shared between the board-detail dark toggle and the create-board preview. Server-whitelisted; unset defaults differ by surface (board-detail → dark, create-board → light).

🐛 Display / UX fixes

Board-picker page: full-viewport background fix (top-level route) + the workspace card now filling height at ≤950px instead of collapsing to content.
🔒 Security hardening
The single chokepoint for every server-side image fetch blocked localhost/127/0/decimal-IP but missed link-local 169.254.169.254 (cloud metadata) and RFC1918 private ranges, and didn't restrict the scheme. Now: restrict to http/https; block IP-literal loopback/private/link-local/CGN hosts via IPAddr. Public hosts unaffected; new spec coverage. (Residual noted: hostname→internal DNS-rebinding needs resolve-and-pin at the HTTP-client layer — filed follow-up.)

Stored-XSS / content-type — ButtonImage#process_params (app/models/button_image.rb, spec/models/button_image_spec.rb)
Image content_type was stored verbatim end-to-end (client → server → S3 Content-Type) with no allowlist. Now coerce any non-image/* type to image/png and drop a data: URI whose MIME isn't image/* (kills a data:text/html payload). SVG passes (legit OpenSymbols type) — script-sanitizing SVG on upload is a filed follow-up. New spec coverage.

Reviewer notes
New user-facing strings (coming_soon, opening_your_board, loading_boards_and_resources) ship with inline English defaults — run the i18n generator before release.
Board-picker ≤950px height fix is pending a final visual confirm.
Two security follow-ups filed (not in this PR): SVG script-sanitization on upload, and SSRF DNS-rebind resolve-and-pin.


swahlquist:

## Automated PR Review

| Pass | Verdict | Findings |
| --- | --- | --- |
| Adversary (DeepSeek-V3.2, ZDR) | `CONCERNS` | 7 |

### Senior-dev review

_No findings._

### Adversary review
Multiple security, privacy, and stability issues found: SSRF bypass via DNS rebinding remains unaddressed, SVG script injection risk persists, and several edge cases could break data isolation or leak PII.

- **[HIGH]** SSRF hardening incomplete, DNS rebinding attack vector still open - Uploader.sanitize_url blocks IP-literal private ranges but acknowledges DNS rebinding (hostname→internal) remains unpatched. An attacker could register a domain resolving to 169.254.169.254 or internal IP, bypassing hostname-based checks and accessing cloud metadata or internal services. Follow-up f...
- **[HIGH]** SVG upload script injection risk persists - ButtonImage#process_params now coerces non-image/* types to image/png but allows SVG (image/svg+xml) with no sanitization. Malicious SVG with  or event handlers could execute when rendered in browser. Follow-up noted but not resolved.
- **[MED]** Race condition in board dark mode preference saves - Both create-board-new.js and board-detail.js call user.save() on rapid toggle of board_dark_mode without coordination. Last write wins may persist wrong state; concurrent Ember Data saves could conflict. Should be throttled/debounced or use a single mutation endpoint.
- **[MED]** Image drop validation insufficient for data: URI MIME spoofing - _applyDroppedImageToLabel checks content_type from data: URI but extracts only first segment (split(/;/)[0]). An attacker could craft data:image/png;base64,text/html;... with dual MIME, bypassing client check. Server-side validation should be sole authority.
- **[LOW]** Board collection loading overlay timeout may leak navigation context - board-collection.js sets 8s safety timeout for loading overlay. If user navigates away before timeout, overlay clearance references destroyed component but may still trigger. While guarded by isDestroyed, timer could fire during teardown, potentially accessing stale DOM.
- **[LOW]** Drag-and-drop image URL extraction vulnerable to HTML injection - _dropped_image_url uses regex src\s*=\s*'" ([^'"]+) on raw HTML from dataTransfer. Malicious HTML with nested quotes or script could break parsing or extract incorrect URL. Should use DOM parser or stricter regex.
- **[LOW]** Missing content-type validation on uploaded image files - Client-side only checks f.type.match(/^image/), but file type can be spoofed. Server should re-validate actual file signature, not just uploaded Content-Type header.

---

PR #372, opened 2026-06-12
Traci/styling/styling updates

traciday:
🎨 Styling continuation · Shepherd-on-Rails build fix · security hardening
traci/styling/styling-updates → staging

Builds on PR #369 (already in staging) with a deploy-infra fix, three new interactions, two display/UX fixes, a Focused-View theming refactor, and a round of security hardening (from two PR security reviews).

🚀 Build / Deploy

ember-shepherd is a v2 addon; ember-auto-import was code-splitting its services/tour into a content-hashed chunk that only Ember's index.html loads. The Rails app concatenates vendor.js + frontend.js via Sprockets and never loaded that chunk → "Could not find module ember-shepherd/services/tour" and the tour service failed to build. Fix: disable code-splitting + pin a stable filename so everything folds into one auto-import-app.js, then wire it onto the Sprockets load path (manifest require + render-build.sh copy + assert_js placeholder fallback). Verified the served application.js now contains the implementation.

✨ New interactions
Drag-and-drop an image onto a Create-a-Board tile — create-board-new.js, create-board-new.hbs

A single per-user preferences.board_dark_mode (server-whitelisted) drives both the board-detail dark toggle and the create-board preview — light by default everywhere, dark only once the user turns it on. Writes are skip-if-unchanged to avoid redundant saves and a last-write-wins race.

🐛 Display / UX fixes

Board-picker page: full-viewport background fix (top-level route) + the workspace card now filling height at ≤950px instead of collapsing to content (visually confirmed).
🔒 Security hardening
The single chokepoint for every server-side image fetch blocked localhost/127/0/decimal-IP but missed link-local 169.254.169.254 (cloud metadata) and RFC1918 private ranges, and didn't restrict the scheme. Now: restrict to http/https; block IP-literal loopback/private/link-local/CGN hosts via IPAddr. Public hosts unaffected; new spec coverage. (Residual: hostname→internal DNS-rebinding — filed follow-up for resolve-and-pin at the HTTP-client layer.)

Image content-type / stored-XSS — ButtonImage#process_params (app/models/button_image.rb, spec/models/button_image_spec.rb)
Image content_type was stored verbatim end-to-end (client → server → S3 Content-Type) with no allowlist. Now: coerce any non-image/* type to image/png; drop a data: URI whose MIME isn't image/* (kills a data:text/html payload); and reject an SVG data: URI carrying active content (<script>/<foreignObject>/on…=/javascript:, decoding base64/percent-encoding first). Static symbol SVGs pass. The Create-a-Board drop also rejects non-image data: URIs client-side. New spec coverage (8 process_params examples). (Residual: remote http-SVG + inline CDN serving — filed follow-up.)

Reviewer notes
New user-facing strings (coming_soon, opening_your_board, loading_boards_and_resources) ship with inline English defaults — run the i18n generator before release.
Light/dark is now a single board_dark_mode preference defaulting light; create-board's preview toggle currently also writes it (symmetric) — can be made read-only if board-detail should be the sole writer.
Two security follow-ups filed (not in this PR): remote-SVG + inline-serving hardening (Content-Disposition: attachment), and SSRF DNS-rebind resolve-and-pin.


swahlquist:

## Automated PR Review

| Pass | Verdict | Findings |
| --- | --- | --- |
| Adversary (DeepSeek-V3.2, ZDR) | `CONCERNS` | 7 |

### Senior-dev review

_No findings._

### Adversary review
The PR introduces moderate security and correctness risks: an incomplete image file MIME check could bypass server-side validation; race conditions exist in user preference saves; and a missing permission check could leak private board data.

- **[MED]** Client-side image MIME check incomplete for dropped files - In create-board-new.js, _applyDroppedImageToLabel checks content_type from url.match(/^data:/). A dropped File is converted to a data URL via read_file → reader.readAsDataURL(), but this conversion may fail silently (e.g., large file). The fallback url could be undefined/null, skipping the content_t...
- **[MED]** Missing permission check in board-collection loading overlay - The new loading overlay in board-collection.js shows loading_board_name based on board.get('name'). If board is a private board the user shouldn't access (e.g., via a guessed key), the overlay reveals the board's name before the transition's permission check. The overlay should only appear after the...
- **[MED]** Race condition in user preference writes across tabs - Both board-detail.js (_persist_board_dark_mode) and create-board-new.js (togglePreviewMode) save user.preferences.board_dark_mode with a dirty-flag trick, but acknowledge a last-write-wins race across tabs. Since preferences are a single blob, concurrent saves from different tabs (or rapid toggles) ...
- **[LOW]** Brand sub-board filtering may incorrectly hide user-created brand-like boards - filterBrandRoots uses a key-pattern classifier to drop brand sub-boards. If a user creates a board with a name that matches the brand pattern (e.g., 'CommuniKate-*'), it could be incorrectly filtered from 'Mine' and 'Public' tabs. This is a correctness issue for edge-case user content.
- **[LOW]** Missing CSRF protection on user preference update endpoints - The user.save() calls in _persist_board_dark_mode and togglePreviewMode rely on Ember Data's default adapter, which likely includes CSRF tokens. However, there's no explicit verification that these endpoints (likely PATCH /api/v1/users/:id) enforce CSRF for non-GET requests. Given the app's mixed Ra...
- **[LOW]** Hardcoded timeout may not clear loading overlay on slow networks - In board-collection.js, runLater(_this, done, 8000) sets an 8-second safety timeout to clear the loading overlay. On very slow networks or with large boards, the board may still be loading after 8 seconds, causing the overlay to disappear prematurely while the board is still not ready, confusing the...
- **[LOW]** Image drop highlight CSS class may persist on rapid dragleave - cellDragLeave calls _clearCellDropHighlight on event.currentTarget. If the user drags quickly across multiple cells, dragleave may fire on the wrong element due to event bubbling, leaving a highlight on a cell that shouldn't have it.

---

PR #373, opened 2026-06-12
security(audit): scrub-failure fallback must not log raw e.message

swahlquist:
Follow-up to #368 (merged). One-line hardening of the fail-open path in `AuditEvent.log_command`.

## Problem
The scrub-failure fallback logged the raw `e.message` it was trying to scrub:

```ruby
detail = (PiiScrubber.scrub_log_line(e.message.to_s) rescue e.message.to_s).truncate(300)
```

The inline `rescue` exists to keep the audit path fail-open if `PiiScrubber` is unavailable (the "lib/ not autoloaded in a Resque worker" case the code comment calls out). But in exactly that case the fallback logs the unredacted `e.message` — and that message is "the one field that can echo input," per the same comment. So the one path where the scrubber is absent became the path that leaks PII into the log line and the Sentry alert. That defeats the purpose of #368 precisely when it matters.

## Fix
Fall back to a non-echoing placeholder; `e.class` still identifies the error:

```ruby
detail = (PiiScrubber.scrub_log_line(e.message.to_s) rescue '[unscrubbable]').truncate(300)
```

## Test
Adds a spec that stubs `PiiScrubber.scrub_log_line` to raise (simulating the constant being unavailable) and asserts the raw email never reaches the log line, only `[unscrubbable]`.

`spec/models/audit_event_spec.rb` is green (6 examples, 0 failures).

Surfaced by the senior-dev review pass on #368 (Medium finding).

### DIFF
--- a/app/models/audit_event.rb
+++ b/app/models/audit_event.rb
@@ -25,10 +25,11 @@ class AuditEvent < ApplicationRecord
       # a non-echoing placeholder, NOT the raw message: if the scrubber is absent
       # we must not leak the unredacted `e.message` we were trying to scrub. Include
       # only the scrubber exception class (never its message) so ops can distinguish
-      # NameError vs LoadError without echoing the persist error text.
+      # NameError vs LoadError without echoing the persist error text. LoadError is a
+      # ScriptError, not StandardError, so bare `rescue` would not catch it.
       detail = begin
         PiiScrubber.scrub_log_line(e.message.to_s).truncate(300)
-      rescue => scrub_err
+      rescue ScriptError, StandardError => scrub_err
         "[unscrubbable:#{scrub_err.class}]"
       end
       message = '[AuditEvent] failed to persist audit record for ' + user_key.to_s + ': ' + e.class.to_s + ': ' + detail
--- a/app/models/audit_event.rb
+++ b/app/models/audit_event.rb
@@ -21,8 +21,10 @@ class AuditEvent < ApplicationRecord
       # `user_key` is an opaque global_id. `e.message` is the one field that can
       # echo input (DB/encoding errors), so it is PII-scrubbed and length-capped.
       # The scrub is guarded so a missing constant (e.g. lib/ not autoloaded in a
-      # Resque worker) can never raise here and defeat fail-open.
-      detail = (PiiScrubber.scrub_log_line(e.message.to_s) rescue e.message.to_s).truncate(300)
+      # Resque worker) can never raise here and defeat fail-open. The fallback is
+      # a non-echoing placeholder, NOT the raw message: if the scrubber is absent
+      # we must not leak the unredacted `e.message` we were trying to scrub.
+      detail = (PiiScrubber.scrub_log_line(e.message.to_s) rescue '[unscrubbable]').truncate(300)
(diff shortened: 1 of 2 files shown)


swahlquist:

## Automated PR Review

| Pass | Verdict | Findings |
| --- | --- | --- |
| Senior-dev (Claude Sonnet 4.6) | `APPROVE` | 4 |
| Adversary (DeepSeek-V3.2, ZDR) | `CONCERNS` | 6 |

### Senior-dev review
The fix correctly replaces the raw-message fallback with a non-echoing placeholder and extends rescue to cover ScriptError/LoadError. The spec additions are thorough and the .with argument narrowing is an improvement.

- **[MED]** scrub_err.class.to_s not truncated; crafted class name could bloat log line - '[unscrubbable:#{scrub_err.class}]' is not length-capped. Ruby class names are bounded in practice, but a pathological stub or proxy object could return an arbitrarily long #class.to_s. Apply .truncate or limit to scrub_err.class.name.to_s (max ~80 chars) for defence-in-depth.
- **[MED]** Existing spec at line 36 still asserts raw 'boom' reaches the log - The test 'logs an error when save! raises' still expects the string 'boom' verbatim. After this PR that path goes through PiiScrubber.scrub_log_line, so the assertion is now load-order-dependent (passes only if PiiScrubber is loaded and leaves 'boom' unchanged). Consider updating to match the scrubb...
- **[LOW]** scrub_err is rescued but its class is still interpolated into a user-visible log string - scrub_err.class is safe (a constant name), but if someone later changes the interpolation to scrub_err.class.name it is still fine; just worth a code comment confirming .class (not .message) is intentional to aid future reviewers.
- **[LOW]** No truncate call on the new fallback branch - The happy path calls .truncate(300) on the scrubbed string; the fallback '[unscrubbable:…]' skips truncation. The value is short today, but keeping truncation uniform would prevent divergence if the format ever expands.

### Adversary review
Fix is an improvement but introduces a non-fail-open rescue and potential regressions.

- **[HIGH]** Fail-open property lost for non-ScriptError/StandardError - New rescue ScriptError, StandardError excludes other Exception subclasses like NoMemoryError, SystemStackError, SignalException. If scrubber raises those, the rescue won't catch and log line generation will raise, breaking fail-open.
- **[MED]** Non-standard exception rescue pattern - Rescuing ScriptError is unusual and may catch SyntaxError, NotImplementedError from scrubber bugs. Bare rescue was intentional to be fail-open; this change tightens scope but could hide real scrubber loading issues.
- **[MED]** Spec stubs break other test isolation - Spec stubs AuditEvent.new with specific arguments, but other tests may call AuditEvent.new differently (e.g., with different keys). This could cause those tests to fail if they rely on the unstubbed method.
- **[LOW]** No scrubber constant missing simulation - Test stubs scrubber method to raise, but doesn't test case where PiiScrubber constant is absent (NameError on constant). That's the original scenario the comment mentions.
- **[LOW]** Potential log format inconsistency - Message format changed: detail now includes scrub_err.class, but original message already includes e.class. This may break log parsing/monitoring tools expecting original format.
- **[LOW]** Placeholder may leak scrubber class info - Revealing scrubber exception class (e.g., LoadError) to logs may expose internal autoloading issues, though less severe than PII.

---

MelissaOneil (approved):

PR #374, opened 2026-06-12
Traci/styling/styling updates

traciday:
🎨 Styling continuation · Shepherd-on-Rails build fix · security hardening
traci/styling/styling-updates → staging

Builds on PR #369 (already in staging) with a deploy-infra fix, three new interactions, two display/UX fixes, a Focused-View theming refactor, a round of security hardening (from two PR security reviews), and i18n.

🚀 Build / Deploy

ember-shepherd is a v2 addon; ember-auto-import was code-splitting its services/tour into a content-hashed chunk that only Ember's index.html loads. The Rails app concatenates vendor.js + frontend.js via Sprockets and never loaded that chunk → "Could not find module ember-shepherd/services/tour" and the tour service failed to build. Fix: disable code-splitting + pin a stable filename so everything folds into one auto-import-app.js, then wire it onto the Sprockets load path (manifest require + render-build.sh copy + assert_js placeholder fallback). Verified the served application.js now contains the implementation.

✨ New interactions
Drag-and-drop an image onto a Create-a-Board tile — create-board-new.js, create-board-new.hbs

A single per-user preferences.board_dark_mode (server-whitelisted) drives both the board-detail dark toggle and the create-board preview — light by default everywhere, dark only once the user turns it on. Writes are skip-if-unchanged to avoid redundant saves and a last-write-wins race.

🐛 Display / UX fixes

Board-picker page: full-viewport background fix (top-level route) + the workspace card now filling height at ≤950px instead of collapsing to content (visually confirmed).
🔒 Security hardening
The single chokepoint for every server-side image fetch blocked localhost/127/0/decimal-IP but missed link-local 169.254.169.254 (cloud metadata) and RFC1918 private ranges, and didn't restrict the scheme. Now: restrict to http/https; block IP-literal loopback/private/link-local/CGN hosts via IPAddr. Public hosts unaffected; new spec coverage.

Image content-type / stored-XSS — ButtonImage#process_params (app/models/button_image.rb, spec/models/button_image_spec.rb)
Image content_type was stored verbatim end-to-end (client → server → S3 Content-Type) with no allowlist. Now: coerce any non-image/* type to image/png; drop a data: URI whose MIME isn't image/* (kills a data:text/html payload); and reject an SVG data: URI carrying active content (<script>/<foreignObject>/on…=/javascript:, decoding base64/percent-encoding first). Static symbol SVGs pass. The Create-a-Board drop also rejects non-image data: URIs client-side. New spec coverage (8 process_params examples).

Review triage documented in code — the two false positives (submenu "permission bypass" → cosmetic accordion; data-URI "File bypass" → Files become data: URLs and hit the same check) and the accepted-LOW cross-tab preference race are explained inline at the relevant sites.

🌐 i18n — public/locales/*.json (13 files)
Ran the string generator: added opening_your_board (board-collection loading overlay) across all 13 locales with English fallbacks (***-flagged for translators), and pruned 4 dead focused_* keys left over from the removed focused-home component. (coming_soon, loading_boards_and_resources were already registered.)

Reviewer notes / known follow-ups (out of scope, filed)
SSRF — DNS rebinding: the IP-literal block doesn't stop a hostname that resolves to an internal IP; the complete fix is resolve-and-pin at the HTTP-client layer (Typhoeus resolve:), which needs staging/DNS testing.
Remote SVG: the data: SVG path is hardened, but an SVG fetched from a remote http URL is re-served inline from the CDN; the clean fix is Content-Disposition: attachment on uploaded objects (changes the S3 upload contract → needs S3 testing).
Light/dark is now a single board_dark_mode preference; create-board's preview toggle currently also writes it (symmetric) — can be made read-only if board-detail should be the sole writer.


swahlquist:

## Automated PR Review

| Pass | Verdict | Findings |
| --- | --- | --- |

### Senior-dev review

_No findings._

### Adversary review
Multiple high/medium security concerns: SSRF fix incomplete (DNS rebinding still possible), SVG content-type bypass via remote URL, race conditions on user preference writes, potential XSS through SVG handling, and PII leakage risk with board collection loading indicators.

- **[HIGH]** SSRF Fix Incomplete - DNS Rebinding Risk - Uploader.sanitize_url blocks IP literals but accepts hostnames that could resolve to internal IPs via DNS rebinding. Cloud metadata endpoints (169.254.169.254) and internal services remain accessible through DNS rebinding attacks.
- **[HIGH]** SVG Content-Type Bypass via Remote URL - SVGs fetched from remote HTTP URLs are re-served inline from CDN without Content-Disposition: attachment. Malicious SVG with  or  could execute in user's browser context when viewed.
- **[LOW]** Race Condition on User Preference Writes - Multiple components (board-detail, create-board-new) write to user.preferences.board_dark_mode without optimistic locking. Cross-tab concurrent saves cause last-write-wins data loss, potentially exposing PII preference states.
- **[LOW]** Potential XSS via SVG Data URI Handling - ButtonImage#process_params checks data: URIs but remote SVG URLs bypass this check. SVG parsing may be incomplete - base64/percent decoding might miss nested malicious content in complex SVG structures.
- **[LOW]** Board Collection Loading Indicator Leaks PII - board_loading overlay shows loading_board_name which could reveal sensitive board names (e.g., medical/therapy boards) during loading, potentially visible in screenshots or screen sharing.

---

MelissaOneil (approved):

PR #375, opened 2026-06-12
Fix/melissa boards page dedup

MelissaOneil:


swahlquist:

## Automated PR Review

| Pass | Verdict | Findings |
| --- | --- | --- |
| Adversary (DeepSeek-V3.2, ZDR) | `CONCERNS` | 8 |

### Senior-dev review
The PR successfully fixes board-page deduplication with a feature-flagged owner-preference system and centralises the broken OpenSymbols URL rewrite. Several issues need attention before merge: a critical regex over-match in the Quick Core brand classifier, missing i18n helper wrapping for user-facing strings, and an unguarded board-list dedup that runs outside the feature flag.

- **[HIGH]** Quick Core root_re and test over-match incidental slugs like school/core-5 - /(^|\/)core-\d{2,}\b/i matches school/core-5 (one digit only, fine) but also any user/core-10 slug. The test fn adds /(^|\/)core-\d{2,}\b/i, so a board named 'Core 10' owned by any user gets classified as Quick Core and filtered by root_re, potentially hiding legitimate user boards. Min digits shoul...
- **[HIGH]** board_list dedup block not gated by boards_page_owner_dedup feature flag - The new dedup block in board_list (user/index.js ~L805) always runs dedupeBoardRows / filterBrandSetRootBoards regardless of the feature flag. Only the *representative selection* inside those calls is flag-gated. The filtering itself (hiding sub-boards, collapsing same-name rows) fires for all users...
- **[HIGH]** Truncation hint string not wrapped in i18n helper — raw English user-facing text - In available-boards-section.hbs the {{t ... key="boards_filter_truncated_hint"}} call does use t, but the fallback string "Showing the first %{count} matches. Keep typing to narrow results." is double-quoted English prose passed as the second positional arg to t. Confirm this matches the project's i...
- **[MED]** boardsPageSearchState depends on mineTagFolderDrillIn but computed key only lists filtered_results - The computed boardsPageSearchState checks this.get('mineTagFolderDrillIn') but its dependent keys do not include 'mineTagFolderDrillIn'. In Ember 3.28 the getter will not recompute when mineTagFolderDrillIn changes, so toggling folder drill-in won't update visible results until another dependency ch...
- **[MED]** boards_page_search_limit computed has no dependent keys — stale if constant ever changes - boards_page_search_limit: computed(function() { return BOARDS_PAGE_SEARCH_LIMIT; }) omits dependent keys, which is fine for a constant but suppresses Ember's static-analysis warning. More importantly the template interpolates this into the truncation hint at render time, but if BOARDS_PAGE_SEARCH_LI...
- **[MED]** rewriteBrokenSymbolUrl applied inside persistence.clean_url may strip user_token from rewritten URL - In both persistence files the rewrite runs *before* the user_token strip. If the rewritten URL happens to embed user_token= (unlikely but possible for signed alt URLs) the token would be stripped correctly, but the reverse—rewrite firing on an already-stripped URL—is fine. The main risk is the rewri...
- **[MED]** Search controller jump_sections computed does not declare appState as a dependent key - In search.js jump_sections calls boardsPagePreferUserNames(app_state) where app_state is a module-level import. If currentUser or feature_flags changes at runtime the computed will not recompute because neither is listed in computed('online_results', ...). The dedup preference could be stale after l...
- **[MED]** dedupeBoardRows orphan passthrough may silently re-introduce brand sub-board rows - Orphan rows are excluded from name-dedup and re-appended unconditionally. If cluster_orphans logic in board_list produces an orphan cluster whose synthetic board wraps a brand sub-board, that sub-board will bypass filterBrandSetRootBoards and appear in the grid. Add a guard or document that orphan r...
- **[LOW]** mineTagFolderSummaries changes cnt variable name to unfilteredCnt inconsistently - In the else-branch the rename from cnt to unfilteredCnt is cosmetic but the filtered branch still uses cnt as a local. This is harmless but creates a confusing asymmetry; the filtered-branch show variable is also re-checked after cnt is set, meaning a tag that matches by name but has zero matching b...
- **[LOW]** boardsPageSearchRows haystack built from search_string — confirm it includes board name and owner - search_string is a server-provided field; if a board record is missing it (e.g. freshly created via createRecord) the haystack is empty and the board is invisible to the live filter. A fallback of (b.get('name') || '') would make the filter robust for locally-created boards.
- **[LOW]** Symbol-URL rewrite comment duplicated in two places - The comment /* OpenSymbols removed arasaac/no_2.png (403 + XML); no.png is the live asset. */ appears identically in both services/persistence.js and utils/persistence.js as well as in symbol-url.js. The source-of-truth comment belongs only in symbol-url.js; the call sites need no inline explanation...
- **[LOW]** Test stubAppState manually implements .get() — will diverge from real EmberObject behaviour - The custom get function in stubAppState handles only two specific dotted paths. Any new code path that calls appState.get('feature_flags') (no dot) or another path will return undefined silently. Prefer EmberObject.create({ currentUser: ..., feature_flags: EmberObject.create({...}) }) so Ember's nat...

### Adversary review
Changes improve deduplication logic and fix broken image URLs but introduce potential performance regressions and an incomplete auth check for user-specific dedupe preferences.

- **[MED]** Potential performance regression in tag folder filtering - The new mineTagFolderSummaries iterates over boardsPageSearchRows for each tag on each filter change; with many tags/boards, this could cause UI lag. Should be memoized or computed more efficiently.
- **[MED]** Missing auth guard for boardsPagePreferUserNames - boardsPagePreferUserNames checks feature flag but not if currentUser exists; if user is logged out (null), appState.get('currentUser.user_name') may error, breaking dedupe.
- **[LOW]** Hardcoded search limit may frustrate users - BOARDS_PAGE_SEARCH_LIMIT = 50 is hardcoded; users with >50 matches may be confused when boards vanish, especially if filter is broad (e.g., 'a'). No UI to show all results.
- **[LOW]** Case-insensitive dedupe may hide intended variations - dedupeByName collapses names differing only in case (e.g., 'CommuniKate alcohol' vs 'CommuniKate Alcohol'); this could hide legitimately different boards if case conveys meaning.
- **[LOW]** Potential data leak in haystack building - boardsPageSearchRows builds haystack from board.search_string + tags; if search_string includes PII (e.g., student names), it's concatenated with tags but not exposed beyond client-side filter. Low risk.
- **[LOW]** Regex injection risk in old filter code - Removed _boardRowMatchesFilter used new RegExp(filter, 'i') without escaping; if filter contains regex special chars from user input, it could throw and break filtering. New code uses indexOf, fixing this.
- **[LOW]** Brand detection may misclassify user boards - Updated BRAND_FAMILIES test matches core-\d{2,} in key; a user board with key 'user/core-10' (two digits) could be incorrectly flagged as Quick Core root, hiding it from default grid.
- **[LOW]** Symbol URL rewrite may break other image paths - rewriteBrokenSymbolUrl globally rewrites '/no_2.png' to '/no.png'; if other symbol libraries use 'no_2.png' legitimately, they'll get wrong image. Low likelihood.

---

PR #376, opened 2026-06-12
Melissa/fix ssrf dns rebinding fix

MelissaOneil:
## Plain-language overview

This closes a security gap where someone could trick our servers into fetching private internal addresses (like cloud metadata or services on our own network) by supplying a normal-looking web address.

We already blocked obvious cases (addresses written as numbers like `169.254.169.254`). This change also covers tricks that use a regular hostname that resolves to an internal address, including DNS rebinding where the address changes between our check and the actual connection.

**What users should notice:** Almost nothing in normal use. Image imports, board imports, webhooks, and the image/audio proxy should keep working for legitimate public URLs. Callbacks or imports aimed at internal/private addresses will be rejected in production.

---

## Summary

- Add `SafeHttp` (`lib/safe_http.rb`) to resolve hostnames once, reject blocked IPs, and pin validated addresses via libcurl `CURLOPT_RESOLVE` so connect-time DNS cannot rebind to internal targets.
- Migrate all user-supplied URL fetch paths from `Typhoeus.*(Uploader.sanitize_url(...))` to `SafeHttp.get/head/post`.
- Route the search controller’s streaming image/audio proxy through `SafeHttp.build_typhoeus_request`.
- Keep `sanitize_url` as the fast, network-free string guard; DNS validation lives at fetch time only.

## Problem

`Uploader.sanitize_url` blocks IP literals in reserved ranges but never resolves hostnames. Call sites used `Typhoeus.get(Uploader.sanitize_url(url))`, so libcurl resolved DNS independently at connect time. Two bypasses remained:

1. **DNS-to-internal** — `evil.attacker.com` → A record `169.254.169.254` passes because the host string is not an IP literal.
2. **DNS rebinding (TOCTOU)** — even resolve-then-check loses if libcurl re-resolves at connect time (`followlocation: true` made redirects worse).

## Solution

| Layer | Role |
|-------|------|
| `sanitize_url` | Fast string checks (scheme, IP literals, encodings) |
| `SafeHttp` | Resolve once → validate all A/AAAA answers → pin via `CURLOPT_RESOLVE` → fetch with `followlocation: false` and manual redirect re-validation (max 5 hops) |

## Commits

1. `fix: close SSRF DNS rebinding with SafeHttp resolve-and-pin` — core module + migrated fetch paths
2. `fix: apply SafeHttp SSRF guard to search proxy fetches` — streaming proxy via `SafeHttp.build_typhoeus_request`

## Migrated call sites

- `app/models/concerns/uploadable.rb` — remote image/audio/video upload fetches
- `lib/uploader.rb` — `remote_zip`
- `lib/converters/utils.rb` — OBF/OBZ import from URL
- `lib/exporter.rb` — log import from URL
- `lib/tasks/openaac.rake` — vocabulary OBZ download
- `app/models/webhook.rb` — outbound webhook POSTs
- `app/controllers/api/search_controller.rb` — image/audio proxy (`#proxy` + button-set cache retry)

**Not migrated:** hardcoded outbound APIs (OpenSymbols, Giphy, Google TTS, S3 uploads, etc.) — not user-supplied SSRF surfaces.

## Behavior notes

- Webhooks to private IPs are blocked in non-dev environments (intentional for cloud SSRF defense).
- Development still skips blocked-IP checks (matches existing `sanitize_url` dev bypass).
- CDN multi-IP hostnames: all validated public IPs are pinned; libcurl picks one.

## Test plan

- [ ] `bundle exec rspec spec/lib/safe_http_spec.rb`
- [ ] `bundle exec rspec spec/lib/uploader_spec.rb -e "sanitize_url" -e "remote_zip"`
- [ ] `bundle exec rspec spec/lib/converters/utils_spec.rb -e "remote_to_boards"`
- [ ] `bundle exec rspec spec/controllers/api/search_controller_spec.rb -e "proxy"`
- [ ] Import an image from a public URL (symbol search, drag-drop, or board detail)
- [ ] Import a board from a public OBF/OBZ URL
- [ ] Confirm proxy still returns image/audio for a normal external URL (authenticated)
- [ ] Confirm `http://169.254.169.254/...` is rejected on import/proxy paths (non-dev)
- [ ] Smoke-test a webhook callback to a public HTTPS endpoint

### DIFF
--- a/app/controllers/api/search_controller.rb
+++ b/app/controllers/api/search_controller.rb
@@ -226,12 +226,19 @@ class Api::SearchController < ApplicationController
     
     error = nil
     s3_cache_miss = false
+    fetched_content_type = nil
+    fetched_body = nil
     begin
       request = safe_proxy_request(uri.to_s)
-      content_type, body = get_url_in_chunks(request)
-      if content_type == 'redirect'
-        request = safe_proxy_request(body)
-        content_type, body = get_url_in_chunks(request)
+      redirects = 0
+      loop do
+        fetched_content_type, fetched_body = get_url_in_chunks(request)
+        break unless fetched_content_type == 'redirect'
+
+        redirects += 1
+        raise BadFileError, 'too many redirects' if redirects > SafeHttp::MAX_REDIRECTS
+
+        request = safe_proxy_request(fetched_body)
       end
     rescue BadFileError => e
       error = e.message
@@ -252,16 +259,16 @@ class Api::SearchController < ApplicationController
     if s3_cache_miss && @api_user
       retried = attempt_button_set_regenerate(url)
       if retried
-        content_type, body = retried
+        fetched_content_type, fetched_body = retried
         error = nil
       end
     end
 
     if !error
-      str = "data:" + content_type
-      str += ";base64," + Base64.strict_encode64(body)
+      str = "data:" + fetched_content_type
+      str += ";base64," + Base64.strict_encode64(fetched_body)
       # Rails 7: render json: expects a hash, not a pre-encoded string
-      render json: {content_type: content_type, data: str}
+      render json: {content_type: fetched_content_type, data: str}
     else
       api_error 400, {error: error}
     end
--- a/app/controllers/api/search_controller.rb
+++ b/app/controllers/api/search_controller.rb
@@ -224,15 +224,13 @@ class Api::SearchController < ApplicationController
       end
     end
     
-    # TODO: add timeout for slow requests
-    request = Typhoeus::Request.new(uri.to_s, followlocation: true)
     error = nil
     s3_cache_miss = false
     begin
+      request = safe_proxy_request(uri.to_s)
       content_type, body = get_url_in_chunks(request)
       if content_type == 'redirect'
-        uri = URI.parse(body)
-        request = Typhoeus::Request.new(uri.to_s, followlocation: true)
+        request = safe_proxy_request(body)
         content_type, body = get_url_in_chunks(request)
       end
     rescue BadFileError => e
@@ -271,6 +269,12 @@ class Api::SearchController < ApplicationController
 
   private
 
+  def safe_proxy_request(url)
+    request = SafeHttp.build_typhoeus_request(url)
+    raise BadFileError, 'blocked or invalid URL' unless request
+    request
+  end
+
   def attempt_button_set_regenerate(url)
     match = url.to_s.match(%r{/button_set_cache/([^/]+)/})
     return nil unless match
@@ -304,7 +308,7 @@ class Api::SearchController < ApplicationController
     return nil unless result && result[:success] && result[:url]
     return nil if result[:url] == url
 
-    retry_request = Typhoeus::Request.new(result[:url], followlocation: true)
+    retry_request = safe_proxy_request(result[:url])
     content_type, body = get_url_in_chunks(retry_request)
     [content_type, body]
   rescue => e
--- a/app/models/concerns/uploadable.rb
+++ b/app/models/concerns/uploadable.rb
@@ -158,7 +158,7 @@ module Uploadable
     if self.url && self.settings && self.settings['content_type'] && self.settings['content_type'].match(/image\/svg/) && !self.settings['rasterized']
       self.settings['rasterized'] = 'pending'
       self.settings['rasterized_at'] = Time.now.iso8601
-      res = Typhoeus.head(Uploader.sanitize_url(URI.escape("#{self.url}.raster.png")), followlocation: true)
+      res = SafeHttp.head(URI.escape("#{self.url}.raster.png"))
       # check if there's already a .raster.png for the image (i.e. on opensymbols)
       if res.success?
         self.settings['rasterized'] = 'from_url'
@@ -220,12 +220,7 @@ module Uploadable
       file.write(Base64.strict_decode64(data))
     else
       self.settings['source_url'] = url if !rasterize
-      res = Typhoeus.get(Uploader.sanitize_url(URI.escape(url)), followlocation: true)
-      if res.headers['Location']
-        redirect_url = res.headers['Location']
-        redirect_url = redirect_url.sub(/\?/, "%3F") if redirect_url.match(/lessonpix\.com/) && redirect_url.match(/\?.*\.png/)
-        res = Typhoeus.get(Uploader.sanitize_url(URI.escape(redirect_url)))
-      end
+      res = SafeHttp.get(URI.escape(url))
       re = /^audio/
       re = /^image/ if file_type == 'images'
       re = /^video/ if file_type == 'videos'
--- a/app/models/webhook.rb
+++ b/app/models/webhook.rb
@@ -1,5 +1,7 @@
 require 'timeout'
 
+require 'safe_http'
+
 class Webhook < ApplicationRecord
   include Async
   include SecureSerialize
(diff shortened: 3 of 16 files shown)


swahlquist:

## Automated PR Review

| Pass | Verdict | Findings |
| --- | --- | --- |
| Adversary (DeepSeek-V3.2, ZDR) | `CONCERNS` | 12 |

### Senior-dev review
The SafeHttp module correctly closes the DNS-rebinding SSRF gap with resolve-and-pin; implementation quality is good. Several lower-severity issues need attention before merge: missing feature flag for behaviour-changing webhook blocking, a DNS timeout test that sleeps in CI, potential Ethon slist memory leak, and a few minor robustness points.

- **[HIGH]** Webhook SSRF block is a silent behaviour change with no feature flag - SafeHttp.post now rejects webhooks to private IPs in production with no feature flag. Existing district/hospital integrations pointing to internal relay servers will silently stop receiving callbacks with res.code=0 and a generic 'blocked' body. This must be behind a feature flag or at minimum emit ...
- **[MED]** DNS timeout spec sleeps DNS_RESOLVE_TIMEOUT+1 seconds in CI - spec/lib/safe_http_spec.rb 'returns nil when DNS resolution times out' calls sleep(6) inside a getaddrinfo stub. This will add ~6 s per test run and is fragile under parallel workers. Replace with allow(Timeout).to receive(:timeout).and_raise(Timeout::Error) or use Timecop.
- **[MED]** Ethon slist_free_all FFI callback may free on GC before Typhoeus uses it - build_resolve_slist returns an FFI::AutoPointer whose finalizer calls slist_free_all. If GC runs between request creation and libcurl consuming the slist, the pointer is freed while in use. Retain a reference on the Typhoeus::Request object (e.g. req.instance_variable_set) or switch to a manually ma...
- **[MED]** prepare_request calls Uploader.sanitize_url then re-parses; lessonpix escaping applied twice - escape_lessonpix_query is called both in prepare_request (on the sanitized URL) and in normalize_redirect_url. On a lessonpix redirect the ? is encoded twice (%253F). The initial call in prepare_request should be removed; escaping belongs only in the redirect normalizer.
- **[MED]** resolve_addresses returns nil on empty answer, not on resolution failure only - If a hostname legitimately has no A/AAAA records, ips.empty? returns nil and the request is silently blocked (same code path as a blocked IP). Distinguishing the two cases would help observability/logging.
- **[MED]** redirect_target uses response.present? from ActiveSupport but checks plain Hash headers - redirect_target calls location.present? which is fine, but also checks Link header and calls extracted.present?. In non-Rails contexts (rake tasks) present? may not be available on strings. Use !extracted.nil? && !extracted.empty? for safety.
- **[LOW]** No timeout default in SafeHttp.get/head — callers can hang indefinitely - The old Typhoeus.get call sites passed followlocation: true but no explicit timeout. SafeHttp passes caller opts through, so if no timeout: is supplied the request has no libcurl timeout. Add a DEFAULT_TIMEOUT constant (e.g. 30 s) merged into request_opts before callers forget.
- **[LOW]** OpenStruct used for failed_response — frozen in Ruby 3.2+ - OpenStruct is deprecated and slated for removal. failed_response returns one that has define_singleton_method called on it in attach_effective_url. Consider a simple Struct or dedicated FailedResponse class to avoid future breakage.
- **[LOW]** uploadable.rb still calls URI.escape which is deprecated/removed in Ruby 3.x - Both assert_raster and upload_to_remote pass URI.escape(url) to SafeHttp. URI.escape was removed in Ruby 3.x. Use URI::DEFAULT_PARSER.escape or CGI.escape on the path component only, or rely on SafeHttp/sanitize_url to normalise.
- **[LOW]** safe_proxy_request raises BadFileError with generic message obscuring the root cause - 'blocked or invalid URL' is shown to authenticated users in the proxy response. A slightly more informative message (e.g. distinguishing 'blocked address' from 'invalid URL') would aid support without leaking internals.
- **[LOW]** ProxyFakeRequest defined inside describe block — pollutes global constant namespace - spec/controllers/api/search_controller_spec.rb defines ProxyFakeRequest as a top-level class inside a describe block. Wrap it in a let or define as an anonymous class via Class.new to avoid constant-name collisions with other specs.
- **[LOW]** CGN_RANGE constant re-created on every blocked_ipaddr? call in sanitize_url path - sanitize_url in uploader.rb previously created IPAddr.new('100.64.0.0/10') inline on every call; this is now delegated to SafeHttp.blocked_address? which uses the module constant CGN_RANGE — that part is fixed. Confirm no remaining inline IPAddr.new('100.64.0.0/10') exist elsewhere after the refacto...

### Adversary review
The SSRF defense implementation has significant gaps: hostname resolution bypasses, IPv6 handling flaws, and incomplete request lifecycle coverage leave several attack vectors open.

- **[HIGH]** Initial URL sanitization bypass via IPv6 zone index in URI.escape - URI.escape(url) before SafeHttp.get/head/post passes hostnames with IPv6 zone indexes (fe80::1%eth0) to sanitize_url; IPAddr parsing strips the zone, but the original escaped URL may be passed to libcurl, potentially bypassing checks if libcurl treats zone differently.
- **[HIGH]** Missing hostname resolution in prepare_request for escaped URLs - SafeHttp.prepare_request calls Uploader.sanitize_url(url) which does not resolve hostnames. If attacker provides http://evil.attacker.com where DNS resolves to 169.254.169.254, sanitize_url passes it. The subsequent resolve_addresses is called, but if attacker controls DNS and returns a public IP in...
- **[HIGH]** CURLOPT_RESOLVE pinning may fail on IPv6 zone indexes - If host string includes zone index (fe80::1%eth0), format_pin_address produces '[fe80::1%eth0]' which libcurl may reject or interpret incorrectly, potentially falling back to regular DNS resolution, reintroducing rebinding risk.
- **[HIGH]** safe_proxy_request returns nil on blocked URL, causing unhandled exception - In search_controller.rb, safe_proxy_request raises BadFileError unless request is truthy. If build_typhoeus_request returns nil (blocked URL), BadFileError is raised with generic message, but the caller expects a request object. The error flow may expose internal details.
- **[MED]** IPv4 embedding detection incomplete for non-standard IPv6 forms - embedded_ipv4 only checks three specific byte patterns (NULL_12, NULL_10+FF_FF, NULL_8+FF_FF). Attacker could encode blocked IPv4 (e.g., 169.254.169.254) in other IPv6 embeddings like ::ffff:0:0:a9fe:a9fe (packed differently) that bypass detection.
- **[MED]** DNS resolution timeout may be too short for adversarial delays - DNS_RESOLVE_TIMEOUT = 5 seconds. Attacker controlling a slow DNS server could delay responses >5s to cause resolution failure, potentially making SafeHttp return failed_response, but some callers might treat that as 'blocked' vs 'network error' differently, causing inconsistent behavior.
- **[MED]** redirect_target does not validate redirect URLs before returning - absolute_redirect_url parses location but does not run it through sanitize_url or SafeHttp checks before returning. If Location header contains a malicious URI scheme (file://, gopher://) or encoded internal IP, it will be passed to next loop iteration's prepare_request, but the intermediate string ...
- **[MED]** skip_blocked_checks? only checks Rails.env.development? - In test or staging environments, blocked checks are active, but those environments may need to reach internal services for integration tests. This could break legitimate use cases and cause false positives.
- **[LOW]** Lessonpix query escaping may double-encode already-encoded queries - escape_lessonpix_query unconditionally replaces ? with %3F if pattern matches. If URL already contains %3F (e.g., from previous redirect), it will become %253F, breaking the URL.
- **[LOW]** build_resolve_slist may leak FFI::AutoPointer if Ethon::Curl.slist_append fails - If slist_append fails (returns nil), the FFI::AutoPointer may be initialized with nil pointer, but slist_free_all could be called on invalid pointer, causing undefined behavior or crash.
- **[LOW]** Missing Content-Type validation in proxy response - get_url_in_chunks returns content_type from remote server. If attacker controls remote server and returns content_type with newline or other injection characters, the data: URI construction could be corrupted, though Base64 encoding mitigates.
- **[LOW]** Webhook callback_attempts logs effective_url which may differ from user-supplied - effective_url is set from sanitized URL, but if redirects occur, it's updated. The logged URL may not match the original user input, complicating audit trails for SSRF attempts.

---

PR #377, opened 2026-06-13
Sanitize SVG uploads server-side to strip scriptable content.

MelissaOneil:
OpenSymbols legitimately serves image/svg+xml, so rejection is not an option. Add SvgSanitizer (Loofah XML scrubber), wire it into ButtonImage data: URIs and Uploadable#upload_to_remote, and force server-side upload for SVG to close the client direct-S3 bypass.
## Summary

- Adds `SvgSanitizer` (`lib/svg_sanitizer.rb`) — a Loofah XML scrubber that strips scriptable SVG content (`<script>`, `foreignObject`, `on*` handlers, dangerous URIs/SMIL) while preserving static symbol markup.
- Wires sanitization into upload chokepoints:
  - `ButtonImage#process_params` — sanitize `data:image/svg+xml` URIs on store (replaces reject-only `svg_data_uri_active_content?`)
  - `Uploadable#upload_to_remote` — sanitize HTTP-fetched and data-URI SVG bytes before S3 upload; fixes percent-encoded SVG data URI decoding
- Closes client direct-S3 bypass for SVG via `requires_server_sanitized_upload?` — SVG always routes through server-side `upload_to_remote` so bytes pass the sanitizer.
- OpenSymbols SVG by-reference URLs (`Uploader.valid_remote_url?`) remain trusted and unchanged.

## Why not reject all SVG?

OpenSymbols (mulberry, noun-project, twemoji) legitimately serves `image/svg+xml`. Blanket rejection would break symbol rendering.

## Test plan

- [ ] Symbol search → pick OpenSymbols SVG → confirm renders on board
- [ ] Skin tone variants on mulberry/noun-project SVGs
- [ ] SVG rasterization (`.raster.png`) still works
- [ ] OBF import with embedded SVG images

## Out of scope

- Post-upload S3 scan job, legacy S3 backfill, feature flag (security hardening always on)


swahlquist:

## Automated PR Review

| Pass | Verdict | Findings |
| --- | --- | --- |
| Adversary (DeepSeek-V3.2, ZDR) | `CONCERNS` | 11 |

### Senior-dev review
The SVG sanitizer is a solid security improvement with good defense-in-depth layering. Several correctness and robustness issues need attention before merge: a logic inversion bug in the upload_success guard, a missing Tempfile#close/unlink in replace_stored_upload_body!, a Loofah CONTINUE-after-remove issue that may skip siblings, and gaps in the verify_stored_s3_upload! path that could leave uns...

- **[HIGH]** Logic inversion in remote_uploader.rb upload_success guard - The condition unless !record.is_a?(ButtonImage) || record.verify_stored_s3_upload!(url) is a double-negation. For a ButtonImage it evaluates unless false || result = unless result, which is correct, BUT for any non-ButtonImage it evaluates unless true || ... = unless true = never executes, silently ...
- **[HIGH]** Tempfile leak in replace_stored_upload_body! — #close never called before #unlink - The ensure block calls file.unlink but never file.close. On Linux this leaves the file descriptor open until GC; on some platforms unlink before close silently fails. Use file.close; file.unlink or switch to Tempfile.create with a block.
- **[HIGH]** Loofah scrubber returns CONTINUE after node.remove — may skip next sibling - In SvgScrubber#scrub, after node.remove the method returns CONTINUE. Loofah's top-down traversal may advance past the now-shifted next sibling node after a removal. The correct idiom is to call node.remove then return STOP (or restructure as bottom-up). This could allow a  immediately before a secon...
- **[HIGH]** verify_stored_s3_upload! re-fetches full body from public S3 URL after Range-sniff shows non-SVG - If the Range sniff returns non-SVG bytes but the full fetch (fetch_uploaded_object_body) returns SVG (e.g. truncated sniff window for BOM-heavy SVG), the early return true on the non-SVG sniff path skips sanitization. The sniff-then-skip optimization is unsound; at minimum ensure sniff-positive trig...
- **[MED]** decode_data_uri_payload uses URI.decode_www_form_component — may double-decode - URI.decode_www_form_component decodes '+' as space in addition to percent sequences, which is correct for form data but wrong for SVG data URIs encoded with CGI.escape/ERB::Util.url_encode (which encode space as %20). Malformed input could survive a round-trip decode. Use CGI.unescape or URI.decode_...
- **[MED]** sanitize_stored_data_url returns data_url unchanged when !result[:changed] for declared SVG — skips content-type spoof - Line return data_url if !result[:changed] && data_url.to_s.match?(/\Adata:image\/svg\+xml/i) short-circuits even when the original data_url had a different encoding or a spoofed prefix detected by looks_like_svg?. If Nokogiri serialization changes nothing but the declared type was wrong, the origina...
- **[MED]** SvgScrubber node_count incremented before STOP check — off-by-one allows MAX_NODES+1 nodes - @​node_count is incremented then immediately checked against MAX_NODES. This means exactly MAX_NODES+1 nodes are processed before STOP is returned, because the last node at count == MAX_NODES+1 was already visited. Consider checking before increment or using >=.
- **[MED]** No feature flag — all existing BoardImage uploads silently re-routed through server; may affect latency SLAs - PR description explicitly opts out of a feature flag, citing 'security hardening always on'. However, forcing all SVG uploads through server-side processing adds a synchronous job step for every board symbol upload. AAC users are disrupted by slow saves. A flag would allow canary rollout. The PR des...
- **[MED]** parse_svg_document uses Nokogiri strict mode — rejects valid OpenSymbols SVGs with XML warnings - The mulberry-cat.svg fixture uses a DOCTYPE with internal entity declarations (&ns_svg;, &ns_xlink;). Nokogiri strict+nonet may raise errors on entity expansion or emit non-fatal errors that are not caught by the err.fatal? || err.error? filter. Test under strict mode with the actual fixture to conf...
- **[LOW]** reject_svg_upload alias exposes internal method name in public API - alias_method :reject_svg_upload, :record_upload_rejection makes reject_svg_upload public. If any external callers (e.g. specs) depend on the old name they will get the new behaviour (saves with validate:false) silently. The alias should be deprecated with a warning or removed if no callers remain.
- **[LOW]** URI.escape (deprecated since Ruby 2.7) still used in upload_to_remote - Line res = SafeHttp.get(URI.escape(url)) uses the deprecated URI.escape. This is pre-existing but the diff touches this method; prefer URI::RFC2396_Parser#escape or Addressable::URI.encode to avoid a Ruby 3.x removal breaking the fetch path.
- **[LOW]** record_upload_rejection saves settings['errored_pending_url'] with UPLOAD_FROM_STORED_DATA_URI sentinel fallback to data_uri - When source is UPLOAD_FROM_STORED_DATA_URI the errored_pending_url falls back to settings['data_uri'], which may be a multi-KB SVG data URI stored in the DB error field. This could cause DB row bloat for large SVGs. Cap or omit the data_uri fallback in the error field.

### Adversary review
SVG sanitization adds defense-in-depth but introduces incomplete coverage, new attack vectors, and potential data loss/DoS.

- **[HIGH]** S3 reupload bypass on verification failure - verify_stored_s3_upload! returns false when sanitization changes content but replace_stored_upload_body! fails (network/timeout), logging rejection. Upload remains accepted (upload_success renders 400), but malicious SVG persists on S3. This breaks the security guarantee.
- **[HIGH]** Incomplete sanitization: CSS style URLs and data: URIs - DANGEROUS_STYLE_PATTERN regex fails to catch url('data:text/html,...') or nested expression(...). Also, strip_external_reference_attributes only strips non-fragment hrefs; inline data: in href may remain if not caught by safe_url?, which blocks data: except image/* prefixes but does not sanitize nes...
- **[HIGH]** XML entity expansion DoS (billion laughs) via DOCTYPE - Loofah/Nokogiri with nonet may still parse internal DTD subsets; SVG fixtures include ]>. If attacker injects recursive entities, CPU/memory exhaustion occurs before scrubber runs.
- **[MED]** Content-Type spoofing via looks_like_svg? - Non-SVG file with <svg in first 4096 bytes (e.g., crafted PNG) triggers sanitization, potentially corrupting file. self.settings['content_type'] = 'image/svg+xml' in sanitize_stored_image_file! could mis-serve binary as SVG.
- **[MED]** Race condition: client direct S3 upload before verification - upload_success controller calls verify_stored_s3_upload! after HEAD succeeds, but malicious SVG may already be served from S3 between upload completion and verification. Window exists for XSS.
- **[MED]** Missing size validation for fetched HTTP SVGs - upload_to_remote checks body.bytesize > SvgSanitizer::MAX_BYTES after download, but large file could already be in memory/Tempfile. DoS via 5MB+ SVG fetch; also missing timeout on SafeHttp.get.
- **[MED]** Percent-encoded data URI decoding inconsistency - decode_image_data_uri_payload uses URI.decode_www_form_component for non-base64, but ButtonImage.sanitize_stored_data_url uses CGI.unescape in old method. Mismatch may cause double-decoding or loss.
- **[LOW]** SVG namespace stripping may break rendering - svg_root? rejects non-namespaced  (ns.nil?). Some valid SVGs may be rejected, causing failure('no_svg_root') and data loss.
- **[LOW]** MAX_NODES limit may reject complex legitimate symbols - Scrubber stops at 10,000 nodes, returning failure. Animated/compound symbols could exceed limit; silent rejection may break functionality.
- **[LOW]** Missing audit log for sanitization changes - Only Rails.logger.info logs stripped content; no persistent record (e.g., settings['sanitized_at']). Forensic investigation hindered.
- **[LOW]** Inconsistent content-type coercion for non-SVG images - self.settings['content_type'] = 'image/svg+xml' set in sanitize_stored_image_file! when looks_like_svg? true, but process_params coerces non-image/* to image/png. Could conflict.

---

PR #380, opened 2026-06-13
compliance(audit): migrate audit system to supported .claude/ layout (Phase 2)

swahlquist:
## What this is

**Phase 2 "Agent migration"** of the approved Audit/Compliance System Modernization. It moves the legacy, tooling-invisible audit system (top-level `skills/`, `subagents/`, `workflows/`) into the supported Claude Code layout under `.claude/`, driven by the Phase 1 findings register. No application code is touched; `audit-reports/FINDINGS.json` is unchanged.

## What ships

**Read-only finder layer**
- `.claude/agents/{privacy,infra,api,dependency}-auditor.md` — real Claude Code agents, read-only by construction (`tools: Read, Grep, Glob, Bash`, no Edit/Write), each preloading its checklist via `skills:`, `memory: project`, scoped `mcpServers`, and a PreToolUse write-blocker via per-agent `hooks:`.
- `.claude/hooks/audit-readonly-guard.sh` — PreToolUse guard. The real gap it closes is **Bash** (the allowlist already blocks Edit/Write, but Bash can still `>`, `sed -i`, `git commit`, `gh pr merge`, install, mutate infra/SQL, or write via interpreter eval). Scoped per-finder so it never affects the main session. ~50 allow/deny cases unit-tested.

**Checklists + orchestrator**
- `.claude/skills/{gdpr-ferpa,soc2-security,api-contract,dependency}-audit/` — ported domain checklists with the canonical `FINDINGS.json` schema embedded (finders emit `status: open` only, code-only evidence, no PII).
- `.claude/skills/audit-run/` — `/audit-run` orchestrator (`disable-model-invocation`), replacing `workflows/full-audit.md`: stamps the audited SHA + diff, fans out the finders in parallel, reconciles, runs the `adversary` agent as verifier, validates citations. Headline = open Critical/High counts, **not** a 0-100 score (decision 5.9.2).
- `scripts/audit-merge.rb` — deterministic register reconciler that enforces governance mechanically: never closes/downgrades, new findings land `open`, a re-surfaced `verified-closed` finding is flagged `regression: true` (never silently flipped), snippets are verified at the SHA before write, and malformed finder JSON is skipped not fatal.

**Cleanup**
- Retired from the fan-out: `ember-stabilization`, `rails-upgrade` (migration-era, shipped) and the `mvp-readiness` 0-100 score.
- DEPRECATED banners on the legacy `skills/`/`subagents/`/`workflows/` READMEs; repo `CLAUDE.md` and `audit-reports/README.md` updated to the new layout.

## Key decisions
1. **`.gitignore`:** `/.claude` was fully ignored, so the migrated system would have been untracked. Changed to `/.claude/*` + precise file-pattern negations to track only the audit subtree (`agents/*.md`, `skills/**/SKILL.md`, `hooks/*.sh`, `settings.json`); local settings, worktrees, and agent-memory stay ignored.
2. **MCP trust-tier annotations → Phase 3** (per plan §5.8), so Phase 2 stays entirely inside LingoLinq-AAC. Finder MCP is scoped via frontmatter (`deepwiki` only; `render` for infra with an explicit read-only tool allowlist + a noted Phase-3 residual).

## Adversary review (done, pre-merge, per plan §5.6 checkpoint 3)
An independent adversary red-team flagged real bypasses in the guard and register-reddening bugs in the reconciler — all fixed in the second commit and re-verified:
- **Critical:** interpreter eval/stdin/heredoc writes (`python -c`, `ruby -e`, `node -e`, `python3 -`, `ruby <<HEREDOC`, `gawk -i`, deno/bun eval) now denied.
- **Critical/High:** runtime (file-less) finding id aligned to `citation-check`'s `expected_id` (was producing id-mismatch FAILs).
- **High:** `git -c …`/`--no-pager` prefixes no longer evade the git blocker; `"CRITICAL"` no longer downgraded to `medium`; hallucinated/stale snippets can no longer overwrite valid evidence.
- **Medium/Low:** `env VAR=`/`npx`/`make` bypasses; per-input JSON rescue; whole-dir gitignore negation tightened; `grep INSERT` SQL false-positive fixed; stale README footer.

## Verification
- Guard denies every adversary repro; both scripts parse; `audit-merge` round-trips new/re-seen/regression/runtime/bad-snippet/malformed-JSON with **`citation-check` exit 0**; **real `FINDINGS.json` unchanged and green** (PASS 29 / FAIL 0 / SKIP 3).

## Out of scope (Phase 3)
compliance-officer agent + compliance calendar, Accessibility ACR, Compliance Posture Report, AI Governance Memo, `config/mcp-servers.json` trust-tier annotations, one-way Notion publish.


swahlquist:

## Automated PR Review

| Pass | Verdict | Findings |
| --- | --- | --- |
| Senior-dev (Claude Sonnet 4.6) | `COMMENT` | 10 |
| Adversary (DeepSeek-V3.2, ZDR) | `CONCERNS` | 12 |

### Senior-dev review
The PR is a well-structured tooling/compliance migration with no application code changes. Most concerns are low-severity advisory items about the guard's inherent denylist limitations, but a few concrete gaps in the guard regex, a missing chmod on the hook script, and one subtle Bash pattern need to be noted before merge.

- **[MED]** audit-readonly-guard.sh missing executable bit (chmod +x) - The hook is invoked via bash "$CLAUDE_PROJECT_DIR/.claude/hooks/audit-readonly-guard.sh" so the missing x-bit is not a hard blocker, but the .gitignore negation tracks *.sh and the file is committed without mode 755. If any future caller drops the explicit bash prefix, it will silently fail or error...
- **[MED]** Bash guard regex for sed -i can be bypassed with long-form flag --in-place - Pattern /\bsed\b[^|]*\s-[a-z]*i/ only catches the short flag. sed --in-place 's/x/y/' file is not matched. The interpreter-eval pattern does cover --inplace for interpreters but not for sed itself. Add |--in-place to the sed arm.
- **[MED]** curl POST-body flags (-d, --data, --data-binary, --json) not blocked - The guard denies -X POST|PUT|PATCH|DELETE and -o/-O, but curl -d @​file https://… or curl --json '{}' sends a write request with the default POST method without an explicit -X. A finder could exfiltrate data or trigger a webhook. Add a pattern for \bcurl\b[^|]*(-d|--data|--json|--data-binary|--data-...
- **[MED]** git_pre Regexp literal interpolated into string regex — Ruby warning and potential match gap - In the guard, git_pre is defined as a Regexp literal (/(?:…)*/) but then interpolated with #{} into another Regexp via %r{…#{git_pre}…}. Ruby interpolates a Regexp as its source string (works), but the surrounding %r{} delimiter and the inner / characters in git_pre may produce an unexpected pattern...
- **[MED]** audit-merge.rb not present in the diff but referenced throughout as critical path - Multiple files (SKILL.md, CLAUDE.md, audit-run/SKILL.md) reference scripts/audit-merge.rb as the deterministic reconciler, but it is absent from the diff. The PR description says it ships, so it may be an accidental omission. If omitted, /audit-run will hard-fail at Step 3 and no audit can complete.
- **[LOW]** .claude/skills/**/SKILL.md gitignore negation uses ** but git only supports one level of **/ - The .gitignore uses !/.claude/skills/*/SKILL.md (single wildcard, fine) but the PR description mentions skills/**/SKILL.md. Verify that the single-level pattern is intentional and that no skill is nested more than one level deep, otherwise its SKILL.md would be silently ignored.
- **[LOW]** infra-auditor lists mcp__render__list_key_value in disallowedTools but list_key_value is not a standard Render MCP tool name - The disallowedTools array includes mcp__render__get_key_value and mcp__render__list_key_value. If the actual Render MCP server exposes these under different names, the allowlist/denylist split is silently incomplete. Verify against the live Render MCP tool manifest before Phase 3 trust-tier annotati...
- **[LOW]** FINDINGS.json new entries reference SHA 8c0629583a0f… which is the PR branch commit, not a merged SHA - Self-referential findings anchored to a branch SHA will cause citation-check.rb to fail on any checkout that does not have that exact commit (e.g., after a squash-merge). Either anchor to a committed file at HEAD after merge, or use evidence.type: doc / runtime so citation-check SKIPs them, consiste...
- **[LOW]** Bash guard blocks npx globally, breaking legitimate read-only introspection (e.g. npx semver) - The pattern \b(npx|bunx|pnpx|make|just|…)\b denies all npx invocations. The dependency-auditor skill mentions it may run npm ls but also might reach for npx audit or npx semver for version checks. This is an intentional over-block per the comments, but worth documenting explicitly in the skill so fi...
- **[LOW]** privacy-auditor agent prompt hardcodes April 2026 as the remediation date for closed findings - The prompt says 'the PiiScrubber-bypass and the ungated Bugsnag/New-Relic findings from the April 2026 COPPA audit are fixed'. Hardcoding a date in an agent system prompt means this assertion becomes stale and misleading as the codebase evolves. Reference the FINDINGS.json ids instead so the adversa...

### Adversary review
The audit system migration introduces robust read-only controls and deterministic reconciliation, but several critical risks remain: the guard denylist can be bypassed, runtime evidence verification is incomplete, and the merge script may fail on malformed inputs despite claims.

- **[HIGH]** Guard denylist bypassable via novel obfuscation - The Bash guard is a denylist; novel command obfuscation (e.g., base64-encoded curl-to-exec, dynamically constructed strings) can bypass it. A determined LLM could circumvent, as the primary control is instruction-based. This breaks the read-only guarantee for infra-auditor's CLI access.
- **[HIGH]** Runtime evidence lacks mechanical verification - Runtime evidence (type:'runtime') is skipped by citation-check, relying on re-running live checks. No automated verification that observed state (e.g., Render service config) matches the recorded snippet, creating a trust gap. Adversary verification is also instruction-based.
- **[HIGH]** Merge script may crash on malformed JSON - scripts/audit-merge.rb claims to skip malformed finder JSON non-fatally, but if a finder returns non-JSON or a structurally invalid object, Ruby's JSON.parse will raise an exception, halting the entire audit. This could be exploited to hide findings.
- **[HIGH]** Infra-auditor MCP allowlist incomplete - Infra-auditor's disallowedTools blocks get_key_value/list_key_value to prevent secret reads, but other MCP read tools (e.g., render get_environment_variables) could also return secret values. Instruction-only control against echoing is insufficient for SOC2/HIPAA.
- **[LOW]** PII leakage via code snippets - Finders are told to cite file:line and column shape from fixtures/seeds with 'real-looking rows', but a snippet could still contain PII if the fixture includes realistic test data (names, emails). No mechanical scrubber for evidence snippets.
- **[LOW]** Concurrent audit runs corrupt register - If two /audit-run orchestrators run concurrently (e.g., in CI), they will race on reading/writing audit-reports/FINDINGS.json and the temp JSON files, causing corruption or loss of findings. No file locking or atomic updates.
- **[LOW]** Adversary verification is not fresh - Adversary agent runs 'with a fresh context', but if spawned via Agent tool within the same Claude Code session, it may inherit the original finder's memory or tool allowances, compromising independence.
- **[LOW]** .gitignore negation may miss future files - .gitignore uses precise file-pattern negations (e.g., !/.claude/agents/*.md). If new agent types are added (e.g., safety-auditor.md), they must match the pattern exactly or will be ignored. This is a maintenance trap.
- **[LOW]** Dependency-auditor CVE lookup incomplete - Dependency-auditor uses 'CVE patterns / advisory-known' but has no defined method to fetch live CVE data; relies on potentially stale deepwiki knowledge. Could miss recent critical vulnerabilities.
- **[LOW]** No audit of audit system integrity - The audit system itself (agents, hooks, scripts) is not audited for vulnerabilities (e.g., guard.sh injection, merge.rb logic bugs). A compromised audit system could suppress findings.
- **[LOW]** Missing idempotency for register updates - audit-merge.rb never closes findings, but if a finding is fixed and the finder stops emitting it, the register retains status:'open'. Requires manual closure, creating drift between code state and register.
- **[LOW]** Skill checklists contain copy-paste errors - Skills like gdpr-ferpa-audit/SKILL.md reference 'Scot' for closure but the actual process involves adversary verification first. This inconsistency could lead to operational confusion.

---

swahlquist:
Added reconcile commit `fb2c252ff` (legacy-dir cleanup) on top of the migration + adversary-fix commits.

**What it does**
- Deletes the legacy top-level `skills/`, `subagents/`, and `workflows/` dirs (21 files) now that the audit system lives under `.claude/`. These were left with DEPRECATED banners in the migration commit; per decision they are fully retired.
- Repoints the live references to the new layout:
  - `CLAUDE.md`: dropped the DEPRECATED table row; legacy dirs described as removed.
  - `.claude/skills/audit-run/SKILL.md`: dropped the "see their READMEs" pointer.
  - `docs/pre-merge-audit-checklist.md`: repointed `workflows/full-audit.md` and `subagents/` links to the `/audit-run` skill and `.claude/agents/`.

**Verification (re-run after this commit)**
- `ruby scripts/citation-check.rb audit-reports/FINDINGS.json` -> GREEN (29 PASS / 0 FAIL / 3 SKIP).
- `audit-readonly-guard.sh` denies `Edit`/`Write` and mutating Bash (`rm`, `sed -i`, file redirects, `git push`, `gcloud ... delete`) and allows reads (`grep`, `git log`).
- Both Ruby scripts parse; working tree clean.

Note: the adversary-review pass (commit `1396ebfc2`) is already part of this branch and closed real guard bypasses (env-prefix evasion, interpreter `-e`/heredoc writes, `git -c k=v push`) and hardened `audit-merge.rb` to verify every snippet at the audited SHA before writing, so a hallucinated finder snippet cannot redden the register.

swahlquist:
## Review remediation (commits `8c0629583` + `e4a5d114d`)

Triage of the automated dual-reviewer pass. Verdicts on the guard were verified by running the guard against each case, not assumed.

### Fixed
- **Render MCP read-only now frontmatter-enforced (was prose-only)** [both reviewers, HIGH/MED]. `infra-auditor` `tools:` allowlists only render read tools + deepwiki reads; `disallowedTools:` denies all render writes AND `get_key_value`/`list_key_value` (secret-value reads). Robust whether or not `tools:` gates `mcpServers` tools.
- **Guard fetch-and-exec bypass** [adversary HIGH]. Now blocks pipe-to-shell (`curl|bash`, `base64 -d|sh`, `cmd|python`), `eval`/`source`, and `sh -c`/`bash -c`. (Verified: these now DENY; `rm`/`sed -i`/redirect/`git push`/`gcloud delete` still DENY.)
- **Guard false positives** [senior-dev MED/LOW]. `git fetch` and `find -exec grep/cat` now ALLOWED (read-only); `-exec` still denied when it runs a mutating command.
- **deepwiki unreachable under tools allowlist**: added deepwiki read tools to all four finders' `tools:`.

### Recorded as `open` findings for Phase 3 (commit `e4a5d114d`, in `FINDINGS.json`)
`LL-97f9001bb4` guard denylist residual bypass; `LL-b5c30235d3` runtime/CLI evidence secret/PII leak (instruction-only); `LL-3483c28f3c` parallel-finder live-infra race; `LL-a2b45c2bcb` finder agent-memory cross-run state; `LL-5f0f4f52f8` audit system not self-audited.

### False positive / already resolved (verified)
- `curl` lowercase verb bypass: **FP** -- `curl -X post` is DENIED (`/i` covers the alternation).
- `psql -c 'SELECT'` blocked: **resolved** in `1396ebfc2` -- `psql SELECT`, `EXPLAIN ANALYZE`, `grep INSERT` all ALLOWED.
- `audit-merge.rb` "not in diff": **stale** -- it is on the branch.
- malformed finder JSON silently dropped: **resolved** in `1396ebfc2` (recorded in run summary).
- `.gitignore` could track a stray secret file: **resolved** in `1396ebfc2` (per-filetype negations).
- docs say legacy dirs removed but not deleted: **resolved** in `fb2c252ff` (deleted).

citation-check after changes: 34 PASS / 0 FAIL / 3 SKIP.

PR #382, opened 2026-06-14
feat(tracking): worldwide location + source/ad-key to HubSpot on signup

swahlquist:
## What

Wires already-captured trial-registration data through to HubSpot. `ExternalTracker` previously only sent city/state for **US** IPs and never sent the **source** (`document.referrer`) or **ad key** (`?ref=` param), even though both are captured and stored at signup.

- **Location worldwide:** drop the US-only gate so city + region (`subdivision`) populate for every country, plus a new `country` field so US vs international leads stay distinguishable. iplocate returns these fields globally.
- **Source:** `settings['referrer']` to the new `lingolinq_referrer` property, **truncated to scheme+host** so a referring URL carrying PII in its path/query never leaves the platform (origin attribution preserved).
- **Ad key:** `settings['ad_referrer']` to the new `lingolinq_ad_referrer` property.

## HubSpot setup (already done)

Created three custom contact properties under Contact Information: `lingolinq_referrer`, `lingolinq_ad_referrer`, `lingolinq_start_code`. The start-code line existed in code but had always been a silent no-op because its property did not exist; it now lands.

## Compliance

GDPR: this widens the **already consent-gated, supporter-only** HubSpot sync (`cookies_opted_out?`, `supporter_registration?`, `external_email_allowed?`, legitimate-interest basis) to include EU city/region. No new consent surface. Referrer is truncated to origin to keep PII in URLs from egressing. Confirm HubSpot DPA covers EU contacts (standard).

## Tests

`spec/lib/external_tracker_spec.rb`: updated existing HubSpot expectations, added non-US (Paris) location case, source/ad-key (with origin truncation) case, and unit tests for the `referrer_origin` helper. **16 examples, 0 failures.**

## Reviewers

- `/review-pr` senior-dev pass: Approve with comments; the one High (referrer PII egress) was addressed via origin truncation in this PR.
- `/adversary-review`: pending.

### DIFF
--- a/lib/external_tracker.rb
+++ b/lib/external_tracker.rb
@@ -87,8 +87,9 @@ module ExternalTracker
     if source.present?
       json[:properties] << {property: 'lingolinq_referrer', value: source}
     end
-    if user.settings['ad_referrer'].present?
-      json[:properties] << {property: 'lingolinq_ad_referrer', value: user.settings['ad_referrer']}
+    ad_key = ad_key_clean(user.settings['ad_referrer'])
+    if ad_key.present?
+      json[:properties] << {property: 'lingolinq_ad_referrer', value: ad_key}
     end
     if user && (user.settings['activations'] || []).length > 0
       json[:properties] << {
@@ -113,17 +114,29 @@ module ExternalTracker
   end
 
   # Reduce a referring URL to just its scheme+host (origin) so any PII in the
-  # path or query string is dropped before the value is sent to HubSpot.
-  # Returns nil when the value is blank or cannot be parsed into an http(s)
-  # origin, so the caller drops it rather than leaking an unparseable string.
+  # path, query string, or userinfo is dropped before the value is sent to
+  # HubSpot. Restricted to http(s) with a present host; returns nil for blank,
+  # non-web, hostless, or unparseable values so the caller drops it rather than
+  # leaking or polluting attribution with a non-origin string.
   def self.referrer_origin(referrer)
     return nil if referrer.blank?
     begin
-      uri = URI.parse(referrer)
-      return nil unless uri.scheme && uri.host
+      uri = URI.parse(referrer.to_s)
+      return nil unless ['http', 'https'].include?(uri.scheme) && uri.host.present?
       "#{uri.scheme}://#{uri.host}"
-    rescue URI::InvalidURIError
+    rescue URI::Error
       nil
     end
+  end
+
+  # The ad key comes straight from the ?ref= URL param (app.js), so it is fully
+  # attacker-controllable and unbounded. Only let through a short campaign-token
+  # shape (letters/digits/._-); anything else (URLs, PII, CSV/formula-injection
+  # payloads, oversized blobs) is dropped rather than egressed to HubSpot.
+  def self.ad_key_clean(ad_referrer)
+    return nil if ad_referrer.blank?
+    value = ad_referrer.to_s
+    return nil unless value.match?(/\A[A-Za-z0-9._\-]{1,64}\z/)
+    value
   end
 end
--- a/lib/external_tracker.rb
+++ b/lib/external_tracker.rb
@@ -29,12 +29,13 @@ module ExternalTracker
       end
     end
     email = user.settings['email']
-    city = nil
-    state = nil
-    if location && (location['country_code'] == 'USA' || location['country_code'] == 'US')
-      city = location['city']
-      state = location['subdivision']
-    end
+    # Capture geolocation for every country, not just the US. iplocate returns
+    # city / subdivision (region/state/province) / country globally, so non-US
+    # trials (largely GDPR-region prospects) now get city + region populated too.
+    # country lets HubSpot still tell US leads apart from international ones.
+    city = location && location['city']
+    state = location && location['subdivision']
+    country = location && location['country']
 
 
     acct = "Communicator Account"
@@ -71,10 +72,24 @@ module ExternalTracker
         {property: 'username', value: user.user_name},
         # {property: 'hs_language', value: 'en'},
         {property: 'state', value: state},
+        {property: 'country', value: country},
         {property: 'account_type', value: acct},
         {property: 'hs_legal_basis', value: 'Legitimate interest – prospect/lead'}
       ]
     }
+    # Marketing attribution captured at signup (see User#process: settings['referrer']
+    # is document.referrer, the "source"; settings['ad_referrer'] is the ?ref= URL
+    # param, the "ad key"). Sent to LingoLinq custom HubSpot contact properties.
+    # referrer is truncated to scheme+host before egress so a referring URL that
+    # happens to carry PII in its path/query (e.g. a webmail or tokenized link)
+    # does not leave the platform; the attribution signal (origin) is preserved.
+    source = referrer_origin(user.settings['referrer'])
+    if source.present?
+      json[:properties] << {property: 'lingolinq_referrer', value: source}
+    end
+    if user.settings['ad_referrer'].present?
+      json[:properties] << {property: 'lingolinq_ad_referrer', value: user.settings['ad_referrer']}
+    end
     if user && (user.settings['activations'] || []).length > 0
       json[:properties] << {
         property: 'lingolinq_start_code', value: user.settings['activations'].map{|a| a['code'] }.compact[-1]
@@ -95,5 +110,20 @@ module ExternalTracker
       Rails.logger.error("ExternalTracker: HubSpot contact sync failed for user #{user.global_id}: status=#{res.code}")
     end
     res.code
+  end
+
+  # Reduce a referring URL to just its scheme+host (origin) so any PII in the
+  # path or query string is dropped before the value is sent to HubSpot.
+  # Returns nil when the value is blank or cannot be parsed into an http(s)
+  # origin, so the caller drops it rather than leaking an unparseable string.
+  def self.referrer_origin(referrer)
+    return nil if referrer.blank?
+    begin
+      uri = URI.parse(referrer)
+      return nil unless uri.scheme && uri.host
+      "#{uri.scheme}://#{uri.host}"
+    rescue URI::InvalidURIError
+      nil
+    end
   end
 end
--- a/spec/lib/external_tracker_spec.rb
+++ b/spec/lib/external_tracker_spec.rb
@@ -269,6 +269,34 @@ describe ExternalTracker do
       res = ExternalTracker.persist_new_user(u.global_id)
       expect(res).to eq('201')
     end
+
+    it "should drop a hostile ad_referrer but still send a valid source" do
+      u = User.create
+      u.settings['email'] = 'testing@example.com'
+      u.settings['preferences'] ||= {}
+      u.settings['preferences']['registration_type'] = 'therapist'
+      u.settings['referrer'] = 'https://www.bing.com/search?q=aac'
+      u.settings['ad_referrer'] = "=cmd|'/c calc'!A1"
+      u.save
+      ENV['HUBSPOT_ACCESS_TOKEN'] = 'hubby'
+      expect(Typhoeus).to receive(:post).with("https://api.hubapi.com/contacts/v1/contact/", {
+        body: {properties: [
+          {property: 'email', value: 'testing@example.com' },
+          {property: 'firstname', value: 'No'},
+          {property: 'lastname', value: 'name'},
+          {property: 'city', value: nil},
+          {property: 'username', value: u.user_name},
+          {property: 'state', value: nil},
+          {property: 'country', value: nil},
+          {property: 'account_type', value: 'Therapist'},
+          {property: 'hs_legal_basis', value: 'Legitimate interest – prospect/lead'},
+          {property: 'lingolinq_referrer', value: 'https://www.bing.com'}
+        ]}.to_json,
+        headers: {'Content-Type' => 'application/json', "Authorization"=>"Bearer hubby"}
+      }).and_return(OpenStruct.new(code: '201'))
+      res = ExternalTracker.persist_new_user(u.global_id)
+      expect(res).to eq('201')
+    end
   end
 
   describe "referrer_origin" do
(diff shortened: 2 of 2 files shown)


swahlquist:

## Automated PR Review

| Pass | Verdict | Findings |
| --- | --- | --- |
| Senior-dev (Claude Sonnet 4.6) | `COMMENT` | 10 |
| Adversary (DeepSeek-V3.2, ZDR) | `CONCERNS` | 10 |

### Senior-dev review
The implementation is well-structured with good PII-scrubbing guardrails; a few issues need attention before merging—particularly the non-standard HubSpot country field name and a missing city property in the no-IP-data test path.

- **[HIGH]** HubSpot built-in country field may conflict with standard contact property - HubSpot's built-in contact property for country is country but it expects an ISO or English country name AND is a standard field that may conflict with custom properties. Sending a raw iplocate string like 'United States' or 'France' may fail or mis-map in HubSpot. Verify the property internal name ...
- **[MED]** GDPR: widening geolocation to EU contacts with no DPA/SCCs verification gate - Comment says 'Confirm HubSpot DPA covers EU contacts' but this is left as a TODO. The code ships now and will start sending EU city+region immediately. The PR should not merge until the legal confirmation is documented; add an explicit env-flag or feature flag so the EU path can be disabled if DPA r...
- **[MED]** referrer_origin silently drops port number, changing semantics for non-standard ports - 'http://example.com:8080/path' becomes 'http://example.com', dropping the port. The spec even asserts this. For attribution this is almost always fine but the comment promises 'scheme+host' which is not the same as 'origin' (origin includes port). At minimum the comment/spec should be consistent; id...
- **[MED]** No feature flag gates the new HubSpot properties being sent to EU/international leads - AAC users (and school-district HubSpot admins) will see city/region/country and referrer data appear on contacts without notice. Per review policy, new user-facing CRM data enrichment that widens data egress to third parties should be behind a feature flag so it can be rolled back if compliance issu...
- **[MED]** country sent as bare iplocate string; no normalisation means inconsistent HubSpot data - iplocate returns country names in English prose ('United States', 'France', 'Île-de-France'). If the HubSpot property is a dropdown or enum it will create spurious 'unknown option' entries. If it is free-text the data will be inconsistent across contacts. Consider mapping to ISO 3166-1 alpha-2.
- **[LOW]** Test for referrer/ad-key case stubs Typhoeus::get via no-op (IP lookup absent) - The referrer/ad-key integration test never sets up a Device with an ip_address, so the geo lookup branch is skipped and city/state/country are nil. This is fine but should be a second test that combines geo + attribution; the current test does not cover the code path where both are present simultane...
- **[LOW]** ad_key_clean regex disallows + and %20 which are common in Google/Meta UTM tokens - Campaign tokens from Google Ads or Meta often contain + (e.g. 'brand+search') or URL-encoded chars. The regex [A-Za-z0-9._\-]{1,64} will silently drop these legitimate values. Consider allowing + and % or documenting that UTM tokens must be re-encoded before storage.
- **[LOW]** Duplicate blank line between city/state/country assignment block - Lines 38-39 have a double blank line after the new geolocation block (minor style/lint issue that should be cleaned up to keep the diff readable).
- **[LOW]** lingolinq_start_code now-live property: no test for the activation-code path - The PR description notes start_code was previously a silent no-op. Now that the HubSpot property exists it will actually fire. There is no new test covering the activation-code branch; add a test to ensure the correct last code is sent and the property name is right.
- **[LOW]** Missing city property assertion in the no-device / no-geo path test - Existing test at line 122 adds country: nil but the original no-geo test was written before city was unconditional; verify the full property list is asserted in order so order-sensitive expect on Typhoeus body does not produce false positives.

### Adversary review
Several security and compliance issues identified: potential PII leakage in IP lookup data sent to third party, inadequate validation for new country field, insufficient referrer sanitization, and missing consent checks for new GDPR data elements.

- **[HIGH]** IP Geolocation may expose sensitive location data without consent - The code now sends city+region data globally to HubSpot. For EU users under GDPR, IP-based location data may constitute personal data requiring explicit consent, not just legitimate interest basis. Need review of lawful basis.
- **[HIGH]** Country field may contain invalid or malicious data - The location['country'] field from iplocate is used without validation. Could contain HTML, scripts, or overflow data that might affect HubSpot integration or downstream systems.
- **[MED]** Referrer origin extraction vulnerable to IDN homograph attacks - referrer_origin uses URI.parse without IDN normalization. Malicious referrer like 'https://аррӏе.com' (homograph) could bypass validation and send spoofed origin to HubSpot.
- **[MED]** Missing data retention/deletion compliance for new fields - No mechanism for users to request deletion of new location/country data from HubSpot. GDPR requires ability to delete all personal data, including newly added fields.
- **[MED]** Inconsistent PII handling between source and ad_key fields - referrer truncates to origin but ad_key_clean only validates format. Malicious actor could still inject PII via allowed characters (e.g., 'user.email@​domain.com_123').
- **[MED]** Missing audit trail for new PII data flows - No logging of what geolocation data or referrer information is sent to HubSpot. Creates compliance gap for data subject access requests and breach notification.
- **[LOW]** Country field may break HubSpot integration if too long - No length validation on country field from iplocate. Some country names could exceed HubSpot property limits causing API failure.
- **[LOW]** Race condition in geolocation lookup - If multiple devices register simultaneously for same user, concurrent iplocate API calls could cause rate limiting or inconsistent location data.
- **[LOW]** Missing validation for subdivision field - State/subdivision field from iplocate is used without sanitization. Could contain special characters that affect HubSpot integration or display.
- **[LOW]** Test coverage insufficient for edge cases - No tests for iplocate returning nil/malformed data, Unicode in location fields, or HubSpot API failures when sending new fields.

---

swahlquist:
## Dual-reviewer pass complete

**Senior-dev pass (`/review-pr`):** Approve with comments. The one High (referrer PII egress) was closed via origin truncation (`referrer_origin`).

**Adversary pass (`/adversary-review`):** ship-with-conditions. In-scope findings fixed in commit a29e03f:
- High — `ad_referrer` egressed raw/unbounded -> added `ad_key_clean` (campaign-token allowlist, 64-char cap; drops URLs/PII/injection payloads).
- Medium — `referrer_origin` accepted non-http schemes / empty host -> restricted to http(s) + present host.
- Medium — rescue too narrow -> broadened to `URI::Error`.

Out of scope, filed as #383:
- High (pre-existing) — CSV/formula injection in the org user report CSV.
- Medium (pre-existing) — `supporter_registration?` is a blocklist rather than an allowlist.

Open before merge: **compliance sign-off** that city-level geolocation of EU/GDPR-region prospects under default-opt-in legitimate-interest is acceptable (HubSpot DPA scope). Code is otherwise green: 22 examples, 0 failures.

swahlquist:
# Compliance Review: PR #382 HubSpot Trial Location + Source Tracking

**Date:** 2026-06-14
**Reviewer:** Compliance auditor (LingoLinq)
**PR:** #382, branch `scot/feat/trial-location-source-tracking`
**Files reviewed:** `lib/external_tracker.rb`, `spec/lib/external_tracker_spec.rb`, plus the guard methods in `app/models/user.rb`
**Data owner sign-off relied on:** Scot Wahlquist (CEO) owns the HubSpot DPA and has approved city-level geolocation for EU prospects under this flow.

---

## 1. Scope

This review covers the data flow changed by PR #382 in the `ExternalTracker.persist_new_user` background job, which fires on supporter (non-communicator) signup. The job:

- looks up the device IP against iplocate.io to derive coarse geolocation, and
- POSTs a marketing contact to HubSpot.

The PR makes three changes to what leaves the platform:

1. **Worldwide geolocation.** Removes the US-only gate so `city` + `subdivision` (region/state/province) + `country` are sent for every country, not just the US. Stated goal: capture city-level location for EU/GDPR-region prospects (`lib/external_tracker.rb:32-38`).
2. **Referrer origin.** Sends `lingolinq_referrer`, derived from `document.referrer` and truncated to scheme+host by `referrer_origin` before egress (`lib/external_tracker.rb:86-89, 116-130`).
3. **Ad key.** Sends `lingolinq_ad_referrer`, derived from the `?ref=` URL param and bounded by `ad_key_clean` to a campaign-token allowlist `[A-Za-z0-9._-]{1,64}` (`lib/external_tracker.rb:90-93, 132-141`).

**External processors in scope:** HubSpot (contact sync, covered by Scot-owned DPA). iplocate.io (receives the device IP for lookup).

**Gating (verified, unchanged by this PR):** the job returns early unless `external_email_allowed?` (excludes org-managed users), `supporter_registration?` (excludes communicators/unspecified), and not `cookies_opted_out?` (`lib/external_tracker.rb:12-18`; guards at `app/models/user.rb:118-144`).

**Data egress summary (per contact):** email, first/last name, username, city, region, country, account type, legal-basis stamp, referrer origin, ad key, and the latest start code. No AAC usage, board content, log data, or clinical/educational record is in this payload.

---

## 2. Per-framework findings

### FERPA (US school student records) - NOT IMPLICATED, with one residual

The egress path is gated to supporter accounts. Communicators (the role students occupy) are excluded by `supporter_registration?`, and org-managed users are excluded by `external_email_allowed?`. A district student record cannot reach HubSpot through the front door of this flow.

**Residual:** `supporter_registration?` is an exclude-by-blocklist (anything not `unspecified` or `communicator`), and `registration_type` is free-form client-supplied (`app/models/user.rb:129-137`). A self-registered user who supplies an unexpected `registration_type` (for example `student`) is treated as a supporter and slips the gate. This is a pre-existing design property, not introduced by PR #382, and it is already filed as the lower-priority item in issue #383. It does not egress a *district-managed* student (those are still caught by `external_email_allowed?`); the exposure is limited to a self-registered individual who both mislabels themselves and provides a real email. See residual risks below.

### HIPAA (US hospital patient data / PHI) - NOT IMPLICATED

No PHI is collected or transmitted by this flow. The payload is marketing-contact data for adult prospects. Patient communication content, clinical notes, and AAC usage data are not part of the payload and are not reachable from this code path. The same blocklist caveat as FERPA applies in theory (a patient self-registering and mislabeling themselves), but the payload still contains no clinical content, only contact + coarse-geo + attribution fields. No BAA question arises here because no PHI moves.

### GDPR (EU personal data) - DEFENSIBLE WITH CONDITIONS

This is the framework that actually bites, because the explicit purpose of the PR is to start capturing EU prospects' city-level location.

**Lawful basis.** The payload (business email, name, username, city/region/country, referrer origin, campaign key) is adult B2B-style prospect/marketing data. Legitimate interest (Art. 6(1)(f)) is a sound and defensible basis for this category, and the code already stamps the legal basis as legitimate interest, prospect/lead. City-level geo derived from IP is coarse (not precise location) and is proportionate to lead qualification and territory routing. With the data owner's approval and a HubSpot DPA in place, the processor-agreement and DPA-scope questions are answered yes.

**The consent-posture problem (the real issue).** The tracking gate is default-opt-in. The signup default template sets `'cookies' => true` (`app/models/user.rb:719`), and `cookies_opted_out?` only returns true on an *explicit* false (`app/models/user.rb:141-144`). A user who never answers a cookie/tracking prompt is a `nil`-or-`true` value and is therefore treated as consenting. Egress proceeds.

For legitimate interest as the basis, this is defensible *because LI does not require affirmative opt-in* - it requires a balancing test, transparency, and an easy opt-out, all of which are satisfiable here. The default-opt-in posture is the correct concern to raise, but it does not by itself sink the flow **provided you do not rely on consent (Art. 6(1)(a)) as the basis.** You cannot have it both ways: the `cookies` flag is framed as a consent signal, yet it defaults to on, which would be invalid if consent were the basis. Since the stamped basis is legitimate interest, treat the `cookies` flag as a *suppression/opt-out* control, not as a consent record, and make sure the privacy notice describes it that way.

**Where city-level geo for EU specifically needs care.** ePrivacy/PECR-style rules govern reading information from the device (cookies, the `document.referrer` the browser exposes, IP-derived location). City-level geolocation of an EU prospect under legitimate interest is acceptable when: (a) it is disclosed in the privacy notice at or before collection, (b) a documented Legitimate Interest Assessment (LIA) exists, and (c) opt-out is genuinely available. The code provides the opt-out mechanism; the open items are the notice and the LIA, which are policy artifacts outside this diff.

**Data subject rights.** Once a contact lands in HubSpot, GDPR access/erasure obligations attach there. This PR does not change deletion behavior. Note for the broader register: the Flusher user-deletion path does not currently clean up HubSpot, so an erasure request requires a manual HubSpot delete. That is a pre-existing gap, not introduced here, but it is now slightly higher stakes because more EU contacts will be created.

### COPPA (under-13) - NOT IMPLICATED BY THIS PAYLOAD, with the same blocklist residual

The flow targets supporter (adult) accounts and is gated away from communicators. The payload carries no child data when the gate holds. The residual is identical to the FERPA one: a self-registered minor who supplies a non-communicator `registration_type` would pass `supporter_registration?`. The mitigating factor is that the data egressed is still contact + coarse-geo + attribution, and a genuine under-13 self-registration is an edge case; the structural fix (inclusion allowlist) is tracked in #383.

---

## 3. Data minimization and retention

Assessed against the stated purpose (marketing attribution + lead territory routing):

- **city + region + country:** Proportionate. Coarse geo for territory routing and EU/US lead segmentation is a legitimate marketing-attribution need. Not precise location. Acceptable.
- **referrer (origin only):** Well-minimized. The `referrer_origin` truncation to scheme+host is a genuine privacy control - it strips path, query, and userinfo, so a referring URL that carries PII (a webmail link, a tokenized invite) leaks only the host. This is the right design and it is tested.
- **ad key:** Well-bounded. `ad_key_clean` enforces a `[A-Za-z0-9._-]{1,64}` allowlist, which blocks URLs, PII, oversized blobs, and CSV/formula-injection payloads from egressing. Good.

**Nothing in the egress payload is excessive for the stated purpose.** The minimization controls on the two new attribution fields are above the bar for this kind of flow.

**One caveat that is not in this diff:** `referrer_origin` and `ad_key_clean` sanitize only the *HubSpot egress* value. The raw `settings['referrer']` and `settings['ad_referrer']` are still stored unbounded at signup and are surfaced raw in the org user-report CSV, which has a separate CSV/formula-injection sink. That is the primary subject of issue #383 and is correctly scoped as a follow-up rather than a blocker for this PR, because this PR's egress path is clean.

---

## 4. Residual risks

| # | Risk | Severity | Status |
|---|------|----------|--------|
| R1 | `supporter_registration?` is exclude-by-blocklist over free-form client `registration_type`; a mislabeled self-registration (e.g. `student`) slips the student/minor gate. Org-managed students still excluded by `external_email_allowed?`. | Medium | Pre-existing; tracked in #383 (inclusion-allowlist fix). Accepted as follow-up. |
| R2 | Default-opt-in tracking posture: `cookies` defaults to `true`, `nil` treated as consenting. Defensible only if the basis stays legitimate interest, not consent. | Medium | Condition C1 below. |
| R3 | No LIA / privacy-notice disclosure for EU city-level geo verified in this review (policy artifacts, outside the diff). | Medium | Condition C2 below. |
| R4 | Erasure: HubSpot contacts are not cleaned up by the Flusher user-deletion path; GDPR erasure needs a manual HubSpot delete. Pre-existing; higher volume now. | Low-Medium | Condition C3 (follow-up). |
| R5 | Raw `referrer`/`ad_referrer` stored unbounded and surfaced in the org report CSV (separate sink). | Medium | Pre-existing; tracked in #383. Not a blocker for this egress path. |

---

## 5. Verdict

**SIGN OFF WITH CONDITIONS.**

The egress path itself is clean and well-minimized. The two new attribution fields have real, tested privacy controls (origin truncation, allowlist bounding). The geolocation expansion is proportionate coarse geo, the legal basis is correctly stamped, the processor DPA is owned and the EU geo expansion is approved by the data owner. FERPA, HIPAA, and COPPA are not implicated by this payload when the supporter gate holds. None of the residuals rise to a blocker.

Conditions (none block merge; C1 should be confirmed in writing, C2-C4 are tracked follow-ups):

- **C1 (confirm before merge, documentation only):** Confirm in writing that the lawful basis for this flow is **legitimate interest**, not consent, and that the `cookies` preference is treated as an opt-out/suppression control rather than a consent record. The default-opt-in posture is acceptable under LI but would be invalid under a consent basis. No code change required if the basis is LI.
- **C2 (follow-up, before or shortly after EU prospect volume ramps):** Ensure a Legitimate Interest Assessment exists and the privacy notice discloses IP-derived city-level geolocation and HubSpot processing for prospects, at or before collection. Policy artifact, not code.
- **C3 (follow-up):** Add HubSpot contact cleanup to the user-deletion / Flusher path so GDPR erasure requests are honored automatically rather than manually. Pre-existing gap, now slightly higher stakes.
- **C4 (follow-up, tracked):** Resolve issue #383 - switch `supporter_registration?` to an inclusion allowlist (closes R1) and fix the org-report CSV injection sink (closes R5). Not a blocker for this PR because this PR's egress path does not feed the CSV sink and does not egress student/patient/minor data when the gate holds.

This memo may be attached to PR #382.

swahlquist:
## C1 confirmed (lawful basis) — cleared

Confirmed by Scot Wahlquist (CEO, data owner, HubSpot DPA holder):

> The lawful basis for this HubSpot tracking flow is **legitimate interest** (Art. 6(1)(f)), not consent. The `cookies` preference is treated as an opt-out / suppression control, not as a consent record.

This satisfies the only pre-merge condition (C1) from the compliance review. The `hs_legal_basis: 'Legitimate interest – prospect/lead'` stamp in the code matches this basis; no code change required.

Remaining conditions are tracked follow-ups, none blocking this PR:
- C2 — privacy-notice disclosure + Legitimate Interest Assessment (policy artifacts)
- C3 — HubSpot contact cleanup on user deletion / GDPR erasure (filed separately)
- C4 — #383 (supporter allowlist gate + org-report CSV injection sink)

PR is cleared to merge on the compliance axis, pending normal code review.

swahlquist:
### Follow-up tracking (none block this PR)

All compliance-review conditions are now either cleared or ticketed:

- **C1** — lawful basis (legitimate interest) — ✅ confirmed above, cleared.
- **C2** — privacy-notice disclosure + Legitimate Interest Assessment → #385
- **C3** — HubSpot contact cleanup on user deletion (GDPR erasure) → #384
- **C4** — supporter allowlist gate + org-report CSV injection → #383

MelissaOneil (approved):

PR #391, opened 2026-06-14
ci: make secret-detection a blocking merge gate

swahlquist:
## Why

From the 2026-06-14 review-workflow evaluation: `secret-detection` was advisory (`continue-on-error: true` + gitleaks `|| true`), so a committed secret could never fail CI. Combined with staging having **no required status checks**, every quality gate was honor-system.

## What

- **Make `secret-detection` blocking** (drop `continue-on-error`, drop `|| true`).
- **Scope gitleaks to the PR/push delta**, not full history, so one historical false positive can't permanently block merges. `--redact` keeps matched values out of the public log.
- This is the code half of recommendation #1/#3. The repo-settings half (enabling required status checks on `staging` branch protection: `rspec`, `build-and-test`, `audit-artifacts-integrity`, `secret-detection`) is applied separately via the GitHub API.

## Scope note

Left `security-scan` (Brakeman / bundle-audit / npm-audit) advisory on purpose — more false-positive surface; can be tightened later. This PR is intentionally tight.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -171,9 +171,13 @@ jobs:
           npm audit --audit-level=high || true
         continue-on-error: true
 
+  # Blocking gate. Unlike the advisory security-scan job above, a leaked secret
+  # MUST fail CI. To stay safe as a required check we scan only the secrets a PR /
+  # push actually INTRODUCES (the commit delta), never the full history: one
+  # historical false positive must not block every future merge. Pair this with
+  # the staging branch-protection required-status-checks list so it gates merge.
   secret-detection:
     runs-on: ubuntu-latest
-    continue-on-error: true
 
     steps:
       - name: Checkout repository
@@ -181,7 +185,7 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Run gitleaks (CLI - no license required)
+      - name: Download gitleaks (CLI - no license required)
         run: |
           GITLEAKS_DOWNLOAD_URL="$(curl -sSfL \
             --retry 3 \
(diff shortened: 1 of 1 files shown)


swahlquist:

## Automated PR Review

| Pass | Verdict | Findings |
| --- | --- | --- |

### Senior-dev review
The blocking gate is a sound improvement but two concrete issues need attention before merge: the gitleaks binary is executed without any integrity verification, and the SHA interpolation in the shell if block is exploitable via crafted branch names.

- **[HIGH]** Unsanitized github.event.pull_request.base.sha interpolated directly into shell - github.event.pull_request.base.sha and head.sha are attacker-controlled strings (a PR from a fork can set them). Interpolating them with ${{ }} directly into a run: block allows shell injection. Use env vars instead: set BASE_SHA and HEAD_SHA as step env, then reference $BASE_SHA in the script.
- **[MED]** gitleaks binary downloaded and executed with no checksum or signature check - The binary is fetched from the latest GitHub release over HTTPS and executed directly. A compromised release or MitM (retry-all-errors widens the window) runs arbitrary code in CI. Pin a version and verify the SHA256 or cosign signature from the release assets before executing.
- **[LOW]** fetch-depth: 0 (full history) is no longer needed after scoping to delta - The PR comment says full-history fetches are avoided to prevent historical false-positives blocking merges, but fetch-depth: 0 still clones the entire history. Set fetch-depth to a small value (e.g. 50) or use a shallow fetch with the two SHAs to keep the intent consistent and reduce clone time.

### Adversary review
Improves security posture by making secret detection a mandatory gate; no apparent flaws in diff.

_No findings._

---

PR #393, opened 2026-06-15
fix(converters): accept ext_coughdrop_* namespace on OBF/OBZ import

swahlquist:
## Problem

Word/letter buttons (the "WordArt" buttons: keyboard keys, inflection words, plain-text buttons) imported from OpenBoards / OpenAAC vocabulary sets render as missing-image placeholders. Symbol buttons import fine; text buttons lose their label and button-level settings.

## Root cause

`Converters::LingoLinq` only read extension attributes from the `ext_lingolinq_*` namespace. OpenBoards / CoughDrop-origin OBF/OBZ files namespace those same attributes as `ext_coughdrop_*` (LingoLinq is a CoughDrop fork). Every `ext_lingolinq_*` read silently dropped the upstream value, so `hide_label`, `add_vocalization`, the board-level settings block, background, protected-board key, and image_url were all lost on import.

## Fix

A single helper, `Converters::LingoLinq.ext_attr(obj, param)`, that reads `ext_lingolinq_<param>` and falls back to `ext_coughdrop_<param>`. Applied uniformly across all 9 read sites (button EXT_PARAMS loop, image_url, translation rules, button apps, board settings block, background, board key, protected-board guard, and the from_external_nested preload).

- Prefers `ext_lingolinq_*` when both namespaces are present, so existing LingoLinq exports are unaffected (backward compatible).
- Returns `nil` only when both are absent, so falsey-but-present values are preserved.

## Testing

`DB_USER=scotw bundle exec rspec spec/lib/converters/lingo_linq_spec.rb` → 54 examples, 0 failures, 3 pending (pre-existing).

## Merge order

This is the root-cause fix the vocabulary-library data relies on to render correctly. Merge this **before** #scot/feat/system-vocab-library (the imported keyboard / Senner-Baud / Project Core boards will only render their word buttons correctly once this lands).

## Not covered here

Defect B (blue image placeholders from S3 re-host at import) is a separate issue, tracked for a follow-up after a real seed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #394, opened 2026-06-15
feat(system-boards): vocal-flair keyboard, Senner-Baud social pages, Project Core

swahlquist:
## What

Fills the gaps in the default vocabulary library shown in the sidebar and signup flow, owned by the `lingolinq` content account (not the legacy `example` user).

### 1. Keyboard in the sidebar
- Adds a committed `public/system-boards/keyboard.obz` — the Vocal Flair keyboard board (the copy of the Vocal Flair board, **not** the old plain programmatic keyboard), extracted from `vocal-flair-84-with-keyboard.obz`.
- `SystemSidebarBoards` now imports this OBZ first and re-keys its root to `<user>/keyboard`, falling back to the legacy `example/keyboard` copy and then the `KeyboardBoard` generator only if the file is missing.

### 2. Senner-Baud social pages in the signup library
- Adds committed `public/system-boards/senner-baud.obz` (43 boards, official Senner-Baud website export).
- New `SystemBoardSources.ensure_senner_baud!` mirrors `ensure_crisis_vocabulary!` — imports the OBZ, re-keys the root to `<user>/senner-baud`, names it "Senner-Baud Social Pages", makes it public.
- Added `senner-baud` to `SIGNUP_LIBRARY_SLUGS` so it appears in the signup library.
- `BetaSeed#ensure_system_content!` calls it alongside crisis-vocabulary with matching logging.

### 3. Project Core 36 imported (not in sidebar)
- `lib/tasks/openaac.rake` now handles single-board `.obf` files (Project Core / Universal Core 36 is one `.obf`, not an `.obz`): dispatches to `Converters::LingoLinq.from_obf` vs `from_obz` by extension.
- Added `project-core.obf` to the import list. Imported boards are public/gallery-visible but **not** added to `SIGNUP_LIBRARY_SLUGS` — Project Core is included but stays out of the sidebar, as requested.

## Compliance

All bundled symbols are from the opensymbols.org library, which LingoLinq has permission to use (confirmed by Scot). No third-party-model data flow.

## Testing

`DB_USER=scotw bundle exec rspec spec/lib/system_board_sources_spec.rb` → 4/4. All three imports verified end-to-end via `rails runner`: keyboard (66 buttons, real labels), Senner-Baud (43 boards), Project Core (`from_obf` path builds the button set).

## Depends on

Functionally relies on #393 (the `ext_coughdrop_*` converter fix) for the imported word/keyboard buttons to render correctly. Merge #393 first.


swahlquist:

## Automated PR Review

### Verdict: `COMMENT`

| --- | --- | --- |

**Senior-dev:** The implementation is solid and well-structured, but two concrete issues warrant attention: a hardcoded production S3 bucket name creates a GDPR/data-residency concern for European clients, and an idempotency gap in ensure_senner_baud! can save a root board with the wrong key before the re-key takes effect.

**Adversary:** Several edge cases and potential security issues found: missing SSL validation for S3 downloads, S3 bucket regex bypass, and concurrency concerns.

### Findings (7 unique)
- **[MED]** Hardcoded prod S3 bucket name leaks prod-environment coupling into all environments - SENNER_BAUD_OBZ_FALLBACK_BUCKET = 'lingolinq-prod-static' is baked into the constant and used in senner_baud_obz_urls. EU-hosted deployments (GDPR) or air-gapped hospital networks (HIPAA) will silently try to pull data from the US prod bucket when STATIC_S3_BUCKET is unset, creating a cross-region d... _(flagged by: senior-dev)_
- **[MED]** Root board key set before save may not persist if save! resets it - In ensure_senner_baud! (and import_obz_utility), board.key = key is set on the in-memory object, then save_without_post_processing is called in the loop, then save! is called again on root. If save! triggers a before_save callback that regenerates the key from user_name+slug, the re-key is silently ... _(flagged by: senior-dev)_
- **[MED]** S3 downloads lack SSL validation - SafeHttp.get downloads OBZ from S3 via HTTPS without verifying SSL certificates, enabling MITM attacks in hostile networks (e.g., hospital WiFi). _(flagged by: adversary)_
- **[LOW]** senner_baud_obz_urls regex accepts bucket names that S3 would reject - The regex \A[a-z0-9][a-z0-9.\-]{1,61}[a-z0-9]\z permits consecutive dots (e.g. a..b) and dot-hyphen sequences that AWS rejects, and allows total length up to 63 chars but the dot-separated label limit is not enforced. A misconfigured STATIC_S3_BUCKET passes validation and generates a URL that 404s s... _(flagged by: senior-dev)_
- **[LOW]** S3 bucket name regex can be bypassed - Regex for bucket names does not reject consecutive dots or dash at end, allowing 'lingolinq-prod-static.' or '..bucket' which would 404 but waste requests. _(flagged by: adversary)_
- **[LOW]** Concurrent import may cause partial visibility - When RecordNotUnique occurs, existing board is saved with public=true but sibling boards may stay unlisted, leading to inconsistent visibility. _(flagged by: adversary)_
- **[LOW]** Tempfile writes may exhaust disk space - OBZ/OBF writes to Tempfile without size checks; large files (e.g., 12 MB Senner-Baud) could fill /tmp on resource-constrained deployment. _(flagged by: adversary)_

---

PR #395, opened 2026-06-15
chore(seed): add lingolinq:rebuild_library task for clean library re-seed

swahlquist:
## What

Adds a guarded `rake lingolinq:rebuild_library` task that deletes all boards owned by the content user (`lingolinq`) and re-seeds the premade library, so converter/import fixes actually reach already-imported boards.

## Why

The `ensure_*` and `openaac:import_vocabularies` importers intentionally **skip boards that already exist** (to preserve identity). That means a plain re-seed does **not** apply fixes like #393 (`ext_coughdrop_*` namespace + wordart recovery) to boards imported before the fix. The only way to recover the dropped data is a fresh re-import.

## How

- `BetaSeed.delete_content_boards!(user)` — destroys all boards owned by the user, returns the count (unit-tested).
- `BetaSeed.rebuild_content_boards!(user)` — delete + `ensure_baseline!` + gallery import (sets `SEED_IMPORT_OPENAAC_VOCABULARIES`).
- `rake lingolinq:rebuild_library` — thin wrapper with guards:
  - `DRY_RUN=1` previews the count + first 50 keys, deletes nothing.
  - `REBUILD_CONFIRM=1` required to execute.
  - Hard **production block** (delete+recreate assigns new board IDs and breaks user copies / home-board references; prod needs an in-place refresh instead — see follow-up).
  - Runs `verify_beta_seed` after re-seeding.

## Safety

Staging-only by design. The prod path is deliberately blocked because delete+recreate is reference-breaking; an in-place refresh is the planned prod-safe follow-up.

## Test

`spec/lib/beta_seed_spec.rb`: 5 examples, 0 failures (3 existing + 2 new for `delete_content_boards!` scoping + nil guard).

## Usage

```bash
# preview
DRY_RUN=1 bundle exec rake lingolinq:rebuild_library
# execute (staging shell)
REBUILD_CONFIRM=1 bundle exec rake lingolinq:rebuild_library
```

### DIFF
--- a/lib/beta_seed.rb
+++ b/lib/beta_seed.rb
@@ -283,9 +283,13 @@ module BetaSeed
   def self.delete_content_boards!(user)
     raise ArgumentError, 'user required' unless user
     scope = Board.where(user_id: user.id)
-    count = scope.count
+    expected = scope.count
     scope.find_each(&:destroy)
-    count
+    remaining = Board.where(user_id: user.id).count
+    if remaining.positive?
+      raise "Delete incomplete: expected #{expected} board(s) removed, #{remaining} remain"
+    end
+    expected
   end
 
   # Required SEED_* env that ensure_baseline! will demand in staging/production.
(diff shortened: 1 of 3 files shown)


swahlquist:

## Automated PR Review

### Verdict: `COMMENT`

| --- | --- | --- |

**Senior-dev:** The rake task and BetaSeed helpers are well-guarded (dry-run, confirm flag, production block, referenced-user check, pre-flight env validation before any deletes). Code is clean and the new tests cover the key safety paths.

**Adversary:** The PR introduces a destructive library rebuild with multiple safety guards but has potential concurrency issues and unclear rollback scope.

### Findings (5 unique)
- **[MED]** Transaction wraps deletes but ensure_baseline! calls external services/network - rebuild_content_boards! runs delete_content_boards! + ensure_baseline! in a single ActiveRecord::Base.transaction. If ensure_baseline! makes HTTP calls (OpenAAC import) and they stall, the DB transaction stays open for the full duration, holding row locks and risking connection pool exhaustion on st... _(flagged by: senior-dev)_
- **[MED]** Potential transaction race condition on board counts - delete_content_boards! counts boards before delete, but concurrent writes could change count before find_each completes, causing misleading 'Delete incomplete' error. _(flagged by: adversary)_
- **[MED]** Rollback scope excludes side-effects from ensure_baseline! - Transaction wraps delete and ensure_baseline!, but ensure_baseline! may have side-effects outside the transaction (e.g., external API calls, file writes) that won't roll back on failure. _(flagged by: adversary)_
- **[LOW]** Rake task prints DB password-adjacent config via db_desc without sanitising host - db[:host] is printed to stdout; if the host string ever contains credentials (e.g. a postgres:// URI stored as host), it would appear in CI/shell logs. Low risk given typical Rails db config, but worth a nil/URI guard. _(flagged by: senior-dev)_
- **[LOW]** ENV restoration may conflict with other threads - with_temporary_env modifies global ENV; concurrent Rake tasks or web threads could see inconsistent env values, leading to unpredictable seed behavior. _(flagged by: adversary)_

---

MelissaOneil:
1. [MED] ENV mutation without restore
Added with_temporary_env, which saves prior values, applies overrides for the block, and restores (or deletes unset keys) in an ensure — even when ensure_baseline! raises.

2. [MED] No rollback if seed fails after delete
rebuild_content_boards! now wraps delete_content_boards! + ensure_baseline! in ActiveRecord::Base.transaction. A seed failure rolls back the deletes instead of leaving an empty library.

3. [MED] Referential integrity on direct destroy

Added the same content_boards_referenced_by_others guard inside rebuild_content_boards! (defense in depth — the rake task already checks this).
Kept destroy (needed for flush_related_records / proper cleanup); the transaction ensures a failed destroy chain rolls back.
4. [LOW] Missing post-delete count validation

delete_content_boards! raises if any boards remain after destroy.
rebuild_content_boards! compares deleted to the pre-delete expected_count.
Tests added
Three specs under .rebuild_content_boards!:

Transaction rollback + ENV restore on seed failure
Refusal when other users reference boards without ALLOW_REFERENCED_DELETE
Successful delete + re-seed when pre-flight checks pass
All three pass. The full file still has one pre-existing failure in .verify_beta_seed (duplicate core_word_list integration from test pollution), unrelated to these changes.

Note: If many users reference a board, flush_related_records may still enqueue a background job before rollback — that only applies when ALLOW_REFERENCED_DELETE=1 is explicitly set on a DB with real user references. The rake task guards against that in the normal staging path.

PR #396, opened 2026-06-16
compliance(audit): add read-only accessibility-auditor finder (WCAG 2.1 AA / EN 301 549)

swahlquist:
## What

Adds a fifth read-only finder, **`accessibility-auditor`**, to the `/audit-run` fan-out, paired with an **`accessibility-audit`** skill. This is the last Phase 3 build item of the Audit/Compliance System Modernization (the other Phase 3 items - compliance-officer, MCP trust-tier annotations, one-way Notion publish, posture/governance templates - already shipped in #381/#387).

The finder statically scans the Ember frontend (templates, components, SCSS) for **deterministic** WCAG 2.1 AA / EN 301 549 defects and emits register-shaped findings (`frameworks:["WCAG"]`, `type:"code"`) into `audit-reports/FINDINGS.json`, giving the existing ACR draft an automated, code-anchored feeder.

## Scope of checks (static only)
- Non-text content / accessible names (`alt`, `aria-label`, `aria-labelledby`) - **both** board grid render paths (`templates/board/index.hbs` AND the `fast_html` builder in `utils/button.js`)
- Info & relationships (semantic headings, label association), landmarks, heading order, `tabindex` misuse
- `:focus-visible` on interactive SCSS selectors
- `lang` attributes; i18n of accessible names (per the double-quote convention)
- Known low-contrast brand-token usage (by name; AA-safe targets defined in `_variables.scss`)

Runtime/AT behavior (live focus order, contrast *rendering*, screen-reader/switch/eye-gaze, axe/Lighthouse) is **explicitly out of scope** - that remains the human ACR / AT-testing pass.

## No Ruby/schema changes
`frameworkEnum` already includes `WCAG`, and `audit-merge.rb` / `citation-check.rb` handle arbitrary code findings.

## Verification
- Frontmatter byte-parity with `privacy-auditor` (read-only tools, PreToolUse guard + PostToolUse logger, `model: opus`, `memory: project`, deepwiki MCP).
- `/tmp` dry-run (never touched the real register): sample 1.4.3 finding -> `audit-merge.rb` `new=1` -> `citation-check.rb` **PASS**, `frameworks:["WCAG"]`, `status:"open"`.
- Orchestrator wiring: 5 finders in Step 2, `--in /tmp/finder-accessibility.json` in Step 3, never-diff-only cadence exception, refused-skip alarm in Step 5.

## Review
[...]

PR #398, opened 2026-06-16
fix(compliance): anchor calendar-render window to generatedDate, not Date.today

swahlquist:
## Problem

The `audit-artifacts-integrity` CI job runs `ruby scripts/compliance-calendar-render.rb --check`, which re-renders `compliance-calendar.md` from its JSON and string-compares against the committed render. The render computed its **"Due within 90 days or overdue"** window from `Date.today`, so the 90-day window slid with the wall clock. Items crossed the boundary as days passed, the committed render drifted from a fresh one, and `--check` went **red on every PR built on a date later than the calendar's last regeneration** - even when nobody touched the calendar. This was blocking unrelated PRs (surfaced on #396).

## Root cause

The render was an **impure function of the clock**, while a "render matches JSON" check must be a **pure function of the JSON**.

## Fix

Resolve the generation date once from `meta.generatedDate` (a JSON field) and use it for **both** the header and the due/overdue window:
- **Present + valid** generatedDate -> window anchored to it; render is now pure w.r.t. the JSON.
- **Present + malformed** -> **hard error** (a hand-edit mistake; silently falling back to the clock would reintroduce the drift this guards against).
- **Absent** -> falls back to `Date.today` (undated scratch JSON; preserves prior behavior).

Also harmonizes the header and window onto one date source (the header previously used `Time.now.utc` while the window used `Date.today`, which could disagree by a day across the UTC/local midnight boundary in the fallback).

## Why no `.md` change

The committed `compliance-calendar.md` was generated on its own `generatedDate` (2026-06-14), so `--check` now passes against the **existing** render with no regeneration, and it will not drift again until the JSON itself changes.

## Verification

- `--check` exits 0, re-render yields no `.md` diff.
- Malformed `generatedDate` (`2026-13-99`) exits 1 with a clear message.
- Absent `generatedDate` still renders via the today fallback.
- Sibling `compliance-notion-publish.rb --check` confirmed free of the same class of bug (it already masks its timestamp).

## Review

Reviewed by the Claude `reviewer` agent (Claude-only compliance toolchain; **not** routed to Codex/DeepSeek). No Critical/High. The Medium it raised (silent fallback on a malformed-but-present date) is fixed here via the hard-error path. One Low (no `spec/` coverage for this standalone script) is deliberately not addressed: there is no existing spec pattern for these scripts and the CI `--check` is the live guard; the pure-function property is now structural (`Date.today` fires only in the absent-key branch).

### DIFF
--- a/scripts/compliance-calendar-render.rb
+++ b/scripts/compliance-calendar-render.rb
@@ -28,7 +28,6 @@ end
 
 data = JSON.parse(File.read(json_path))
 meta = data['meta'] || {}
-generated = meta['generatedDate'] || Time.now.utc.strftime('%Y-%m-%d')
 owner = meta['owner'] || 'compliance-officer'
 
 def framework_label(fw)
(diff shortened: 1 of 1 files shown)


swahlquist:

## Automated PR Review

### Verdict: `APPROVE`

| --- | --- | --- |

**Senior-dev:** The fix correctly anchors the render window to meta.generatedDate, making the render a pure function of the JSON and eliminating CI drift. Logic, error handling, and fallback behavior are all sound.

**Adversary:** PR fixes critical drift issue by making render a pure function of JSON, addressing the root cause cleanly.

### Findings (0 unique)
_No findings._

---

PR #400, opened 2026-06-16
test(specs): stabilize repo-wide Worker.scheduled? rspec flake

swahlquist:
## Problem
The required `rspec` check flakes **repo-wide** on the `Worker.scheduled?` Resque specs and fails on `staging` itself, not just PRs - it just forced an admin-merge. Latest staging CI (run 27601259798) failed with:

```
expect(Worker.scheduled?(ExternalTracker, :persist_new_user, u.global_id)).to eq(true)
  expected: true
       got: false      # external_tracker_spec.rb:22 (the :16 example)
expect(Worker.scheduled?(Flusher, :flush_user_completely, u2.global_id, u2.user_name)).to eq(true)
  expected: true
       got: false      # flusher_spec.rb:409 (the :403 example)
```

Both are **false negatives** - the job IS scheduled, `scheduled?` just can't find it.

## Root cause (verified, not guessed)
`BoyBand::WorkerMethods#scheduled_for?` (`boy_band-0.1.16`) short-circuits:
```ruby
idx = queue_size(queue)        # cached size
return false if idx > 500      # "big queues mustn't be searched this way"
```
`queue_size` caches `sizeof/<queue>` in Redis with a **30s TTL** and only recomputes when the cached value is `0`. `Worker.flush_queues` (in `spec_helper.rb` `before(:each)`) empties the queue *lists* but **never clears that cache**. So once any earlier example inflates `sizeof/default` past 500, every `Worker.scheduled?` returns `false` for the next **30 wall-clock seconds**, regardless of the now-flushed queue. Whether a spec lands in that window varies run-to-run (machine speed, GC) -> a timing flake that passes on re-run.

`config.order = "defined"` + single-process CI + a fresh Redis service rule out order-dependence and cross-process races, leaving wall-clock-vs-TTL as the only nondeterminism. (`RedisInit.reset_queue_pressure_cache!` only nils an in-memory var; it does not touch any Redis size key.)

### Deterministic reproduction
```
redis-cli set "lingolinq-test:sizeof/default" 600
bundle exec rspec spec/lib/external_tracker_spec.rb:16 spec/lib/flusher_spec.rb:403
# => both fail with the exact "expected true, got false" CI signature
```
Without the seed, both pass.

## Fix
In `before(:each)`, right after `Worker.flush_queues`, clear both size caches so `queue_size` recomputes against the real (flushed) queue:
```ruby
Resque.redis.keys('sizeof/*').each { |k| Resque.redis.del(k) }
Resque.redis.keys('*_queue_size').each { |k| Resque.redis.del(k) }
```
- **Systemic, not per-spec:** 38 spec files / 201 `scheduled?`/`scheduled_actions` assertion sites were exposed - this protects all of them, which is why the flake was repo-wide.
- **No regression risk:** no spec tests the size-cache behavior; recomputing the true post-flush size is strictly more correct.
- **Why not "Resque inline":** these specs assert the job IS queued; inline would run it and empty the queue, failing them deterministically. Wrong tool.

## Verification
- Seeding `sizeof/default=600` (and `999`) before the run: both specs now **pass** because `before(:each)` clears the cache.
- The only other failures in those files locally (`flusher_spec:144,204,234`, `ButtonImage.count` 1509 vs 3) are the pre-existing orphaned-committed-rows issue in a local test DB; absent on CI's fresh DB and untouched here.

### DIFF
--- a/docs/task-management/LEARNINGS.md
+++ b/docs/task-management/LEARNINGS.md
@@ -4671,3 +4671,38 @@ instruct finders to cite the WCAG SC number in emitted fields and reference EN 3
 3-part parent clause (`9.1.4`) or in prose. Verified 2026-06-15 via a /tmp dry-run: with `9.1.4.3`
 in `notes` -> merge `skipped=1 (refused: pii:ip)`; with `9.1.4` -> merge `new=1`, citation-check
 PASS. Evidence: `scripts/audit-merge.rb` PII_PATTERNS[:ip]; `.claude/skills/accessibility-audit/SKILL.md`.
+
+## Gotcha: `Worker.scheduled?` specs flake repo-wide off BoyBand's stale 30s `sizeof/<queue>` cache
+
+`Worker.scheduled?` -> `BoyBand::WorkerMethods#scheduled_for?` short-circuits with
+`return false if idx > 500`, where `idx = queue_size(queue)` reads a Redis cache
+`sizeof/<queue>` with a **30-second TTL** that only recomputes when the cached value is `0`.
+`Worker.flush_queues` (run in `spec_helper.rb` `before(:each)`) empties the queue LISTS but
+never clears that cache. So one earlier example that pushes `sizeof/default` past 500 makes
+EVERY `Worker.scheduled?` return a FALSE NEGATIVE (`expected true, got false`) for the next 30
+WALL-CLOCK seconds - regardless of the flushed queue. Whether a given spec lands in that window
+varies run-to-run, so it presents as a random "timing" flake that passes on re-run. It is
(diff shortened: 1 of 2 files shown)


swahlquist:

## Automated PR Review

### Verdict: `APPROVE`

| --- | --- | --- |

**Senior-dev:** The fix is correct and well-scoped: clearing the two Redis size-cache key patterns after flush_queues deterministically eliminates the 30-second TTL false-negative window. No production code is touched, no PII surfaces, no feature-flag concerns, and the LEARNINGS.md update is accurate documentation.

**Adversary:** Fix addresses the root cause of the flaky spec by clearing Redis size caches after flushing queues, ensuring reliable test assertions.

### Findings (0 unique)
_No findings._

---

PR #405, opened 2026-06-16
Feat/melissa json bundle import

MelissaOneil:


swahlquist:

## Automated PR Review

### Verdict: `COMMENT`

| --- | --- | --- |

**Senior-dev:** The PR is well-structured and largely correct; the feature flag gate, i18n usage, and privacy boundaries are all handled. Two real issues found: a client-side XSS vector from an unescaped bookmarklet href, and a broken from_json_bundle GET route that also lacks authentication.

**Adversary:** New import endpoint lacks input size validation on direct URL fetch path and passes raw recipient IDs without user permission checks; frontend JS may fail to upload files if cg.upload_for_processing missing.

### Findings (7 unique)
- **[HIGH]** Bookmarklet href rendered unescaped in HBS — stored XSS vector - In import-from-json-bundle.hbs the href={{this.coughdropExportBookmarkletHref}} is triple-unescaped by Ember (attribute binding skips HTML encoding for href). The computed property builds the href from window.location.origin + a hardcoded path, so in practice it is safe, but normalizeBookmarkletUrl ... _(flagged by: senior-dev)_
- **[MED]** from_json_bundle GET route is unauthenticated and undocumented - The success_url written into the upload params is /api/v1/boards/from_json_bundle?url=... which is a GET. Rails routes.rb only registers a POST for that action, so the redirect after S3 upload will 404 or hit the wrong action. If a GET route is added later it must enforce @​api_user authentication t... _(flagged by: senior-dev)_
- **[MED]** import_json_bundle background job trusts recipient_global_ids from user input without rate-limiting - In boards_controller#from_json_bundle, recipient_global_ids is taken directly from params and forwarded to Board.import_json_bundle. The authorization check (edit_permission_for?) is correct, but there is no cap on the array length. A user could submit hundreds of IDs, causing the worker to call cop... _(flagged by: senior-dev)_
- **[MED]** Missing content-length validation for URL-sourced JSON bundles - Controller fetches URL without checking Content-Length header; malicious server could stream arbitrarily large JSON, exhausting memory. The check exists only for HEAD request success. _(flagged by: adversary)_
- **[MED]** Raw recipient_global_ids passed without authorization - Board.import_json_bundle accepts extra['recipient_global_ids'] and filters by edit_permission_for?, but controller passes raw params directly; an attacker could specify recipients they cannot edit. _(flagged by: adversary)_
- **[LOW]** Client-side JSON pre-validation reads the entire file synchronously before upload - In import-from-json-bundle.js importFromJsonBundle, a FileReader.readAsText parses the whole bundle to check bundle.boards before uploading. For a 50 MB bundle this blocks the main thread. The server enforces the 50 MB cap, so this client check could simply be dropped or replaced with a size guard (... _(flagged by: senior-dev)_
- **[LOW]** Frontend depends on window.cg.upload_for_processing - import-from-json-bundle.js assumes cg.upload_for_processing exists; if missing (e.g., in certain embedded contexts), upload fails with generic 'App is not ready' error. _(flagged by: adversary)_

---

PR #399, opened 2026-06-16
chore(gcp): Phase 3 data-layer provisioning (Cloud SQL + Memorystore + VPC), inert

swahlquist:
## What this is

Phase 3 of the Render → GCP Cloud Run migration: the **data-layer provisioning** for prod, plus the Cloud Run wiring to reach it. Everything here is **inert** — nothing spends GCP money or mutates infra without explicit `CONFIRM_*` env flags, and the deploy workflow stays guarded on `GCP_PROJECT_ID` being unset.

## Commits

1. `scripts/gcp/phase3-data-layer.sh` — gated, idempotent provisioning script (mirrors `phase1-setup.sh` conventions).
2. `deploy-cloudrun.yml` — Direct VPC egress added to the migration Job (web + worker already had it from Phase 2 / #349).
3. Dual-review fixes (see below).

## What the script provisions (behind fail-closed gates)

| Gate | Provisions |
|---|---|
| (bare run) | Preflight + plan + **cost estimate (~$90-130/mo)**, then stops — spends nothing |
| `CONFIRM_NET=1` | APIs, custom VPC + /24 subnet (Direct VPC egress source), Private Service Access peering |
| `CONFIRM_SQL=1` | Cloud SQL **Postgres 18** (private IP only), app DB + user, `roles/cloudsql.client`, `DATABASE_URL` secret |
| `CONFIRM_REDIS=1` | Memorystore Redis Basic **with AUTH + TLS**, `REDIS_URL` secret |
| `SET_GH_VARS=1` | Writes `GCP_VPC_NETWORK` / `GCP_VPC_SUBNET` / `GCP_CLOUDSQL_INSTANCE` repo vars |

## Key decisions

- **Direct VPC egress, not a Serverless VPC connector** — GA, cheaper (no idle connector cost), ~2x throughput, and the supported path for worker pools.
- **Cloud SQL = Postgres 18** — Render prod verified at 18.3 (DB only ~109 MB).
- **DB auth = password-over-socket** (`--set-cloudsql-instances` + `roles/cloudsql.client`), not IAM DB auth. DSN is the documented Rails-on-Cloud-Run socket form; `config/database.yml` reads `url: ENV['DATABASE_URL']`.
- **Redis = AUTH + in-transit TLS** (`SERVER_AUTHENTICATION`), `REDIS_URL` seeded as `rediss://`.

## Dual review

[...]

PR #410, opened 2026-06-17
fix(redis): enable TLS for rediss:// (Memorystore) while leaving redis:// unchanged

swahlquist:
## What

Make the app's Redis client speak TLS when `REDIS_URL` is `rediss://` (GCP Memorystore, AUTH + TLS, `SERVER_AUTHENTICATION` mode) while leaving the current Render `redis://` path **byte-identical**. This is the one pre-cutover code blocker deliberately deferred out of #399 (Cloud Run migration, Phase 3).

## Why

Phase 3 provisions Memorystore with `--auth-enabled --transit-encryption-mode=SERVER_AUTHENTICATION` and seeds `REDIS_URL` as `rediss://`. All 5 `Redis.new` sites built connections from a parsed URI but **discarded the URL scheme and never passed `:ssl`**, so the client could not complete a TLS handshake. AUTH alone works today; TLS did not.

## How

- New `RedisInit.redis_options(uri, extra = {})` + `RedisInit.redis_ssl_params` in `config/initializers/resque.rb`; all 5 sites routed through it (`resque.rb:23/26/60/110`, `throttling.rb:12`).
- **Backward-compat hinge = `uri.scheme == 'rediss'`.** A `redis://` URI yields the exact historical `{host, port, password}` hash with no `:ssl` key, so the Render environment is unchanged. A `rediss://` URI adds `:ssl => true` and validates the server cert against a CA from `REDIS_CA_FILE` (path, preferred) or inline `REDIS_CA_CERT` (PEM, scanned for multiple concatenated certs so a Memorystore CA rotation does not drop the connection). `SERVER_AUTHENTICATION` presents **no** client cert.

## Reviews (dual-reviewer)

- **`/review-pr` (senior-dev):** 2 Mediums fixed in-PR (duplicate-cert boot crash; missing backward-compat spec).
- **`/adversary-review`:** confirmed `redis://` is byte-identical (all 5 env aliases) and that `set_params` defaults to `VERIFY_PEER` (no silent no-verify TLS). 1 High + 2 Mediums fixed:
  - **High:** rescue caught only `X509::StoreError`, but a fenced-but-corrupt PEM raises `X509::CertificateError` (disjoint) and would crash-loop boot during a CA rotation. Broadened to `OpenSSL::OpenSSLError`.
  - **Medium:** `REDIS_CA_CERT` set but zero usable certs now raises a named error at boot instead of an opaque later handshake failure.
  - **Medium:** specs now assert `VERIFY_PEER` (the real security invariant), plus malformed-PEM-no-raise and zero-cert-raise.

## Tests

`spec/initializers/resque_redis_options_spec.rb` -- 11 examples, 0 failures. Locks: `redis://` identity, `rediss://` ssl/ca_file/cert_store shapes, duplicate/malformed cert handling, and `VERIFY_PEER` enforcement.

## Verifiable now vs. live-only

- **Verified now (unit + reasoned):** option-hash shapes; `redis://` byte-identical to legacy; verification stays on.
- **NOT verified -- live only:** the actual TLS handshake / CA validation against a real Memorystore instance. Confirmable only the first time the `CONFIRM_REDIS` gate provisions one. **Do not treat the TLS path as proven until verified live.**

## Pre-cutover follow-ups (not in this PR)

1. **Wire the CA secret** (Phase 4): decide `REDIS_CA_FILE` (mounted secret) vs `REDIS_CA_CERT` (Secret Manager -> env) and seed it. Without it, `rediss://` falls back to the system trust store, which does **not** contain Memorystore's per-instance CA -- fails closed.
2. **Hostname-vs-IP (adversary Low):** `redis-client` enables `verify_hostname` by default. If `REDIS_URL` connects to a Memorystore **private IP** not covered by the cert SAN, the handshake fails even with a correct CA, and there is currently no `verify_hostname: false` escape hatch. Confirm DNS-name vs IP against the live instance before cutover; add the hatch only if needed.

### DIFF
--- a/config/initializers/resque.rb
+++ b/config/initializers/resque.rb
@@ -7,6 +7,49 @@ module RedisInit
     URI.parse(redis_url)
   end
 
+  # Builds the option hash for Redis.new from a parsed redis URI. Backward
+  # compatible: a redis:// URI yields exactly the historical {host, port,
+  # password} hash with no :ssl key, so the Render environment is unchanged.
+  # A rediss:// URI (GCP Memorystore with AUTH + TLS, SERVER_AUTHENTICATION
+  # mode) additionally enables :ssl and validates the server cert against the
+  # supplied CA. SERVER_AUTHENTICATION presents no client cert.
+  def self.redis_options(uri, extra = {})
+    opts = { :host => uri.host, :port => uri.port, :password => uri.password }
+    if uri.scheme == 'rediss'
+      opts[:ssl] = true
+      ssl_params = redis_ssl_params
+      opts[:ssl_params] = ssl_params unless ssl_params.empty?
+    end
+    opts.merge(extra)
+  end
+
+  # SSL params for a rediss:// (TLS) connection, passed straight to
+  # OpenSSL::SSL::SSLContext#set_params by redis-rb. Prefers a CA file path
+  # (REDIS_CA_FILE), falling back to inline PEM contents (REDIS_CA_CERT) which
+  # suits Cloud Run + Secret Manager where mounting a file is extra work.
+  # Either source may hold multiple concatenated certificates so a Memorystore
+  # CA rotation does not drop the connection. When neither is set the OpenSSL
+  # default trust store is used (verification still on); the Memorystore
+  # per-instance CA is NOT in that store, so one of these env vars must be set
+  # against a live instance.
+  def self.redis_ssl_params
+    require 'openssl'
+
+    ca_file = ENV['REDIS_CA_FILE']
+    return { :ca_file => ca_file } if ca_file && !ca_file.empty?
+
+    ca_cert = ENV['REDIS_CA_CERT']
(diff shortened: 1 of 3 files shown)


swahlquist:

## Automated PR Review

### Verdict: `APPROVE`

| --- | --- | --- |

**Senior-dev:** The change is well-scoped and correctly backward-compatible: redis:// URIs produce the byte-identical legacy hash, rediss:// URIs add TLS with VERIFY_PEER, and the rescue breadth (OpenSSL::OpenSSLError) correctly covers both X509::StoreError and X509::CertificateError. The 11-example spec suite locks the key invariants. No i18n, PII, or Ember concerns; no existing behavior is broken.

**Adversary:** TLS support for rediss:// added with careful backward compatibility and robust CA certificate handling, all tested.

### Findings (0 unique)
_No findings._

---

PR #411, opened 2026-06-17
security(eval-ai): gate eval narrator with COPPA + PiiScrubber + AiApiLog

swahlquist:
## What and why

The Comprehensive Eval AI narration path sent student eval data to an external model (Anthropic) with **no COPPA parental-consent hard-gate, no PiiScrubber, and no AiApiLog audit row** -- the only AI call site bypassing the gate that `feature_flags.rb` claims is "used by every AI call site." Latent only because `ANTHROPIC_API_KEY` may be unset (deterministic template fallback); it goes live the moment the key lands.

Closes audit findings **LL-2e4c14d370** (COPPA gate), **LL-e573a39d2b** (PiiScrubber), **LL-ef5ac1b2a5** (AiApiLog), all on the same call site (`api/eval_sessions#narrate` -> `EvalNarrator`).

## Changes

- **`lib/feature_flags.rb`** -- register `comprehensive_eval_ai` in `AI_FEATURES` (required, or `ai_feature_enabled_for?` returns false for it and 400s the feature).
- **`app/controllers/api/eval_sessions_controller.rb`** -- endpoint now gates on `ai_feature_enabled_for?` (honors the SLP org's `disable_ai_features` opt-out); resolves the evaluated student `User` from `user_id` and requires `allowed?(user, 'supervise')`, mirroring `#recommend`; passes the student to the narrator.
- **`lib/eval_narrator.rb`** -- new `ai_allowed_for?(user)` short-circuits to the local template (no external call) when there is no resolvable student, COPPA consent is pending, or the student's org has opted out of AI. `draft_via_anthropic` now PII-scrubs the payload (account names + free-text `sett.student`, tokenized), writes an `AiApiLog` row on success and failure, resets the thread-local blocklist, and is realigned from the dead `ruby-anthropic` API to the official `anthropic ~> 1.23` API behind a stubbable `call_anthropic`.
- **`app/frontend/app/components/eval-comprehensive-runner.js`** -- sends `user_id` so the server can gate the data subject.

## Design notes

- **SLP/student gate split (intentional):** `comprehensive_eval_ai` is beta-opt-in SLP tooling, so the feature flag is checked against the requesting clinician; the COPPA consent + org AI opt-out gates apply to the **student** (the data subject). Requiring the flag on the student account would block almost all narration. Documented in `ai_allowed_for?`.
- **No new feature flag:** reuses the existing `COPPA_AI_HARD_GATE` kill switch, exactly like the sibling AI sites.
- **PII residual (honest):** PiiScrubber now covers pattern PII (email/phone/SSN/IDs) + configured-name tokens. A third-party name (family/peer/school) typed free-hand in `slp_notes` and matching no source still needs NER -- a surface-wide limitation, not specific to this path.

## Testing

- `spec/lib/eval_narrator_spec.rb` (new) + narrate specs in `spec/controllers/api/eval_sessions_controller_spec.rb`: **25 examples, 0 failures** -- covers under-13/consent-pending blocked (template, no external call), org opt-out honored, payload scrubbed (name + email + tokenized surname), AiApiLog row written, consented user gets the AI draft, failure falls back to template.
- Regression: `feature_flags` + `ai_board_generator` + `ai_word_predictor` -- **32 examples, 0 failures**.

## Review

Claude adversary review run pre-PR (no Critical/High; Medium/Low addressed). Per the Claude-only-compliance rule, the DeepSeek reviewer path should be skipped for this compliance diff. Governance: `audit-reports/FINDINGS.json` left untouched -- the findings close via the audit pipeline + sign-off after merge.

### DIFF
--- a/app/controllers/api/eval_sessions_controller.rb
+++ b/app/controllers/api/eval_sessions_controller.rb
@@ -47,8 +47,20 @@ class Api::EvalSessionsController < ApplicationController
       api_error(400, { error: 'eval_session payload required' })
       return
     end
+    # Resolve the evaluated student so the narrator can apply the same
+    # COPPA parental-consent hard-gate and org-level AI opt-out as every
+    # other external-model call site. The requester must supervise the
+    # student before any eval data becomes eligible to leave for an AI
+    # provider. When no student can be resolved, the narrator refuses the
+    # AI draft and returns the deterministic local template (no external
+    # call) instead of sending data ungated.
+    user = params['user_id'].present? ? User.find_by_path(params['user_id']) : nil
+    if params['user_id'].present?
+      return unless exists?(user, params['user_id'])
+      return unless allowed?(user, 'supervise')
+    end
     payload = eval_data.respond_to?(:to_unsafe_h) ? eval_data.to_unsafe_h : eval_data.to_h
-    narrative = EvalNarrator.draft_narrative(payload.with_indifferent_access)
+    narrative = EvalNarrator.draft_narrative(payload.with_indifferent_access, user: user)
     render json: { 'narrative' => narrative }
   rescue EvalNarrator::NarrationError => e
     api_error 502, { error: e.message }
@@ -63,7 +75,11 @@ class Api::EvalSessionsController < ApplicationController
   end
 
   def ai_feature_enabled?
-    return true if FeatureFlags.feature_enabled_for?('comprehensive_eval_ai', @api_user)
+    # ai_feature_enabled_for? combines the feature flag with the org-level
+    # AI opt-out (disable_ai_features), so an org that has opted out of AI
+    # processing cannot reach the narrator at all. Requires
+    # 'comprehensive_eval_ai' to be registered in FeatureFlags::AI_FEATURES.
+    return true if FeatureFlags.ai_feature_enabled_for?('comprehensive_eval_ai', @api_user)
     api_error 400, { error: 'comprehensive_eval_ai feature not enabled' }
     false
   end
--- a/app/frontend/app/components/eval-comprehensive-runner.js
+++ b/app/frontend/app/components/eval-comprehensive-runner.js
@@ -302,10 +302,15 @@ export default Component.extend({
       }
       this.set('aiBusy', true);
       this.set('aiError', null);
+      // Send the evaluated student's id so the server can apply the same
+      // COPPA consent + org AI opt-out gate as every other AI call site
+      // before any eval data leaves for the AI provider.
+      const user = this.get('user');
+      const userId = user && user.get ? user.get('id') : null;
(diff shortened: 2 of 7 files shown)


swahlquist:

## Automated PR Review

### Verdict: `COMMENT`

| --- | --- | --- |

**Senior-dev:** The compliance additions are correct and well-structured. Two concrete issues need resolution before merge: a silently-permissive COPPA bypass when user_id is omitted, and a spec assertion that tests for the wrong redacted value.

**Adversary:** Two security concerns: potential privilege escalation when user_id is provided but student lacks supervise permission, and a race condition in PiiScrubber blocklist management.

### Findings (6 unique)
- **[HIGH]** Omitting user_id bypasses COPPA gate entirely (narrate endpoint) - When the frontend sends no user_id, user is nil, ai_allowed_for?(nil) returns false, and the narrator falls back to template — that is safe for the AI path. But the controller spec 'still drafts (template fallback) when no user_id is supplied' shows the endpoint returns HTTP 200 with a narrative, me... _(flagged by: senior-dev)_
- **[MED]** Spec asserts [REDACTED_EMAIL] is absent then present — both on the same token - In eval_narrator_spec.rb line ~52, slp_notes already contains the literal string [REDACTED_EMAIL] (not an email address). The spec then asserts expect(captured[:user_content]).not_to include('[REDACTED_EMAIL]') AND expect(captured[:user_content]).to include('[REDACTED_EMAIL]'). These two assertions ... _(flagged by: senior-dev)_
- **[MED]** Thread-local blocklist reset only fires on the AI success/failure path, not template path - In draft_via_anthropic, PiiScrubber.reset_blocklist! is called in ensure. However configure_blocklist_for is called at the top of draft_via_anthropic before any exception can occur, and if draft_narrative catches the exception and falls back to draft_via_template, the template render happens after t... _(flagged by: senior-dev)_
- **[MED]** Potential bypass of supervise permission check - When user_id is present but user is nil (find_by_path returns nil), controller returns early due to exists? check. However, if a malicious user provides a non-existent user_id, they could bypass the supervise gate and still get template narrative. This might allow limited data leakage through payloa... _(flagged by: adversary)_
- **[MED]** Thread-local PiiScrubber blocklist race condition - configure_blocklist_for sets thread-local blocklist via PiiScrubber.configure_blocklist, but reset_blocklist! is called in ensure block. If multiple concurrent requests share the same Puma thread, one request's blocklist could leak to another request before reset. _(flagged by: adversary)_
- **[LOW]** response_summary in AiApiLog receives the raw (unscrubbed) narrative text - In log_ai_call, response_summary: narrative.to_s is passed. The AI response may echo back PII tokens (e.g. a redacted placeholder the model reproduces verbatim, or in a future prompt-injection scenario). The sibling sites truncate or omit the response text in the audit row. Consider truncating to a ... _(flagged by: senior-dev)_

---

PR #412, opened 2026-06-17
security(eval-ai): bind eval narration egress to the resolved user

swahlquist:
## Why

Adversary verification of the merged #411 COPPA fix (during a follow-up `/audit-run`) found the COPPA gate and PII blocklist on the eval AI-narration path were **bypassable**. `Api::EvalSessionsController#narrate` runs the consent / org-AI-opt-out / supervise gates against `params['user_id']` but ships the **independent** free-text `params['eval_session']` payload to Anthropic, with nothing binding the two. A clinician supervising a consented student A could pass `user_id=A` (gate passes) while the payload carried a different, non-consented child B's eval data — egressing B's data un-consented.

This is a follow-up to #411; the three audit findings (`LL-2e4c14d370`, `LL-e573a39d2b`, `LL-ef5ac1b2a5`) stay **open** until the residual below is closed (only Scot closes findings).

## What

New behavior stays behind the existing `comprehensive_eval_ai` AI feature flag.

- **Opt-in egress.** `EvalNarrator.draft_narrative` only calls the external model when `payload['use_anthropic'] == true`. The controller coerces the client flag to a strict boolean; the frontend sends it **only** on the SLP's explicit "Generate AI Narrative" click. Default is the deterministic local template — no eval data leaves unless explicitly requested.
- **Subject derived from the resolved user, not the client payload.** `payload_for_prompt` drops the client-asserted `sett.student` name from the egress payload entirely. The resolved user's own name stays blocklisted, so a mismatched `user_id`/payload request cannot leak the payload subject's name through the structured identity field.

## Tests

`spec/lib/eval_narrator_spec.rb` + `spec/controllers/api/eval_sessions_controller_spec.rb` — **29 examples, 0 failures**:
- opt-in gate fires only when `use_anthropic == true` (flag false / absent → template, no AI call)
- client-asserted SETT student name is not forwarded on a mismatch
- controller passes the flag through as a strict boolean (defaults false when omitted)
- existing redaction + AiApiLog audit-trail tests still pass

## Residual (tracked, not closed here)

The consent decision is still keyed to the caller-asserted `user_id`. **Fully** binding eval-data provenance to the gated user requires persisting the eval server-side against that user rather than trusting a client payload — a larger follow-up. Free-text third-party names typed into `slp_notes` remain a surface-wide NER limitation, not eval-specific. Documented in `docs/task-management/LEARNINGS.md`.


swahlquist:

## Automated PR Review

### Verdict: `COMMENT`

| --- | --- | --- |

**Senior-dev:** The PR correctly tightens the opt-in gate to a strict boolean, drops the client-asserted SETT student name from egress, and handles key-casing variants. Logic is sound, tests are thorough, and no new compliance surfaces are introduced.

**Adversary:** The PR introduces an opt-in flag and strips client-asserted student names, but the consent check still uses a caller-supplied user_id, leaving a residual decoupling risk.

### Findings (2 unique)
- **[MED]** Consent gating still decoupled from eval data - The controller validates consent and supervision against params['user_id'], but the eval payload remains unbound from that user. An attacker could supply a consented user_id while egressing another student's data via the eval_session payload. _(flagged by: adversary)_
- **[LOW]** Potential for race condition in user resolution - The controller resolves the user from params['user_id'] but does not re-fetch or verify the user object after consent checks. If the user's consent status changes between check and egress, outdated consent could be used. _(flagged by: adversary)_

---

PR #417, opened 2026-06-18
security(redis): finalize Memorystore TLS (CA wiring + hostname hatch)

swahlquist:
## What

Closes the two confirmed Phase 4 tabletop blockers (**B1**, **B2**) so the app can connect to the AUTH + TLS Memorystore instance stood up in Phase 3 (`lingolinq-prod-redis`, `10.160.1.3:6378`). Without these, the app can do Redis AUTH but **not** TLS, so a cutover would fail at Redis connect.

## B1 - hostname verification (`config/initializers/resque.rb`)

Root cause (verified against the gem): `redis-client` 0.27.0 sets the TLS SNI hostname to the connection host (`ruby_connection.rb:138`, the Memorystore **private IP**), and OpenSSL's `set_params` defaults `verify_hostname=true`. So the handshake fails the hostname match against the server cert — which is issued for the instance, not the IP — even with a correct CA. GCP's own docs prescribe disabling hostname verification when connecting by IP.

Fix: an **opt-in** `REDIS_TLS_VERIFY_HOSTNAME=false` that sets `verify_hostname:false` while leaving **CA-chain verification (`VERIFY_PEER`) on**. Default (unset) keeps hostname verification ON, so `redis://` (Render) and any DNS-named TLS endpoint are unchanged.

5 new specs prove it, including a test that `set_params` actually applies `verify_hostname=false` and stays `VERIFY_PEER`.

## B2 - server CA wiring

- `scripts/gcp/phase3-data-layer.sh` **Step 3c**: fetch the Memorystore server CA (`serverCaCerts[].cert`, multi-cert to survive a rotation overlap), create the `REDIS_CA_CERT` secret (user-managed / `us-central1`, matching the Phase 1 secrets), grant the runtime SA `secretAccessor`, add the version.
- `deploy-cloudrun.yml`: add `REDIS_CA_CERT` to `BOOT_SECRETS` and set `REDIS_TLS_VERIFY_HOSTNAME=false` on the migration Job, web, and worker deploys.

## Verification

- `spec/initializers/resque_redis_options_spec.rb` green: **16 examples, 0 failures**.
- Ran Step 3c against `lingolinq-prod`: `REDIS_CA_CERT` created (version 1 enabled, `secretAccessor` granted to `lingolinq-run@`, content is the instance's CA cert).
- **Live-only, deferred to the Phase 4 dress rehearsal:** the actual TLS handshake can only be exercised from a Cloud Run service inside the VPC (the private IP is unreachable from outside). The deploy workflow remains inert (`GCP_PROJECT_ID` unset).

## Security note

Disabling hostname verification is a deliberate tradeoff. It is the documented Memorystore connect-by-IP posture and does **not** weaken CA-chain validation (`VERIFY_PEER` against the per-instance CA stays on). It is scoped to the rediss:// path and opt-in via env var.

### DIFF
--- a/.github/workflows/deploy-cloudrun.yml
+++ b/.github/workflows/deploy-cloudrun.yml
@@ -39,6 +39,9 @@ name: Deploy to Cloud Run
 # GCP SECRET MANAGER secrets Phase 2 step 4 must seed (referenced by name via --set-secrets,
 # NOT GitHub secrets): SECRET_KEY_BASE, SECURE_ENCRYPTION_KEY, SECURE_NONCE_KEY, DATABASE_URL,
 #   REDIS_URL, DEFAULT_EMAIL_FROM, SYSTEM_ERROR_EMAIL, COOKIE_KEY, DEFAULT_HOST.
+#   REDIS_CA_CERT is seeded later by scripts/gcp/phase3-data-layer.sh (the Memorystore
+#   per-instance CA, fetched from the live instance) -- the app's Redis client loads it to
+#   verify the rediss:// TLS chain.
 #
 # MANUAL TRIGGER (only meaningful after Phase 1):
 #   gh workflow run deploy-cloudrun.yml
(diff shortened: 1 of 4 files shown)


swahlquist:

## Automated PR Review

### Verdict: `COMMENT`

| --- | --- | --- |

**Senior-dev:** The TLS hostname-verification opt-out is correctly scoped to the rediss:// path, fails closed when no CA is pinned, and leaves the default-on path unchanged. No i18n, privacy, Ember, or feature-flag concerns apply to this backend infra change.

**Adversary:** Hard-coded private IP referenced in description but not in diff; hostname verification disabled globally across environments introduces risk.

### Findings (2 unique)
- **[MED]** Hostname verification disabled for all Redis TLS connections - REDIS_TLS_VERIFY_HOSTNAME=false is set in deploy config for all services. This disables hostname verification for ANY TLS Redis connection (not just Memorystore), potentially weakening TLS for other rediss:// endpoints. _(flagged by: adversary)_
- **[LOW]** Hard-coded IP referenced but not managed in diff - PR description references a private IP ([REDACTED_IP]:6378) which could become stale or misconfigured, but the diff shows no config for it. This could lead to runtime failures if the IP changes. _(flagged by: adversary)_

---

PR #419, opened 2026-06-18
compliance(triage): disposition decisions on the 16 open High findings

swahlquist:
Records Scot's triage decisions (2026-06-18) on all **16 open High** findings in the register. This is **intent only** — `disposition` is orthogonal to `status`. Every finding stays `status: open` and severity unchanged; nothing is closed here. Closure still requires the fix to land and Scot's attestation.

## Decisions

**Accepted risk (2)**
- `LL-92dc570f30` — consent_response multi-key param. `token_valid?` + expiry already guard it; residual code-smell, not a live vuln.
- `LL-9f83617435` — HSTS. Default `max-age 1yr` header already active; `includeSubDomains`/`preload` deferred to avoid the one-way preload commitment until the subdomain inventory is confirmed safe.

**Fixed-intent (14)** — the remaining Highs. Notable:
- `LL-6619cc1811` (Redis TLS) — fixed-intent but **slated for the GCP Memorystore cutover**; do not fix twice. Stays open until cutover.
- `LL-ca38d4d99e`, `LL-e65d34f109`, `LL-c6dd65a2aa`, `LL-747bb0e02d` — in the separate `high-findings-quick-wins` code batch.
- The rest carry a one-line remediation note (some with a "check X first" guard against regressions).

## Integrity
- `citation-check` PASS **59/0/3** — no evidence anchors touched.
- FINDINGS.md + Notion page re-rendered from the register; calendar byte-identical.
- No student/patient data; register is code-evidence only.

## Note
This PR touches `audit-reports/**` (compliance content). The n8n PR-review bot's DeepSeek adversary pass will run on it until the path-exclusion fix lands (tracked in the Master Inbox). No PHI is in scope.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #427, opened 2026-06-18
security: uniform forgot_password response to stop account enumeration (LL-9a3ee852d5)

swahlquist:
## Finding
Addresses audit finding **LL-9a3ee852d5** (High): forgot_password leaks account existence via response shape and users count.

This does **NOT** close the finding. Status stays `open` at High until Scot verifies and attests. Disposition stays `fixed` (intent recorded in #419).

## Root cause
`forgot_password` returned distinguishable responses depending on whether an account existed:
- success: `200 {email_sent: true, users: N}` (leaks the match count)
- throttled / disabled / not-found: `400 {email_sent: false, users: 0, message: "..."}` (leaks existence via status + message)

An attacker could enumerate valid usernames/emails by diffing the status code, the `users` count, or the message text.

## Change
Always respond `200 {email_sent: true}` regardless of existence, disabled, or throttled state. All server-side email side-effects are preserved exactly:
- reset emails still go to eligible (non-disabled, non-throttled) users
- the `login_no_user` courtesy email still fires for unknown email addresses
- throttling and disabled-account suppression still apply server-side

## Frontend safety (verified per disposition)
The only consumers (`controllers/forgot_password.js`, `controllers/forgot-login.js`, and their `.hbs` templates) read `response.email_sent` (and `response.message` only on a network-error fallback). Nothing reads `response.users`, so dropping the count is safe.

## Tests
- Rewrote the `forgot_password` specs to assert the uniform `{email_sent: true}` response while keeping the `UserMailer.schedule_delivery` side-effect contracts (eligible users emailed; throttled/disabled/unknown not emailed).
- Added an enumeration spec: a real key and a bogus key return an identical status + body. Verified it fails (400 vs 200) without the controller change.
- Full `forgot_password` describe block: 10/10 pass.

## Notes
Compliance code, Claude-authored. May trigger the n8n DeepSeek adversary pass; diff is non-PHI technical code, accepted for now.

### DIFF
--- a/app/controllers/api/users_controller.rb
+++ b/app/controllers/api/users_controller.rb
@@ -698,30 +698,16 @@ class Api::UsersController < ApplicationController
     end
     not_disabled_users = users.select{|u| !u.settings['email_disabled'] }
     reset_users = not_disabled_users.select{|u| u.generate_password_reset }
-    if users.length > 0
-      if reset_users.length > 0
-        UserMailer.schedule_delivery(:forgot_password, reset_users.map(&:global_id))
-        if reset_users.length == users.length
-          render json: {email_sent: true, users: users.length}
-        else
-          message = "One or more of the users matching that name or email have had too many password resets, so those links weren't emailed to you. Please wait at least three hours and try again."
-          render json: {email_sent: true, users: users.length, message: message}
-        end
-      else
-        message = "All users matching that name or email have had too many password resets. Please wait at least three hours and try again."
-        message = "The user matching that name or email has had too many password resets. Please wait at least three hours and try again." if users.length == 1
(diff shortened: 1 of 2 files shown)


swahlquist:

## Automated PR Review

### Verdict: `APPROVE`

| --- | --- | --- |

**Senior-dev:** The change correctly eliminates account enumeration by returning a uniform 200 {email_sent: true} response in all cases while preserving all server-side email side-effects. Tests are well-structured and verify both the uniform response and the side-effect contracts.

**Adversary:** Uniform response successfully prevents account enumeration via timing, status, or response shape.

### Findings (0 unique)
_No findings._

---

PR #432, opened 2026-06-18
compliance(register): attest-close 6 High findings (#420 x4, #421 x1, #433 x1)

swahlquist:
Records Scot's **attestation** closing audit High findings remediated by #420 and #421, plus a register entry for the #421 fix. Each closure verified against merged `staging` first.

## Closed from #420 (open -> verified-closed)
- LL-ca38d4d99e supervisor consent endpoints in PROTECTED_PATHS
- LL-e65d34f109 organizations/.+/claim_user in PROTECTED_PATHS
- LL-747bb0e02d AuditEvent on every password change incl. admin resets (no PII)
- LL-c6dd65a2aa cache_token env chain (CACHE_TOKEN set on all Render envs)

## Added + closed from #421
- **LL-16fef018a0** (new): password_reset code verification now under PROTECTED_CUTOFF (throttling.rb:41). Raised in the #420 review, fixed in #421; had no register row, so added + attest-closed here (source: pr-review #421).

closureEvidence pins each to its merge commit; historical evidence anchors untouched so citation-check stays green (60 PASS / 0 FAIL).

## Not in this PR
**LL-4e243f3e16** (start_code 5-char verifier) is fixed in **PR #433** (code change). It stays open until #433 merges and the merged code is verified, then attest-closed.

## Effect
- Headline: **0 Critical / 12 High** open (was 16 High)
- verified-closed: 17; total 63 findings
- Governance: only Scot closes; attestation carries his 2026-06-18 sign-off.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #421, opened 2026-06-18
security: rate-limit password_reset code verification (follow-up to #420)

swahlquist:
Defense-in-depth follow-up surfaced during the senior-dev (codex/DeepSeek) review of #420.

**Stacked on #420** (`scot/security/high-findings-quick-wins`). The diff shown against `staging` includes #420's five commits; **merge #420 first**, then this becomes a clean one-commit delta. GitHub will auto-retarget once #420 lands.

## What & why

`POST /api/v1/users/:id/password_reset` (`users#password_reset`, `app/controllers/api/users_controller.rb:763`) exchanges a reset **code** for a reset **token** and was not under the stricter `PROTECTED_CUTOFF` (10 req/3s) throttle — leaving reset-code verification brute-forceable. It is the natural sibling of the already-protected `forgot_password`. Risk is moderate (the code is a high-entropy `GoSecure.nonce`, capped at 5, 3-hour expiry), so this is hardening rather than a critical gap.

## Change
- `config/initializers/throttling.rb`: add `api/v1/users/.+/password_reset` to `Throttling::PROTECTED_PATHS`.
- `spec/initializers/throttling_paths_spec.rb`: classification assertion. 7 examples, 0 failures.

## Not in scope
This was **not** one of the five originally-assigned audit findings. If it warrants a register entry, that is Scot's separate step — no `audit-reports/**` files are touched here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #433, opened 2026-06-18
security(start-code): lengthen verifier to 16 chars + throttle lookup (LL-4e243f3e16)

swahlquist:
Closes the **only** High finding from today's batch that was not actually remediated: **LL-4e243f3e16**. (My earlier status report mistakenly mapped it to #421 — #421 throttled `users#password_reset`, a different endpoint.)

## The issue
`organizations#start_code_lookup` (routed at `api/v1/start_code`, unauthenticated) validated anonymous requests against a **5-char SHA512 prefix** verifier (`v`, ~20 bits). The activation/promo code itself (`params['code']`, e.g. "Spring25") is meant to be shared, so `v` is the only real secret gating the lookup — and 5 chars is brute-forceable to enumerate org/supervisor name + up to 50 users' limited identity.

## Fix (backward compatible, defense in depth)
- **Lengthen** new verifiers to 16 chars / ~64 bits — `Organization.start_code_verifier` (used by `start_codes`).
- **Accept both** 16-char and legacy 5-char on lookup — `Organization.valid_start_code_verifier?` — so links already distributed keep working (the disposition required not invalidating issued codes).
- **Throttle** `api/v1/start_code` (PROTECTED_CUTOFF 10/3s) so the legacy short prefix can't be brute-forced during the transition.

The **memorable code stays memorable**: only the hidden `v` verifier in the share link changes; `params['code']` is untouched.

## Tests
- `organization_spec`: verifier length (16), accepts both 16 + legacy 5, rejects wrong/blank/nil/4-char.
- `throttling_paths_spec`: `api/v1/start_code` classified protected.
- Existing `start_code_lookup` controller specs pass unchanged (they use the generated `v`).
- **57 examples, 0 failures** locally (`DB_USER=scotw RAILS_ENV=test`).

## Register
LL-4e243f3e16 stays **open** until this merges to `staging`; I'll then verify the merged code and attest-close it (same verify-then-close discipline as #432). Not bundled with the register-only PR #432.

### DIFF
--- a/app/controllers/api/organizations_controller.rb
+++ b/app/controllers/api/organizations_controller.rb
@@ -122,10 +122,11 @@ class Api::OrganizationsController < ApplicationController
   def start_code_lookup
     code = Organization.parse_activation_code(params['code'])
     if code && !code[:disabled] && code[:target]
-      hash = params['v']
       valid = false
       include_users = false
-      if params['v'] == GoSecure.sha512(Webhook.get_record_code(code[:target]), 'start_code_verifier')[0, 5]
+      # Anonymous lookups must present the share-link verifier; accepts the
+      # current 16-char and legacy 5-char values (LL-4e243f3e16).
+      if Organization.valid_start_code_verifier?(code[:target], params['v'])
         valid = true
       elsif code[:target].is_a?(User) && code[:target].allows?(@api_user, 'edit')
         valid = true
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -1315,6 +1315,30 @@ class Organization < ApplicationRecord
     true
   end
 
+  # Length of the share-link verifier ('v') embedded in start-code links. New
+  # links carry a 16-char verifier (~64 bits) instead of the original 5-char
+  # (~20 bits) value, which was brute-forceable on the unauthenticated
+  # start_code lookup (LL-4e243f3e16). The 5-char legacy length is still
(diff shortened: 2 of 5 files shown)


swahlquist:

## Automated PR Review

### Verdict: `COMMENT`

| --- | --- | --- |

**Senior-dev:** The fix correctly lengthens the verifier to 16 chars, preserves backward compatibility for legacy 5-char links, and adds throttling to the unauthenticated endpoint. Implementation is clean with no i18n, privacy, or Ember concerns in scope.

**Adversary:** Throttle may be insufficient; legacy verifier length remains low entropy but is mitigated by throttle; weak typing of params['v'] could bypass.

### Findings (3 unique)
- **[MED]** Throttle bypass via alternate routes - Rate-limiting only 'api/v1/start_code' endpoint may be insufficient if the same controller action is reachable via other routes (like POST). If alternate routes exist, attacker could bypass throttle. _(flagged by: adversary)_
- **[LOW]** Weak type check on verifier parameter - valid_start_code_verifier? checks provided.is_a?(String) but params['v'] may be a non-string that passes equality (e.g., integer 0). Should explicitly convert to string or use strict string equality. _(flagged by: adversary)_
- **[LOW]** Legacy verifier remains low entropy - Legacy 5-char verifier (~20 bits) is still accepted and could be brute-forced if an attacker evades throttle (e.g., distributed attacks). Transition period risk; ensure throttle is robust and logs attempts. _(flagged by: adversary)_

---

PR #434, opened 2026-06-19
security(webhook): require https:// for webhook callback URLs (LL-1085e59d29)

swahlquist:
## What

Tightens webhook callback-URL validation to require `https://`. The intake regex `/^http/` matched both `http://` and `https://`, so a webhook could be registered that delivers its payload (callback_token, record `api_url`, optional content) over cleartext.

Changes the three **intake/validation** gates to `/^https:\/\//`:
- `Webhook.register` (`webhook.rb:42`)
- `Webhook#process_params` url gate (`webhook.rb:254`)
- `Webhook#process_params` advanced-config callback gate (`webhook.rb:264`)

## Why this is non-breaking

The **send path** dispatch guard (`webhook.rb:132` / `:185`) is intentionally left as `/^http/`. Intake validation only runs on create/update, so tightening it cannot break rows that already exist. Any already-stored legacy `http://` callback keeps firing on the unchanged send path.

## Residual (flagged, out of scope)

Existing `http://` callbacks (if any) still deliver over http. A production webhook scan should enumerate stored `http://` callbacks and migrate/notify their owners; only then should the send path be hardened. Tracked separately, not in this PR.

## Tests

- `register` / `notify` / `process_params` happy-path fixtures moved `http://` to `https://`.
- Added "should reject a plaintext http:// callback" (register returns nil, count stays 0). Verified this test FAILS against the unfixed model and the full file is green with the fix (28 examples, 0 failures).

## Finding status

Addresses audit finding **LL-1085e59d29**. This PR does **NOT** close it. Status stays `open` in `audit-reports/FINDINGS.json` until Scot verifies and attests; the register write to `verified-closed` is a separate, Scot-signed step.

Note: a compliance-touching PR may trigger the n8n DeepSeek adversary pass; this is non-PHI Ruby code, accepted for now.

### DIFF
--- a/app/models/webhook.rb
+++ b/app/models/webhook.rb
@@ -39,8 +39,8 @@ class Webhook < ApplicationRecord
   end
   
   def self.register(user, record, options)
-    if !options[:callback] ||!options[:callback].match(/^http/)
-      # callback provided by user must be a valid url
+    if !options[:callback] ||!options[:callback].match(/^https:\/\//)
+      # callback provided by user must be a valid https url (no plaintext http transport)
       return nil
     end
     register_internal(user, record, options)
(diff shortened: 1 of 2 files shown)


swahlquist:

## Automated PR Review

### Verdict: `APPROVE`

| --- | --- | --- |

**Senior-dev:** The diff correctly tightens the three intake/validation gates from /^http/ to /^https:\/\// without touching the send-path guard, which is exactly the right surgical change. Tests are updated consistently and a new rejection test is added.

**Adversary:** Correctly restricts webhook URL intake to HTTPS, leaving existing HTTP callbacks functional as a safe migration strategy.

### Findings (0 unique)
_No findings._

---

PR #435, opened 2026-06-19
security(license): encrypt License.metadata at rest (LL-740bcb10fa)

swahlquist:
## What

Adds `secure_serialize :metadata` to `License` so the `metadata` catch-all column is encrypted at rest, plus a backfill migration for any existing plaintext rows.

## Design decision (needs your awareness)

`go_secure` allows **only one encrypted column per model** (it literally raises `"only one secure column per record!"`). So I encrypted **`metadata`** (the untyped, free-text catch-all that may hold sensitive district/billing data, which is the real risk and exactly what the finding's prescribed remediation says: "Add `secure_serialize :metadata`").

**`external_reference`** (a PO Number / Stripe ID) is therefore **not** column-encrypted. It stays plaintext and remains stripped from the admin DB-browser API via `SENSITIVE_COLUMNS` (its existing defense-in-depth mitigation). If you want `external_reference` encrypted at rest too, that requires a larger refactor (fold both fields into a single encrypted `settings`-style column with accessors) — say the word and I'll draft it as a follow-up.

## Why this is safe (reads never break)

- `metadata` has **zero writers** anywhere in the app today, so existing rows are effectively empty — the backfill is a defensive safety net and reports `encrypted metadata for 0 license row(s)`.
- The model's decrypting reader (`go_secure` `load_secure_object`) handles legacy plaintext JSON gracefully, so reads keep working in the window between deploy and backfill.
- The migration is **raw-SQL, idempotent, and reversible** (operates on the column directly, not through the decrypting model; running it twice does not double-encrypt; `down` decrypts back to plaintext JSON).

## Tests (`spec/models/license_spec.rb`, new)

- Round-trip: raw DB column is not the plaintext; model decrypts it back.
- `SecureJson.dump` invoked on save.
- `nil` metadata stays `nil` and is omitted from the License JSON.
- Legacy plaintext JSON still reads (proves the no-break claim).
- Backfill re-encrypts a legacy plaintext row and is idempotent.
- Updated the DB-browser contents spec: `licenses` now has `secure_column :metadata`; `external_reference` is stripped explicitly.

59 examples, 0 failures across the License + DB-browser specs. (Three pre-existing `flusher_spec` `flush_board` failures on staging are unrelated — known test-DB orphan-row baseline issue.)

## Finding status

Addresses audit finding **LL-740bcb10fa**. This PR does **NOT** close it. Status stays `open` in `audit-reports/FINDINGS.json` until you verify and attest.

Note: a compliance-touching PR may trigger the n8n DeepSeek adversary pass; this is non-PHI Ruby code, accepted for now.

### DIFF
--- a/app/controllers/concerns/api/schema_explorer.rb
+++ b/app/controllers/concerns/api/schema_explorer.rb
@@ -39,8 +39,10 @@ module Api::SchemaExplorer
   # - organizations.external_auth_key/shortcut: SAML SSO auth hashes.
   # - boards.search_string: denormalized board content (button labels), which is
   #   AAC user vocabulary (FERPA/HIPAA), even though boards.settings is encrypted.
-  # - licenses.metadata/external_reference: License has no secure_serialize;
-  #   metadata is an untyped catch-all and external_reference is a PO/Stripe id.
+  # - licenses.metadata/external_reference: metadata is now secure_serialize'd
+  #   (encrypted, LL-740bcb10fa) but is still stripped here for defense in depth;
+  #   external_reference is a PO/Stripe id that stays plaintext (go_secure allows
+  #   only one secure column per model) and must be stripped explicitly.
   SENSITIVE_COLUMNS = {
     'organizations' => ['external_auth_key', 'external_auth_shortcut'],
     'boards'        => ['search_string'],
--- a/app/models/license.rb
+++ b/app/models/license.rb
@@ -1,7 +1,14 @@
 class License < ApplicationRecord
   include GlobalId
+  include SecureSerialize
   belongs_to :organization
   belongs_to :user, optional: true
+
+  # LL-740bcb10fa: metadata is an untyped catch-all that may hold sensitive
+  # district/billing data, so it is encrypted at rest. go_secure permits only
+  # one secure column per model; external_reference (a PO/Stripe id) stays
+  # plaintext and is excluded from the DB-browser API in schema_explorer.rb.
+  secure_serialize :metadata
 
   validates :organization_id, presence: true
   validates :seat_type, inclusion: { in: %w[student supervisor] }
--- a/db/migrate/20260618120000_encrypt_license_metadata.rb
+++ b/db/migrate/20260618120000_encrypt_license_metadata.rb
@@ -6,8 +6,10 @@ class EncryptLicenseMetadata < ActiveRecord::Migration[7.2]
   # expected to touch 0 rows; it is written to be safe and idempotent if data
   # does exist.
   #
-  # We operate on the raw column via SQL (not the model) so we are not affected
-  # by the model's decrypting accessor while migrating.
+  # We read the raw column via SQL (not the model) so the decrypting accessor
+  # does not interfere, and write back through update_all (bound/quoted by
+  # ActiveRecord, no string interpolation; update_all bypasses the encrypting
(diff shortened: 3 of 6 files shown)


swahlquist:

## Automated PR Review

### Verdict: `COMMENT`

| --- | --- | --- |

**Senior-dev:** The implementation is structurally sound and the security intent is correct, but two concrete issues need addressing before merge: a SQL injection vector in the spec and a logical inversion in the migration's idempotency guard.

**Adversary:** Potential risk with encryption downgrade migration logic and lack of transaction safety in migration.

### Findings (5 unique)
- **[HIGH]** SQL injection via string interpolation in spec SQL query - spec/models/license_spec.rb lines use "SELECT metadata FROM licenses WHERE id = #{l.id}". Even in specs this pattern is unsafe as a pattern and l.id is an ActiveRecord integer so it works today, but the correct form is License.connection.select_value("SELECT metadata FROM licenses WHERE id = ?", l.i... _(flagged by: senior-dev)_
- **[HIGH]** Migration idempotency guard is logically inverted — skips plaintext, re-encrypts encrypted rows - In up, GoSecure::SecureJson.load(raw) succeeds on **valid plaintext JSON** too (it falls back to JSON.parse), so already_encrypted = true fires for plaintext rows and they are skipped. Only rows that fail to load (corrupt data) would be re-encrypted. The guard should detect whether the blob is alrea... _(flagged by: senior-dev)_
- **[MED]** down migration silently skips rows it cannot decrypt, leaving them encrypted after rollback - If GoSecure::SecureJson.load raises (key rotation, corrupt blob), decrypted is nil and the row is skipped with no warning. After db:rollback the reverted plaintext model will read those rows as raw ciphertext and likely blow up. At minimum a say warning per skipped row and a raised error if any row ... _(flagged by: senior-dev)_
- **[MED]** Migration down may lose data on non-JSON values - The down method in migration attempts to decrypt to JSON, but raw metadata could be a non-JSON string. If GoSecure::SecureJson.load returns a non-hash/array string, to_json will escape it (e.g., '"string"'), changing original plaintext value. _(flagged by: adversary)_
- **[LOW]** Migration not atomic with potential partial updates - Migration uses disable_ddl_transaction! and updates rows individually in a loop. If interrupted, some rows may be encrypted while others remain plaintext, causing inconsistent state. _(flagged by: adversary)_

---

swahlquist:
Addressed the automated review:

- **HIGH "SQL injection in migration UPDATE"** - cleared in bb750709e by switching both UPDATEs to `License.where(id: id).update_all(metadata: ...)`. This was a false positive (`connection.quote` + `id.to_i` are safe and the values are not user-controlled), but `update_all` removes the string interpolation entirely.
- **MED "idempotency check inverts its own logic"** - false positive. The migration calls `GoSecure::SecureJson.load` (the raw primitive, which **raises** on plaintext), not the model's `load_secure_object` (which has the plaintext-JSON fallback). So plaintext -> raises -> `already_encrypted=false` -> re-encrypted. Proven by the passing spec "re-encrypts a legacy plaintext row and is idempotent".
- **MED "race between migration and model writes"** - theoretical only: `metadata` has no writers anywhere in the app, and a concurrently-written value would already be encrypted (correctly skipped by the guard).
- **LOW id.to_i style** - resolved by the `update_all` change.

Spec still green (5 examples, 0 failures).

PR #436, opened 2026-06-19
security(license): encrypt external_reference at rest (LL-740bcb10fa follow-up)

swahlquist:
## What

Follow-up to **#435**: also encrypts `License.external_reference` (a PO Number / Stripe ID) at rest, so **both** sensitive License columns are now encrypted.

**Stacked on #435** (`scot/security/license-metadata-encryption`). Review/merge #435 first; this PR's diff is only the `external_reference` changes.

## How (and why it's a separate mechanism)

`go_secure` allows only **one** `secure_serialize` column per model, and #435 used that slot for `metadata`. So `external_reference` is encrypted **manually** with the same `GoSecure::SecureJson` primitive already used for `device.settings` and `user_integration` secrets — a custom getter/setter on the model:
- **Reader** decrypts, and tolerates legacy plaintext (`rescue -> raw`) so rows written before the backfill still read correctly.
- **Writer** encrypts at assignment.
- **Backfill** (`20260618130000`) re-encrypts any pre-existing plaintext value; raw-SQL, idempotent, reversible, touches **0 rows** today (no writers exist yet).

This is the additive path you approved ("draft the external_reference follow-up"). It does leave License with two encryption mechanisms (secure_serialize for `metadata`, manual for `external_reference`). If you'd rather consolidate to a single encrypted column holding both (architecturally cleaner, but supersedes #435 rather than stacking), say so and I'll draft that instead.

## Tests

`spec/models/license_spec.rb`: round-trip (raw column is not plaintext; model decrypts back), blank stays blank + omitted from JSON, legacy plaintext still reads, `JsonApi::License` serializes it correctly, backfill re-encrypts + idempotent. License spec 10 examples, DB-browser specs 54 examples, 0 failures.

## Finding status

Addresses audit finding **LL-740bcb10fa** (completing both columns). Does **NOT** close it — status stays `open` until you verify and attest.

### DIFF
--- a/app/models/license.rb
+++ b/app/models/license.rb
@@ -5,10 +5,32 @@ class License < ApplicationRecord
   belongs_to :user, optional: true
 
   # LL-740bcb10fa: metadata is an untyped catch-all that may hold sensitive
-  # district/billing data, so it is encrypted at rest. go_secure permits only
-  # one secure column per model; external_reference (a PO/Stripe id) stays
-  # plaintext and is excluded from the DB-browser API in schema_explorer.rb.
+  # district/billing data, so it is encrypted at rest via secure_serialize.
+  # go_secure permits only one secure_serialize column per model, so the
+  # second sensitive field (external_reference) is encrypted manually below.
   secure_serialize :metadata
+
+  # LL-740bcb10fa (follow-up): external_reference is a PO Number / Stripe id.
+  # secure_serialize is already taken by :metadata, so this column is encrypted
+  # manually with the same GoSecure::SecureJson primitive used for device and
+  # user_integration secrets. The reader tolerates legacy plaintext (rescue ->
+  # raw) so existing rows keep reading correctly before the backfill runs; the
+  # writer encrypts at assignment.
+  def external_reference
+    raw = read_attribute(:external_reference)
+    return raw if raw.blank?
+    GoSecure::SecureJson.load(raw)
+  rescue StandardError
+    raw
+  end
+
+  def external_reference=(value)
+    if value.blank?
+      write_attribute(:external_reference, value)
+    else
+      write_attribute(:external_reference, GoSecure::SecureJson.dump(value))
+    end
+  end
 
   validates :organization_id, presence: true
   validates :seat_type, inclusion: { in: %w[student supervisor] }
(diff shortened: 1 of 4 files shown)


swahlquist:

## Automated PR Review

### Verdict: `COMMENT`

| --- | --- | --- |

**Senior-dev:** The manual encrypt/decrypt pattern is sound and consistent with existing GoSecure usage in the codebase. No i18n, feature-flag, or data-isolation issues are introduced.

**Adversary:** Manual encryption may bypass model callbacks and validations; migration has potential data corruption edge cases.

### Findings (5 unique)
- **[MED]** Migration down silently drops non-string encrypted values - In down, when decrypted is not a String it calls .to_json, but if GoSecure::SecureJson.load returns a non-JSON-serializable object or the original plaintext was an integer-like string, the round-trip may not restore the exact original bytes. Low real-world risk since external_reference is always a P... _(flagged by: senior-dev)_
- **[MED]** Direct attribute write bypasses encryption - Code that directly calls license[:external_reference] = value or license.update_column(:external_reference, value) will bypass the manual encrypting setter, leaving plaintext in the DB. _(flagged by: adversary)_
- **[MED]** Migration fails on non-string values - If an external_reference contains a non-string (e.g., a number) and is plaintext, the down migration's plaintext = decrypted.is_a?(String) ? decrypted : decrypted.to_json may produce invalid JSON, corrupting the rollback. _(flagged by: adversary)_
- **[LOW]** Raw SQL in specs uses string-interpolated l.id (integer injection risk in tests) - The test lines SELECT external_reference FROM licenses WHERE id = #{l.id} interpolate an ActiveRecord id directly. Fine for integers today, but inconsistent with the parameterized style used in the migration; worth switching to select_value("...", [[nil, l.id]]) for consistency. _(flagged by: senior-dev)_
- **[LOW]** Exception mask could hide bugs - The rescue StandardError in the getter could silently catch unrelated errors (e.g., JSON parsing errors from corrupted data) and treat them as legacy plaintext, masking data corruption. _(flagged by: adversary)_

---

swahlquist:
Addressed the automated review:

- **HIGH "SQL injection in migration"** - cleared in 87d58a371 by switching both UPDATEs to `License.where(id: id).update_all(external_reference: ...)` (same false-positive reasoning as #435; interpolation now removed).
- **MED "down() loses non-String values"** - not applicable here: `external_reference` is always a scalar string, so the `is_a?(String)` branch always takes the plaintext path. The `.to_json` fallback only matters for the metadata migration (a hash).
- **LOW "race in getter"** - the getter already wraps decryption in `rescue StandardError -> raw`.
- **LOW "manual encryption bypasses audit/key-rotation"** - the known two-mechanism tradeoff. Key rotation is not bypassed (SecureJson uses the same GoSecure keys as secure_serialize); License has no paper_trail, so versioning is not a factor.

Spec still green (10 examples, 0 failures).

PR #437, opened 2026-06-19
Feat/melissa ember 4 12 upgrade

MelissaOneil:


swahlquist:

## Automated PR Review

### Verdict: `COMMENT`

| --- | --- | --- |

**Senior-dev:** Mostly a clean Ember 4 upgrade with good refactors. Two concrete issues need attention before merge: a developer machine IP committed into .ember-cli, and the reopenClass→static-property migration silently breaks any subclass that calls super or uses inheritance-aware class methods on Board/Image/Sound/Video/Organization.

**Adversary:** Multiple changes incompletely migrate from toArray() to slice(), risking missing or undefined functions on RecordArray. New proxy configuration exposes internal IP in committed file.

### Findings (8 unique)
- **[HIGH]** Machine-specific IP address committed to .ember-cli proxy field - .ember-cli now contains "proxy": "http://[REDACTED_IP]:5000". Even if the IP is a LAN address it should not be in a committed file. Per the comment added in the same file, machine-specific hosts belong in .ember-cli.local (already gitignored). This leaks infrastructure topology and will break every ... _(flagged by: senior-dev)_
- **[MED]** reopenClass methods replaced with plain static properties — breaks instanceof/inheritance consumers - Board, Image, Sound, Video, Organization had class-side methods moved from reopenClass({...}) to direct property assignment on the constructor (e.g. LingoLinq.Board.clear_fast_html = function(){}). Any code that calls these via a subclass prototype chain or uses this expecting the class context will... _(flagged by: senior-dev)_
- **[MED]** highlight-outlet observer fires on every 'settings' change without cleanup — potential memory leak - In highlight-outlet.js, syncHighlightControllerModel is an observer('settings', 'highlightController', …). The component uses tagName: '' (glimmer-style) but extends classic Component, so the observer is registered on the instance. If the service model changes frequently (scanning highlight loop) an... _(flagged by: senior-dev)_
- **[MED]** Potential RecordArray toArray/slice migration errors - Replacements of toArray() with slice() assume all objects have a slice method; RecordArrays and possibly other non-array enumerables may not, leading to runtime errors. _(flagged by: adversary)_
- **[MED]** Committed proxy config contains internal IP - Proxy setting in .ember-cli includes a redacted but previously real internal IP address ([REDACTED_IP]), which should not be committed to version control. _(flagged by: adversary)_
- **[LOW]** BaseModel.save() calls this.store.createRecord after a type-resolution failure silently falls through - In base.js save(), if modelNameForRecord(this) returns null the code falls through to this._super() without logging. The original code in app.js used this._internalModel.modelName which always returned a string; the new helper can return null for Ember Data 4 models that expose modelName only on the... _(flagged by: senior-dev)_
- **[LOW]** findBy replaced with find(lambda) may be less efficient - Multiple instances of findBy replaced with find and a lambda function; functionally correct but may have slight performance impact. _(flagged by: adversary)_
- **[LOW]** New component highlight-outlet lacks route integration - Added highlight-outlet component references highlight/highlight2 controllers but no route changes show these outlets are rendered; could be dead code. _(flagged by: adversary)_

---

MelissaOneil:
<html>
<body>
<p style="--tw-space-y-reverse: 0; margin-block: 0px 16px; margin: 0px 0px 6px; word-break: break-word; color: rgb(101, 123, 131); font-family: &quot;Segoe WPC&quot;, &quot;Segoe UI&quot;, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(253, 246, 227); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Addressed all nine audit findings. Here’s how each was handled:</p><h3 class="mt-6 mb-2 font-semibold text-xl" data-streamdown="heading-3" style="--tw-space-y-reverse: 0; margin-block: 0px 16px; margin-bottom: 4px; font-size: 1.214em; line-height: 1.42; --tw-font-weight: 600; font-weight: 590; margin-top: 16px; color: rgb(101, 123, 131); font-family: &quot;Segoe WPC&quot;, &quot;Segoe UI&quot;, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(253, 246, 227); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Fixed in code</h3><div class="ui-scroll-area" data-scroll-padding="4" data-visibility="hover" data-direction="horizontal" style="--tw-space-y-reverse: 0; margin-block: 0px 16px; --scrollbar-size: 6px; --scrollbar-inset: 0px; --scrollbar-thumb-top-offset: 6px; --scroll-area-scroll-padding: 4px; position: relative; overflow: hidden; display: grid; grid-template: 1fr / 1fr; margin: 1em 0px; border-color: color(srgb 0.380392 0.380392 0.380392 / 0.08); border-style: solid; border-width: 0.555556px; border-image: none 100% / 1 / 0 stretch; border-radius: 6px; color: rgb(101, 123, 131); font-family: &quot;Segoe WPC&quot;, &quot;Segoe UI&quot;, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(253, 246, 227); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><div class="ui-scroll-area__viewport" style="grid-area: 1 / 1; min-height: 0px; max-height: 100%; border-radius: inherit; overflow: auto hidden; scroll-padding: 0px 4px; overscroll-behavior: contain auto; scrollbar-width: none !important;"><div class="ui-scroll-area__content" style="box-sizing: border-box; width: auto; min-width: 100%; max-width: 100%; min-height: 100%; display: flex; flex-direction: row; height: fit-content;">
Finding | Resolution
-- | --
[HIGH] BaseModel.save() loses dirty state | save() now uses eachAttribute + record.get() (same scope as old _internalModel._attributes), not serialize().
[HIGH] reload modelName broken | Restored _internalModel.modelName lookup with fallbacks; no longer relies on constructor.modelName alone.
[HIGH] slice() breaks RecordArray behavior | peekAll() loops that only iterate now use .forEach() on the live RecordArray. .slice() kept only where a snapshot is needed (serializer unload loop), with a comment.
[HIGH] findBy → find breaks reactivity | User device computeds now depend on devices.[]; raw JSON uses property access (d.current_device), not .get().
[MED] highlight-outlet computed side-effect | ctrl.set('model', …) moved to an observer; computed only returns the controller lookup.
[MED] BaseModel.save PII risk | eachAttribute limits attrs to declared model fields (same as legacy _attributes), not full serializer output with relationships.
[LOW] CI Node 14 vs README Node 20 | Updated app/frontend/.github/workflows/ci.yml from Node 14 → 20 (root CI was already on 20).

</div></div></div>
</body>
</html>

> > Automated, model-generated review - may be influenced by PR content; advisory only, never auto-merges.
> 
> ## Automated PR Review
> ### Verdict: `COMMENT`
> _Consolidated from two model passes; the more severe verdict wins._
> 
> Pass	Verdict	Raw findings
> Senior-dev (Claude Sonnet 4.6)	`COMMENT`	4
> Adversary (DeepSeek-V3.2, ZDR)	`CONCERNS`	5
> **Senior-dev:** Solid Ember 4.x migration work: DS.Model reopen replaced with BaseModel, toArray/findBy/any/map replaced with slice/find/some, and reopenClass replaced with direct static assignments. Two concrete issues found worth addressing before merge.
> 
> **Adversary:** Multiple high-severity regressions introduced in model reload logic and array access patterns, potentially breaking data isolation and app stability.
> 
> ### Findings (9 unique)
> * **[HIGH]** BaseModel.save() uses serialize() to clone attrs, losing unsaved dirty state fidelity - Old code used this._internalModel._attributes (private API, correctly removed), but replacement uses this.serialize({includeId:false}) which serializes the persisted snapshot, not the in-flight dirty attributes. For a tmp_ record with unsaved changes, the new createRecord will miss those changes, si... _(flagged by: senior-dev)_
> * **[HIGH]** Model reload logic broken due to incorrect modelName reference - BaseModel.reload uses this.constructor.modelName, which is undefined for Ember Data models; should use this._internalModel.modelName or this.constructor.modelName from DS.Model. This will cause force_reload to be incorrectly set and break caching. _(flagged by: adversary)_
> * **[HIGH]** Array method replacements may break Ember Data RecordArray behavior - Multiple instances replacing .toArray() with .slice() on RecordArrays (e.g., in peekAll) lose Ember Data's observable array behaviors and could cause UI updates to fail. _(flagged by: adversary)_
> * **[HIGH]** findBy replaced with custom find breaks reactivity - Device settings and other components replace Ember's .findBy with manual .find loops, losing automatic property observation and potentially causing stale UI state. _(flagged by: adversary)_
> * **[MED]** Hardcoded internal IP in .ember-cli proxy will be committed to repo - .ember-cli 'proxy' key is set to 'http://[REDACTED_IP]:5000'. Even redacted in this diff, if the real IP appears in the committed file it leaks infrastructure topology. This file should use an env var or be in .gitignore for local overrides. _(flagged by: senior-dev)_
> * **[MED]** highlight-outlet.js mutates a looked-up controller inside a computed property - highlightController computed calls ctrl.set('model', settings) as a side-effect inside a computed getter. Side-effects in computed properties cause unpredictable re-entrant invalidation in Ember and violate the computed contract; use an observer or action instead. _(flagged by: senior-dev)_
> * **[MED]** Potential data isolation break in BaseModel.save for tmp records - BaseModel.save serializes with this.serialize({ includeId: false }) which may include sensitive attributes not intended for server, risking PII leakage in offline-to-online sync. _(flagged by: adversary)_
> * **[LOW]** CI workflow pins Node 14 (EOL) while README now requires Node 20+ - .github/workflows/ci.yml uses node-version: 14.x but README.md was updated to require Node.js 20+. CI will pass on an unsupported runtime, masking any Node 20 incompatibilities. _(flagged by: senior-dev)_
> * **[LOW]** Hardcoded IP in proxy configuration - .ember-cli proxy contains a hardcoded redacted IP; while not exploitable, it suggests environment-specific configuration may be committed. _(flagged by: adversary)_
> 
> _Automated review - Claude Sonnet 4.6 (senior-dev) + DeepSeek-V3.2 (adversary, ZDR). Advisory only._


Finding | Resolution
-- | --
[HIGH] BaseModel.save() loses dirty state | save() now uses eachAttribute + record.get() (same scope as old _internalModel._attributes), not serialize().
[HIGH] reload modelName broken | Restored _internalModel.modelName lookup with fallbacks; no longer relies on constructor.modelName alone.
[HIGH] slice() breaks RecordArray behavior | peekAll() loops that only iterate now use .forEach() on the live RecordArray. .slice() kept only where a snapshot is needed (serializer unload loop), with a comment.
[HIGH] findBy → find breaks reactivity | User device computeds now depend on devices.[]; raw JSON uses property access (d.current_device), not .get().
[MED] highlight-outlet computed side-effect | ctrl.set('model', …) moved to an observer; computed only returns the controller lookup.
[MED] BaseModel.save PII risk | eachAttribute limits attrs to declared model fields (same as legacy _attributes), not full serializer output with relationships.
[LOW] CI Node 14 vs README Node 20 | Updated app/frontend/.github/workflows/ci.yml from Node 14 → 20 (root CI was already on 20).

[MED/LOW] Hardcoded IP in .ember-cli | Committed proxy is 127.0.0.1:5000. Added .ember-cli.local to .gitignore + comment for machine-specific overrides.
-- | --
[LOW] Hardcoded IP (duplicate) | Same as above.

Issue #314, opened 2026-05-30
Test harness: ember test does not execute behavioral unit tests (AMD loader defect)


swahlquist:
## Summary

`ember test` does not execute **any** behavioral unit/integration tests. A full run reports ~1369 tests with only 3 behavioral (1 smoke + 2 skipped acceptance); the remaining ~1366 are ESLint/TemplateLint meta-tests. Every behavioral unit spec silently fails to register, so its assertions never run.

## Evidence

A full `ember test` on current `staging` (via a feature branch merged up to staging) shows:

```
behavioral (non-lint) result lines: 3
  ok   - Acceptance | smoke: arithmetic works
  skip - Acceptance | board-detail empty state (x2)
lint result lines: 1366
# tests 1369  # pass 1367  # skip 2  # fail 0
```

Pre-existing committed unit specs that do **not** run despite being imported in `tests/test-helper.js`:

- `tests/unit/controllers/copying-board-test.js` (`Unit | Controller | copying-board`)
- `tests/unit/controllers/user-board-detail-image-cache-test.js`
- `tests/unit/utils/*-test.js` (telemetry, action-lock, eval-*, etc.)

Filtering with `ember test --filter="copying-board"` returns only the ESLint/TemplateLint tests for those files; the behavioral `module()` never matches.

## Root cause (suspected)

`tests/test-helper.js` runs with `start({ loadTests: false })` and relies on a manual loader. Its own comment (lines 21-26) documents the AMD registration-timing problem and points to `tests/README-test-loader.md`. The requirejs sweep (lines 27-45) plus the explicit `import` statements (lines 59-63) are intended to force-register modules, but in practice only the **acceptance** import registers behavioral tests; the **unit** imports do not. The mechanism that distinguishes them needs investigation.

## Impact

- No unit-level regression protection. New unit tests added to PRs (for example #313 for issue #293) are lint-checked but their assertions never execute, giving false confidence.
- CI green does not mean unit tests passed; it means lint passed.

## Suggested next steps

1. Confirm whether `loadTests: true` (ember-qunit's own loader) registers the unit specs, and why the manual loader does not.
2. If the AMD timing issue is real on Ember 3.28, find a registration approach that covers unit specs (not just acceptance), or move to a supported loader.
3. Add a guard test that asserts a known unit spec actually ran, so this regression cannot recur silently.

## Notes

[...]

PR #445, opened 2026-06-19
fix(flags): gate supervisor-consent UI and eval/quick route behind their flags

swahlquist:
## What
Two feature-flag coverage quick wins from the audit (`outputs/docs/2026-06-19-feature-flag-coverage-audit.md`). No data exposure in either; these close disruptive-UI / defense-in-depth gaps where the backend already enforced the flag.

## GAP B: supervisor-consent UI ungated on the frontend
`pending-consent-requests` rendered unconditionally in `supervision-settings` (component template lines 68 and 270) even though the backend enforces `supervisor_consent_flow` (OFF by default). A supervisor in an org without the flow saw a dead-end "pending consent requests" UI whose actions fail server-side.

Both include sites are now wrapped in `{{#if this.appState.feature_flags.supervisor_consent_flow}}`.

The `/consent/:token` route is **intentionally left reachable** so any already-sent consent-email links keep working. Only the settings-page UI is gated.

## GAP C: `eval/quick` route had no route-level guard
Entry links and the API already gate `quick_screen_eval`, but a direct URL rendered the eval shell before the API rejected it. Added a `beforeModel` guard (via the `app-state` + `router` services) that redirects home when the flag is off. Defense-in-depth only.

## Validation
- ESLint (project v5.16.0) and ember-template-lint clean on both files under Node 20.
- Diff against the pristine baseline: **no new rule violations introduced**.
- Note: `ember test` only runs lint in this repo (AMD loader defect, issue #314); no behavioral unit run is possible for template/route guards.

## Not included (needs a product decision)
GAP A (board-import flag mismatch, `paste_html_import` vs `multi_user_board_import`) is held for a product call, per the audit. GAP D (four TEMPORARY-ON flags) is a release-gate checklist item.

### DIFF
--- a/app/frontend/app/routes/eval/quick.js
+++ b/app/frontend/app/routes/eval/quick.js
@@ -1,11 +1,23 @@
 import Route from '@ember/routing/route';
 import RSVP from 'rsvp';
+import { inject as service } from '@ember/service';
 import LingoLinq from '../../app';
 import EvalSession from '../../utils/eval_session';
 import i18n from '../../utils/i18n';
 
 export default Route.extend({
   title: i18n.t('quick_screen_eval_title', "Quick Screen"),
+  appState: service('app-state'),
+  router: service('router'),
+
+  beforeModel(/* transition */) {
+    // Defense-in-depth: entry links and the API already gate quick_screen_eval, but
+    // guard the route too so a direct URL does not render the eval shell before the
+    // API rejects it. Redirect home when the flag is off for this user.
+    if(!this.appState.get('feature_flags.quick_screen_eval')) {
+      this.router.transitionTo('index');
+    }
+  },
 
   model(params) {
     if (params.user_id) {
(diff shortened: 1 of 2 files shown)


swahlquist:

## Automated PR Review

### Verdict: `APPROVE`

| --- | --- | --- |
| Senior-dev (Claude Sonnet 4.6) | `APPROVE` | 1 |

**Senior-dev:** Both changes correctly gate UI behind feature flags. The template guards are straightforward, and the beforeModel guard is a sound defense-in-depth addition with no regressions.

**Adversary:** PR adds frontend guards for supervisor-consent UI and quick eval route as intended; no security risks introduced.

### Findings (1 unique)
- **[MED]** beforeModel guard does not return the transition promise, allowing model() to run - When the flag is off, this.router.transitionTo('index') is called but its promise is not returned. Ember will continue into model() before the redirect resolves, potentially triggering unnecessary API calls or errors. Add return this.router.transitionTo('index'); to abort the hook chain. _(flagged by: senior-dev)_

---

PR #449, opened 2026-06-19
compliance(infra): Terraform for the incident-evidence S3 bucket (breach runbook §4.1 step 4)

swahlquist:
## What

Adds a reviewable Terraform module under `infra/incident-evidence-bucket/` that provisions the write-once forensic evidence bucket the breach runbook (`docs/legal/BREACH_RUNBOOK.md` §4.1 step 4, open-gap #2 in §12) references but which does **not exist yet**. Closes the Bucket-B incident-evidence task at the code level.

## What it provisions (once applied)

- `lingolinq-incident-evidence`, **Object Lock COMPLIANCE mode, 7-year default retention**
- Versioning ON, all four public-access blocks ON, `BucketOwnerEnforced` (ACLs off)
- SSE-S3 (AES-256) default; SSE-KMS optional via a variable
- Bucket policy: TLS-only; `PutObject`/read allowed **only** to the Incident Commander (Scot) and Tech Lead (Melissa) principals; write denied to everyone else
- MFA-delete intentionally omitted (COMPLIANCE-mode lock supersedes it, per runbook)

## ⚠️ NOT YET APPLIED — and not appliable from here

`terraform apply` is a **deliberate, admin-credentialed, irreversible** step, not part of this PR:

1. **COMPLIANCE-mode Object Lock is irreversible.** A written object cannot be deleted or shortened by anyone (incl. AWS root) for 7 years. A mistaken upload is stuck and billed the whole time. The module sets `prevent_destroy` and ships a throwaway-GOVERNANCE `test.tfvars` + `validate.sh` so the lock/policy are proven in a disposable bucket **first**.
2. **The app credentials cannot do it.** Verified live: the only identity available (`lingolinq-app` IAM user, acct `239044785114`) has no IAM access and cannot create an Object-Lock bucket or attach a bucket policy. Apply needs an admin principal.
3. **Three decisions + two ARNs are required from Scot before apply:** D1 region (default us-west-2, co-located with prod data), D2 lock mode (COMPLIANCE vs GOVERNANCE — irreversibility acceptance), D3 encryption (SSE-S3 vs SSE-KMS), plus the Incident Commander + Tech Lead IAM principal ARNs for `write_principal_arns`.

## Verification

- `terraform fmt` clean, `terraform validate` passes (TF v1.14.6, AWS provider >= 5.40).
- No bucket, role, or policy was created. Live AWS state was only **read** (caller identity, bucket list, bucket regions).

Full D1-D5 rationale: `~/ai-company-brain/outputs/plans/2026-06-19-incident-evidence-bucket-provisioning-plan.md`.

### DIFF
--- a/infra/incident-evidence-bucket/.gitignore
+++ b/infra/incident-evidence-bucket/.gitignore
@@ -1,0 +1,7 @@
+# Never commit local TF state or filled-in vars (may carry account ARNs / secrets).
+*.tfstate
+*.tfstate.*
+.terraform/
+.terraform.lock.hcl
+terraform.tfvars
+crash.log
--- a/infra/incident-evidence-bucket/README.md
+++ b/infra/incident-evidence-bucket/README.md
@@ -98,8 +98,9 @@ open-gap #2 closed.
 - **D4 name:** `lingolinq-incident-evidence`
 - **D5 method:** Terraform (this module)
 
-These are the module defaults, so `terraform.tfvars` only needs the two
-assume-role principal ARNs filled in
-(`incident_commander_trusted_principal_arns` = Scot,
-`tech_lead_trusted_principal_arns` = Melissa). Full rationale:
+These are the module defaults. `terraform.tfvars.example` is now pre-filled
+with the two assume-role principal ARNs for this account (`239044785114`):
+`wahlquist-admin` -> Incident Commander, `melissa-oneil` -> Tech Lead. Confirm
+both usernames are correct at apply time (the apply fails clearly if either ARN
+does not resolve). Full rationale:
 `~/ai-company-brain/outputs/plans/2026-06-19-incident-evidence-bucket-provisioning-plan.md`.
--- a/infra/incident-evidence-bucket/README.md
+++ b/infra/incident-evidence-bucket/README.md
@@ -1,0 +1,92 @@
+# Incident-evidence S3 bucket (Terraform)
+
+Provisions the write-once, tamper-proof forensic evidence bucket required by
+`docs/legal/BREACH_RUNBOOK.md` §4.1 step 4 and open-gap #2 (§12). The runbook
+hardcodes the name `s3://lingolinq-incident-evidence`; its commands assume this
+bucket exists before the first district contract redline.
+
+## What it creates
+
+| Resource | Setting |
+|----------|---------|
+| S3 bucket | `lingolinq-incident-evidence`, Object Lock enabled at creation |
+| Object Lock | `COMPLIANCE` mode, 7-year default retention (Decision D2) |
+| Versioning | Enabled |
+| Encryption | SSE-S3 (AES-256) by default; SSE-KMS optional (Decision D3) |
+| Public access | All four blocks ON |
+| Ownership | `BucketOwnerEnforced` (ACLs disabled) |
+| Bucket policy | TLS-only; write/read allowed ONLY to the Incident Commander + Tech Lead principals; PutObject denied to everyone else |
+
+MFA-delete is intentionally NOT configured: COMPLIANCE-mode Object Lock blocks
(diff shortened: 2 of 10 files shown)


swahlquist:

## Automated PR Review

### Verdict: `COMMENT`

| --- | --- | --- |

**Senior-dev:** The Terraform module is well-structured and the security controls are sound. Two concrete issues worth addressing before apply: real IAM usernames for named staff are committed to the example tfvars (PII/OPSEC concern), and the validate.sh Object Lock check has a logic bug that will always report PASS even when no lock is configured.

**Adversary:** The PR adds a well-documented Terraform module for a compliance-critical S3 bucket but contains a high-severity logic error: the bucket policy's 'DenyWriteFromOthers' condition incorrectly denies all writes, blocking the intended principals.

### Findings (5 unique)
- **[HIGH]** Bucket policy denies all PutObject, blocking intended writers - In main.tf, the 'DenyWriteFromOthers' statement uses condition StringNotEquals with aws:PrincipalArn against a list of ARNs. The condition will evaluate to true for any principal, including those in local.effective_write_principals, because StringNotEquals does not work with lists; the condition fai... _(flagged by: adversary)_
- **[MED]** Real IAM usernames for named individuals committed to example tfvars - terraform.tfvars.example and test.tfvars.example hardcode arn:aws:iam::239044785114:user/wahlquist-admin and arn:aws:iam::239044785114:user/melissa-oneil plus the account ID 239044785114. Staff ARNs tied to real names in a public AGPLv3 repo are a social-engineering/account-enumeration surface and e... _(flagged by: senior-dev)_
- **[MED]** validate.sh Object Lock check always passes regardless of config - Line: [ "$mode" != "NONE" ] && check ... ok || check ... bad. Because check always returns 0, the || branch never executes; the bad path is unreachable. An unconfigured bucket reports PASS. Fix: evaluate the condition first, then call a single check invocation with the result string. _(flagged by: senior-dev)_
- **[MED]** IAM policy missing s3:DeleteObject for bypass in GOVERNANCE mode - data.aws_iam_policy_document.evidence_rw lacks s3:BypassGovernanceRetention. In GOVERNANCE mode (test.tfvars), the admin running validation cannot delete test objects via terraform destroy or manual deletion without this permission, breaking cleanup flow. _(flagged by: adversary)_
- **[LOW]** validate.sh does not handle missing logging bucket gracefully - If enable_access_logging = true and log_target_bucket is absent/invalid, terraform apply will fail. However, the validation script's check for TLS-only policy uses a fragile grep that could match unrelated text. _(flagged by: adversary)_

---

swahlquist:
## Update: dedicated incident IAM roles added (per Scot)

The module now creates two dedicated, MFA-gated roles and auto-wires their **stable ARNs** into the bucket policy, rather than putting personal user ARNs into a 7-year-locked policy:

- `lingolinq-incident-commander` (assumed by Scot) and `lingolinq-tech-lead` (assumed by Melissa)
- Each assumable **only with MFA** (`require_mfa_to_assume`, default on) by the principals you name
- Each carries a scoped identity policy granting read/write of **this bucket only**
- Bucket policy writers = the two role ARNs; a precondition requires at least one writer

**Why roles over user ARNs:** staff changes (someone leaves, a new Tech Lead) only touch a role's trust policy, never the immutable bucket policy. That's the whole point of not hardcoding people into a 7-year lock.

**Apply-time fill-in is now just two values:** `incident_commander_trusted_principal_arns` (Scot's IAM user ARN) and `tech_lead_trusted_principal_arns` (Melissa's). `create_incident_roles = false` falls back to the supply-your-own-`write_principal_arns` model.

The throwaway test path sets `create_incident_roles = false` so `validate.sh --probe-write` can upload as the caller directly.

`terraform fmt` clean, `terraform validate` passes. Decisions confirmed: COMPLIANCE / SSE-S3 / us-west-2.

PR #452, opened 2026-06-19
compliance(tooling): register -> Notion findings DB sync script

swahlquist:
## One-way register -> Notion findings DB sync

Part of the new-compliance-dev handoff. The findings register (`audit-reports/FINDINGS.json`) stays the single source of truth; this script mirrors it into a Notion database so a dev gets a board-style view that never drifts from code.

- Idempotent upsert keyed by LL-id: PATCH existing rows, create new, `--prune` archives findings removed from the register.
- `--dry-run` validates the mapping with no token (confirmed: reads all 79 findings).
- Requires `NOTION_TOKEN` (Notion internal integration secret, stored in 1Password) + `NOTION_FINDINGS_DB_ID`. To make it unattended, wire to a Render cron.
- Compliance content is Claude-only; the register is PII-free (code/path evidence), so nothing identifiable reaches Notion.

Companion to the seeded "LingoLinq Compliance Findings (LL)" Notion database and the Compliance Engineering onboarding/handoff page.

### DIFF
--- a/.github/workflows/sync-findings-to-notion.yml
+++ b/.github/workflows/sync-findings-to-notion.yml
@@ -1,0 +1,33 @@
+name: Sync findings register to Notion
+
+# Keeps the "LingoLinq Compliance Findings (LL)" Notion board in sync with the SSOT
+# (audit-reports/FINDINGS.json) whenever the register changes on staging. One-way:
+# register -> Notion. Only writes register-owned columns; human-owned columns
+# (Owner, Target date, Program notes, Needs Scot decision) are never touched.
+# Compliance content stays Claude-only; the register is PII-free (code/path evidence).
+
+on:
+  push:
+    branches: [staging]
+    paths:
+      - 'audit-reports/FINDINGS.json'
+  workflow_dispatch: {}
+
+concurrency:
+  group: notion-findings-sync
+  cancel-in-progress: false
+
+jobs:
+  sync:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: '3.4.3'
+          bundler-cache: false
+      - name: Sync register -> Notion board
+        env:
+          NOTION_TOKEN: ${{ secrets.NOTION_TOKEN }}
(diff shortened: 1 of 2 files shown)


swahlquist:

## Automated PR Review

### Verdict: `COMMENT`

| --- | --- | --- |

**Senior-dev:** The workflow hardcodes the Notion database ID in plaintext and the sync script contains an unguarded HTTP persistent-connection singleton that leaks across requests in edge cases. Both are easily fixable.

**Adversary:** Script may unintentionally overwrite Notion content; lacks robust error handling for Notion API failures and edge cases in input data.

### Findings (6 unique)
- **[MED]** Notion DB ID hardcoded in workflow YAML instead of a secret - NOTION_FINDINGS_DB_ID is set to 1f8451c4a17b4f5b868878ac4386b805 in the workflow file committed to the public repo. Even though this is an AGPLv3 public fork, exposing the live Notion DB ID makes it trivial for anyone to enumerate or attempt to query the database if an integration token is ever comp... _(flagged by: senior-dev)_
- **[MED]** No validation for Notion response content - Line 133 assumes 'results' array always present in API response. Notion could return nil for 'results', causing NoMethodError on .each. _(flagged by: adversary)_
- **[MED]** Missing idempotency key handling - If two syncs run concurrently (concurrency group cancels false), they may race on POST/PATCH operations, causing duplicate rows or inconsistent state. _(flagged by: adversary)_
- **[LOW]** http singleton not reset on connection error; persistent conn may replay stale state - The @​http instance is memoised with ||= and never closed or re-opened on failure. If the underlying TCP connection is reset mid-sync, the memoised object retains the broken socket and every subsequent api() call raises a low-level IO error rather than reconnecting. Add error rescue with @​http = ni... _(flagged by: senior-dev)_
- **[LOW]** Hardcoded Notion database ID in workflow - Line 29 of YAML embeds DB_ID 1f8451c4a17b4f5b868878ac4386b805, making fork/deployment inflexible and leaking internal Notion identifier. _(flagged by: adversary)_
- **[LOW]** Prune archives without safety check - --prune archives rows removed from register without confirmation prompt or backup, risking accidental data loss if register is malformed. _(flagged by: adversary)_

---

PR #468, opened 2026-06-23
Traci/styling/styling updates

traciday:
🏠 Supervisor home redesign · Ember 4.12 migration · preference-aware board routing
Overview
This PR delivers a redesigned supervisor home dashboard, completes the Ember 3.28 → 4.12 migration, and makes board navigation honor each user's view-style preference, plus a handful of focused UI refinements.

ℹ️ Scope note: this long-lived styling branch has merged staging back in many times, so a lot of earlier work (the create-board redesign, the guided-tour system, the board-detail responsive overhaul, and the server-side security hardening) already shipped through staging and is intentionally not re-listed here. Everything below is what's genuinely new vs staging.

✨ Supervisor dashboard / home redesign (headline)
A role-aware home for supervisors, supporters, and admins — cards, hierarchy, and tiles instead of a flat list.

New cards in authenticated-view.hbs: Rooms (md-card--rooms, tiles → organization.room), Communicators Need Attention (md-card--attention, flagged supervisees with avatar + status pill), and My Organizations reworked from a single click-through button into per-org tiles (role · rooms · plan).
Layout engine — dashboard_sections.js: adds rooms + attention sections, a new communicatorsNeedingAttention() helper + ATTENTION_STATUS_IDS, a supervisor-specific default order, full-width packing for Boards/Caseload/Rooms/Attention/Org, and a corrected hasOrgManagement (supporters alone no longer trigger the Org card).
Header/data — authenticated-view.js: computes role label (Admin/Manager/Supervisor), naturally-sorted rooms (Intl.Collator), first-4 + overflow, and the attention/orgs lists.
New Dashboard::GlowAvatar (glow-avatar.js): samples a communicator's avatar, extracts a saturation-weighted representative color, and emits --avatar-glow-* CSS vars for a soft per-user glow (with a safe fallback for cross-origin-tainted images).
Backend support — lib/json_api/user.rb: new org_status_for(user) derives no-home-board vs tree-deciduous, now emitted in the supervisor branch so it survives the /supervisees refetch that fires for 10+ communicators (otherwise the largest caseloads silently emptied the Attention card). app.js registers the derived no-home-board status.
Responsive SCSS cascade at ≤1024 / 768 / 640 / 460 px + prefers-reduced-motion; 6 new optimized PNG card/tile art assets.
⬆️ Ember 4.12 migration (mechanical, no behavior change)
ember-data / ember-source → ~4.12, eslint-plugin-ember 11.x, swapped to ember-template-lint (drives the large package-lock.json churn).
23 Ember-Data models migrated DS.* → @ember-data/model RFC-395 imports — verified import-only, no logic changes.
New .eslintrc.js pinning the 11.x ruleset for this classic (non-Octane) app — Octane-architecture rules off, real-bug/deprecation rules kept as warnings.
Deprecation fixes: transitionToRoute → router.transitionTo, component mouseEnter hook → didInsertElement DOM listener, and highlight-outlet switched off the Ember-4-rejected object-first (action obj "name") form to bound function wrappers.
🧭 Preference-aware board navigation
New board_view.js → board_view_route(user): opens board-detail (modern) by default, board-alt (classic) only when preferences.board_view_style === 'classic'.
Routed 5 entry points through it: the catch-all board route, board-icon & board-preview overlays, the dashboard board strip (goToBoard), and the org default-home-board link. Board creation/import flows intentionally stay on board-detail (edit flow needs it).
📐 board-detail: short-screen scrolling
Extended the existing square-collapse rule to https://github.com/media (max-width: 1024px), (max-height: 700px) so wide-but-short windows keep square buttons and scroll

vertically — preserving the original tight button gaps instead of cramming/stretching the grid.
🎨 Create-board refinements
AI-mode intro checklist (replaces the AI subtitle), paint color labels with an aria-live selected-color readout, and preview chips that respect the user's text-position so they match the live grid.
📝 Record-note modal restyle
Converted the "New Note" modal from legacy Bootstrap (form-horizontal / col-sm-) to the modern md- modal system — same functionality (topic, status, text/video toggle), new look.
🌐 i18n
~43 new en keys (dashboard/supervisor, create-board, import/export) regenerated across 13 locales.

Triaged all 8 adversarial findings against the code:

Fixed (3):

#7 room-count pluralization → now uses i18n.t('n_rooms', "room", {count}).
#8 glow onload now guards against stale/torn-down callbacks.
#6 added a CSS.supports('color', …) check before applying a custom paint color (note: it was already .trim()d, and the primary control is <input type="color">, but this hardens the hand-typed hex field).
False positives (4):

#2 (CSS injection) — the interpolated values are all numeric: rgb is hslToRgb() → Math.round ints (0–255), strong/soft are .toFixed(3). Output is always rgba(<int>,<int>,<int>,<float>); hostile pixels change only the numbers, no CSS syntax is injectable.
#5 (img.src XSS) — assigning a URL to .src never executes javascript:/data: as script; an  isn't a navigation context. Not a sink.
#4 (observer before attach) — didInsertElement also calls _applyGlow() after attach, so the glow renders on initial paint; the observer only handles later src changes.
#3 (org_status tenant scope) — the endpoint scopes by user.supervisees (the existing supervision authz), and org_status_for only reflects each supervisee's own home-board/org state, which the supervisor is already entitled to. No cross-tenant exposure.
#1 (feature flag) — accepting as-is: the flag policy primarily protects communicator-facing AAC speak surfaces or features we want to gate behind a purchased product; this is the supervisor home (clinicians/parents) and the cards extend an already-established surface rather than introducing a new gated feature.


swahlquist:

## Automated PR Review

| --- | --- | --- |
| Adversary (DeepSeek-V3.2, ZDR) | `BLOCK` | 8 |

**Senior-dev:** The PR is a large, well-structured set of changes. Most findings are mechanical and low-risk, but two concrete issues stand out: a new user-facing status label missing i18n indirection at registration time, and a glow-avatar observer that fires before element attachment on src changes.

**Adversary:** Critical security regression in glow-avatar exposes cross-origin pixel leakage; multiple permission checks missing for new dashboard data; race condition in org_status_for can corrupt user objects.

### Findings (12 unique)
- **[HIGH]** Cross-origin pixel theft via getImageData in glow-avatar - glow-avatar.js line 91 calls getImageData on canvas with user-controlled avatar URLs (src). This violates same-origin policy and will throw SecurityError for non-CORS images, but for CORS images it leaks pixel data to client JavaScript. This is a security regression - avatar pixels should never be a... _(flagged by: adversary)_
- **[HIGH]** Missing authorization for org_status_for endpoint - lib/json_api/user.rb org_status_for exposes supervisee home-board/org state but lacks validation that requesting user is supervisor of those supervisees. Attacker could enumerate user IDs to determine who has home boards set. _(flagged by: adversary)_
- **[HIGH]** Race condition in org_status_for can corrupt user objects - org_status_for modifies user objects in-place (adds org_status). If same user requested concurrently by different supervisors, shared object mutations cause data corruption. Must return new derived object. _(flagged by: adversary)_
- **[MED]** glow-avatar srcChanged observer may fire before didInsertElement - The srcChanged observer on src can trigger _applyGlow() before didInsertElement has run (e.g. when Ember sets attrs during component init). At that point this.element is null, so _applyGlow returns early and the glow is never applied. The PR description claims didInsertElement covers initial paint, ... _(flagged by: senior-dev)_
- **[MED]** no-home-board status label evaluated at module-load time outside i18n ready - In app.js the new entry {id: 'no-home-board', label: i18n.t('no_home_board_status', "No home board set")} is evaluated when the user_statuses array literal is constructed at module load. All other status objects in that array use no label field — labels are looked up dynamically later. If i18n is no... _(flagged by: senior-dev)_
- **[MED]** Missing permission check for rooms data - authenticated-view.js exposes supervised_units to all users via computed properties without verifying user has supervisor role. Normal communicators shouldn't see rooms card or count. _(flagged by: adversary)_
- **[MED]** PII leakage via attentionCommunicators - attentionCommunicators exposes avatar_url, user_name, and status labels for all supervisees to any supervisor. Need to verify each supervisee hasn't opted out of supervisor visibility (privacy settings). _(flagged by: adversary)_
- **[MED]** Insecure direct object reference in roomsAllOrgId - roomsAllOrgId uses first room's organization_id for 'view all' link. If user has rooms in multiple orgs but only access to some, link may direct to unauthorized org page. _(flagged by: adversary)_
- **[LOW]** attentionCommunicators exposes avatar_url and user_name of supervisees in computed snapshot - The computed attentionCommunicators emits avatar_url (a potentially signed S3 URL) and user_name into a plain JS object that is bound into the template. If the template ever serializes or logs this array (e.g. Ember Inspector, error reporting), it creates an unintentional PII surface. This is low ri... _(flagged by: senior-dev)_
- **[LOW]** board strip cap changed from 5 to 4 without feature flag - The home board strip now shows at most 4 boards (home + 3 favorites) plus the Crisis tile, down from 5+Crisis. This silently removes a tile for existing users with 5 favorites on their dashboard — a concrete behavior change for AAC communicators that should be behind a feature flag or at minimum a p... _(flagged by: senior-dev)_
- **[LOW]** XSS vector via board_view_route user preference - board_view_route reads user.preferences.board_view_style without sanitization. If attacker can inject malicious value (e.g., via API), could manipulate route transitions. Should validate against whitelist ['modern','classic']. _(flagged by: adversary)_
- **[LOW]** Missing input validation for paint color label - set_paint_mode accepts label parameter without sanitization (line 2251). Label displayed in UI could contain HTML/JS if called from compromised code path. Should escape HTML. _(flagged by: adversary)_

---

Issue #5, opened 2025-09-18
Priority: Ember Framework Upgrade (3.12 → Modern Version)


swahlquist:
## Problem Statement

Current Ember 3.12 (September 2019) is causing significant development friction:
- **5+ years of technical debt** blocking modern development
- **Node.js compatibility issues** requiring Docker isolation
- **Deprecated Bower dependencies** with security vulnerabilities
- **Poor local development experience** with frequent hanging/crashes

## Business Impact

- **Reduced Developer Velocity**: Legacy tooling slows development 25-40%
- **Security Risk**: Outdated dependencies with known vulnerabilities  
- **Maintenance Overhead**: 50%+ time spent on compatibility issues
- **Recruitment Challenges**: Difficult to find developers familiar with old versions

## Proposed Solution

**Incremental Upgrade Path:**
```
Ember 3.12 → 3.28 (LTS) → 4.12 (LTS) → 5.x (Latest)
```

### Phase 1: 3.12 → 3.28 (2 weeks)
- Minimal breaking changes
- Security patches
- Better Node.js support

### Phase 2: 3.28 → 4.12 (4 weeks)  
- Modern JavaScript features
- Performance improvements
- Node.js 18 support

### Phase 3: 4.12 → 5.x (6 weeks)
- Latest features
- Complete NPM transition
- Future-proof architecture

## Current State

- **Branch**: `fix/ci-pipeline-test` contains JavaScript namespace fixes
- **Documentation**: Comprehensive upgrade research in `.ai/docs/EMBER_UPGRADE_RESEARCH.md`
- **Local Development**: Currently requires Docker due to version constraints

## Resources Required

- **Timeline**: ~3.75 months for full upgrade
- **Skills**: Ember expertise, modern JavaScript, testing
- **Testing**: Comprehensive validation at each phase

## Acceptance Criteria

- [ ] Local development works without Docker constraints
- [ ] Modern Node.js/NPM compatibility (18+)
- [ ] All existing functionality preserved
- [ ] Improved build times and developer experience
- [ ] Security vulnerabilities resolved

## Next Steps

1. **Create upgrade branch**: `upgrade/ember-modernization`
2. **Audit current dependencies** and breaking changes
3. **Set up comprehensive testing** strategy
4. **Begin Phase 1**: Upgrade to Ember 3.28 LTS

## Context

[...]

PR #469, opened 2026-06-23
Traci/styling/styling updates

traciday:
🏠 Supervisor home redesign · Ember 4.12 migration · preference-aware board routing
Overview
This PR delivers a redesigned supervisor home dashboard, completes the Ember 3.28 → 4.12 migration, and makes board navigation honor each user's view-style preference, plus a handful of focused UI refinements.

ℹ️ Scope note: this long-lived styling branch has merged staging back in many times, so a lot of earlier work (the create-board redesign, the guided-tour system, the board-detail responsive overhaul, and the server-side security hardening) already shipped through staging and is intentionally not re-listed here. Everything below is what's genuinely new vs staging.

✨ Supervisor dashboard / home redesign (headline)
A role-aware home for supervisors, supporters, and admins — cards, hierarchy, and tiles instead of a flat list.

New cards in authenticated-view.hbs: Rooms (md-card--rooms, tiles → organization.room), Communicators Need Attention (md-card--attention, flagged supervisees with avatar + status pill), and My Organizations reworked from a single click-through button into per-org tiles (role · rooms · plan).
Layout engine — dashboard_sections.js: adds rooms + attention sections, a new communicatorsNeedingAttention() helper + ATTENTION_STATUS_IDS, a supervisor-specific default order, full-width packing for Boards/Caseload/Rooms/Attention/Org, and a corrected hasOrgManagement (supporters alone no longer trigger the Org card).
Header/data — authenticated-view.js: computes role label (Admin/Manager/Supervisor), naturally-sorted rooms (Intl.Collator), first-4 + overflow, and the attention/orgs lists.
New Dashboard::GlowAvatar (glow-avatar.js): samples a communicator's avatar, extracts a saturation-weighted representative color, and emits --avatar-glow-* CSS vars for a soft per-user glow (with a safe fallback for cross-origin-tainted images).
Backend support — lib/json_api/user.rb: new org_status_for(user) derives no-home-board vs tree-deciduous, now emitted in the supervisor branch so it survives the /supervisees refetch that fires for 10+ communicators (otherwise the largest caseloads silently emptied the Attention card). app.js registers the derived no-home-board status.
Responsive SCSS cascade at ≤1024 / 768 / 640 / 460 px + prefers-reduced-motion; 6 new optimized PNG card/tile art assets.
⬆️ Ember 4.12 migration (mechanical, no behavior change)
ember-data / ember-source → ~4.12, eslint-plugin-ember 11.x, swapped to ember-template-lint (drives the large package-lock.json churn).
23 Ember-Data models migrated DS.* → @ember-data/model RFC-395 imports — verified import-only, no logic changes.
New .eslintrc.js pinning the 11.x ruleset for this classic (non-Octane) app — Octane-architecture rules off, real-bug/deprecation rules kept as warnings.
Deprecation fixes: transitionToRoute → router.transitionTo, component mouseEnter hook → didInsertElement DOM listener, and highlight-outlet switched off the Ember-4-rejected object-first (action obj "name") form to bound function wrappers.
🧭 Preference-aware board navigation
New board_view.js → board_view_route(user): opens board-detail (modern) by default, board-alt (classic) only when preferences.board_view_style === 'classic'.
Routed 5 entry points through it: the catch-all board route, board-icon & board-preview overlays, the dashboard board strip (goToBoard), and the org default-home-board link. Board creation/import flows intentionally stay on board-detail (edit flow needs it).
📐 board-detail: short-screen scrolling
Extended the existing square-collapse rule to https://github.com/media (max-width: 1024px), (max-height: 700px) so wide-but-short windows keep square buttons and scroll

vertically — preserving the original tight button gaps instead of cramming/stretching the grid.
🎨 Create-board refinements
AI-mode intro checklist (replaces the AI subtitle), paint color labels with an aria-live selected-color readout, and preview chips that respect the user's text-position so they match the live grid.
📝 Record-note modal restyle
Converted the "New Note" modal from legacy Bootstrap (form-horizontal / col-sm-) to the modern md- modal system — same functionality (topic, status, text/video toggle), new look.
🌐 i18n
~43 new en keys (dashboard/supervisor, create-board, import/export) regenerated across 13 locales.

Triaged all 8 adversarial findings against the code:

Fixed (3):

https://github.com/lingolinq/LingoLinq-AAC/pull/7 room-count pluralization → now uses i18n.t('n_rooms', "room", {count}).
https://github.com/lingolinq/LingoLinq-AAC/pull/8 glow onload now guards against stale/torn-down callbacks.
https://github.com/lingolinq/LingoLinq-AAC/pull/6 added a CSS.supports('color', …) check before applying a custom paint color (note: it was already .trim()d, and the primary control is , but this hardens the hand-typed hex field).
False positives (4):

https://github.com/lingolinq/LingoLinq-AAC/pull/2 (CSS injection) — the interpolated values are all numeric: rgb is hslToRgb() → Math.round ints (0–255), strong/soft are .toFixed(3). Output is always rgba(,,,); hostile pixels change only the numbers, no CSS syntax is injectable.
https://github.com/lingolinq/LingoLinq-AAC/issues/5 (img.src XSS) — assigning a URL to .src never executes javascript:/data: as script; an  isn't a navigation context. Not a sink.
https://github.com/lingolinq/LingoLinq-AAC/pull/4 (observer before attach) — didInsertElement also calls _applyGlow() after attach, so the glow renders on initial paint; the observer only handles later src changes.
https://github.com/lingolinq/LingoLinq-AAC/pull/3 (org_status tenant scope) — the endpoint scopes by user.supervisees (the existing supervision authz), and org_status_for only reflects each supervisee's own home-board/org state, which the supervisor is already entitled to. No cross-tenant exposure.
https://github.com/lingolinq/LingoLinq-AAC/pull/1 (feature flag) — accepting as-is: the flag policy primarily protects communicator-facing AAC speak surfaces or features we want to gate behind a purchased product; this is the supervisor home (clinicians/parents) and the cards extend an already-established surface rather than introducing a new gated feature.

Round-2 triage — all 5 are false positives or the already-accepted repeat:

#1 (canvas "leaks" cross-origin data) — the opposite is true: getImageData() throws SecurityError on a tainted canvas before reading any pixel, so it only succeeds when the avatar server sends CORS headers (we set crossOrigin='anonymous'). The result never leaves the client — it's reduced to one averaged RGB written as a CSS var. Non-CORS avatars just fall back to the SCSS glow. Added a comment documenting this.
#4 (avatar PII) — same: client-side averaging of an image the supervisor is already authorized to view; nothing is stored, sent, or newly disclosed.
#3 (attention data-isolation) — the card maps user.supervisees (the backend's already-scoped /supervisees). Authorization is the supervision relationship, not org membership; a multi-org supervisor correctly sees their whole caseload.
#5 (board_view_route permission) — that helper only picks a view style (returns a route name); board access is enforced when the route loads the board (server-side allowed?). A routing helper is the wrong layer for an authz check.
#2 (feature flag) — same as the prior round; accepting as-is since this is the supervisor home, not a communicator speak surface. @Scot to close.


swahlquist:

## Automated PR Review

### Verdict: `COMMENT`

| --- | --- | --- |

**Senior-dev:** The PR is a large, well-structured migration with good i18n hygiene. Two issues are worth flagging before merge: a silent early-return in apply_custom_paint_color that breaks the existing Custom color workflow on browsers without CSS.supports, and a board-strip behavior change (cap reduced from 5→4 starred/other boards before Crisis append) that alters an existing UI without a feature flag.

**Adversary:** The PR introduces a new component (glow-avatar.js) that performs image analysis on user avatars via canvas, potentially risking cross-origin data leakage if CORS misconfigured. Additionally, the board routing helper relies on client-side preference without server-side revalidation, which could allow unauthorized board access if route-level auth is missing.

### Findings (5 unique)
- **[HIGH]** Potential cross-origin data leakage in glow-avatar - The glow-avatar component uses getImageData on a canvas after loading a cross-origin image with crossOrigin='anonymous'. If the server does not send appropriate CORS headers, the canvas becomes tainted and throws a SecurityError. However, if the server incorrectly sends CORS headers for avatars that... _(flagged by: adversary)_
- **[MED]** CSS.supports guard silently swallows valid colors on older browsers - apply_custom_paint_color now returns early when window.CSS || window.CSS.supports is falsy. On browsers that lack CSS.supports (e.g. older WebViews used by some AAC hardware), every valid custom color is silently dropped — no error, no feedback. The input type=color already constrains browser-picker... _(flagged by: senior-dev)_
- **[MED]** Board strip cap silently changed from 5 to 4 (breaking existing layout) - The ordered-boards cap was reduced from 5 to 4 before the Crisis tile append, so the strip now shows 5 total instead of 6. This is an unguarded behavior change to an existing surface used by communicators. Per the PR's own feature-flag policy, communicator-facing layout changes should be behind a fl... _(flagged by: senior-dev)_
- **[MED]** Board routing relies on client-side preference without auth revalidation - The board_view_route helper selects view style based on user.preferences.board_view_style, but board access authorization depends on server-side checks when the route loads. If route-level auth is missing or bypassed, a user could access boards they shouldn't by manipulating the preference or route.... _(flagged by: adversary)_
- **[LOW]** glow-avatar observer fires before element is attached on first render - srcChanged observer calls _applyGlow() whenever src changes, including the initial binding before didInsertElement. At that point this.element is null, so _applyGlow returns early — the guard is correct, but didInsertElement must also call _applyGlow() to cover the first paint. The existing code doe... _(flagged by: senior-dev)_

---

traciday:
Round-2 triage — all 5 are false positives or the already-accepted repeat:

#1 (canvas "leaks" cross-origin data) — the opposite is true: getImageData() throws SecurityError on a tainted canvas before reading any pixel, so it only succeeds when the avatar server sends CORS headers (we set crossOrigin='anonymous'). The result never leaves the client — it's reduced to one averaged RGB written as a CSS var. Non-CORS avatars just fall back to the SCSS glow. Added a comment documenting this.
#4 (avatar PII) — same: client-side averaging of an image the supervisor is already authorized to view; nothing is stored, sent, or newly disclosed.
#3 (attention data-isolation) — the card maps user.supervisees (the backend's already-scoped /supervisees). Authorization is the supervision relationship, not org membership; a multi-org supervisor correctly sees their whole caseload.
#5 (board_view_route permission) — that helper only picks a view style (returns a route name); board access is enforced when the route loads the board (server-side allowed?). A routing helper is the wrong layer for an authz check.
#2 (feature flag) — same as the prior round; accepting as-is since this is the supervisor home, not a communicator speak surface. @Scot to close.

Round-3 triage — 4 false positives + 1 accepted:

#1 (glow cross-origin "leak") — same as the prior round, re-framed. Even when CORS permits the read, it's the supervisee avatar the supervisor is already viewing, and only an averaged color is computed client-side and never transmitted. There's a code comment documenting this now.
#2 (CSS.supports drops colors on old browsers) — misreads the condition: if(window.CSS && window.CSS.supports && !window.CSS.supports('color', color)) return;. When CSS.supports is absent, the guard is skipped and the color is applied — it only rejects invalid colors when the API exists. No valid color is ever dropped.
#4 (board_view_route auth) — same as the prior round: it's a pure view-style selector returning a route name; board access is enforced when the route loads the board (server-side allowed?).
#5 (observer before attach) — didInsertElement also calls _applyGlow() after attach, so the first paint is covered (as the finding itself notes the guard is correct).
#3 (board strip 5→4) — intentional: the strip is home + up to 3 boards + Crisis = 5 tiles, tuned for the ≤1024px responsive rule that keeps the home + Crisis tiles when the others collapse. Minor change to the home navigation hub (not the speak surface). Accepting as-is. The design change was intentional.

MelissaOneil (approved):

PR #472, opened 2026-06-23
feat(cutover): Render write-reject mode for Cloud Run cutover freeze (4.WR1)

swahlquist:
## What

Adds an env-gated Rack middleware (`WriteFreeze::Middleware`) for the Phase 5 Render -> GCP Cloud Run cutover. When the operator sets `WRITE_FREEZE`, data-mutating requests return **HTTP 503 + `Retry-After`** so no late or stale-DNS write lands on the soon-to-be-abandoned Render DB during the cutover freeze and soak. Reads pass through untouched (users can still view and speak boards) and a minimal **auth/session allowlist** keeps sign-in working so users are not locked out for the duration of the soak.

**Default (`WRITE_FREEZE` unset): zero behavior change.** The middleware is always in the stack but inert until the operator sets the var at freeze start and unsets it on rollback.

Implements the decided write-freeze mechanism (Option 1 + 2: 60s DNS TTL + Render kept up in write-reject mode). Runbook step 1: `scripts/gcp/PHASE5-CUTOVER-RUNBOOK.md`. Plan: `~/ai-company-brain/outputs/plans/2026-06-23-cloudrun-write-reject-mode-plan.md`.

## Design

- **Rack middleware, not a `before_action`** so it uniformly covers the `lib/json_api` write paths and any non-controller endpoints before routing. Mirrors the existing `Throttling` pattern: a testable module of constants/logic + `config.middleware.use` in a reopened `Application`.
- **Denylist-by-default + auth allowlist.** `GET/HEAD/OPTIONS` always pass. `POST/PUT/PATCH/DELETE` are rejected unless the path matches the auth allowlist (login / token / refresh / SAML / google SSO). Password reset is **deliberately excluded** (writes a reset token lost on the abandoned DB; not needed to sign in with existing creds) and documented inline.
- **`WRITE_FREEZE` ENV gate, not a feature flag.** This is an operator-controlled infra maintenance mode, not a user-facing product feature, so it is env-gated (analogous to other env-gated infra behavior) rather than an `AVAILABLE_FRONTEND_FEATURES` entry. The only user-facing surface is the maintenance response, whose copy is routed through i18n. (Flagging this explicitly against the feature-flag rule for review; if you prefer routing through `lib/feature_flags.rb`, that is a small follow-up.)
- **503 body** is i18n'd (`config/locales/en.yml`, double-quoted, no em dashes) and content-negotiates JSON (`{error, retry_after}`, matching the existing `{error: ...}` envelope) vs an HTML maintenance page.

## Tests

- `spec/initializers/write_freeze_paths_spec.rb` — classification (enabled?/mutating?/allowlisted?/reject?).
- `spec/features/write_freeze_spec.rb` — end-to-end through the full Rack stack: 503 + `Retry-After`, read + auth pass-through, JSON vs HTML negotiation, per-verb rejection.
- **17 examples, 0 failures.** Throttling specs still green (no middleware-ordering regression).

## Out of scope / follow-ups

- DNS TTL change (ops step, runbook step 1).
- Offline-client 503 + `Retry-After` handling verification (dress-rehearsal checklist item).
- Dress-rehearsal coverage sweep enumerating mutating routes to confirm data-writes 503 while reads/auth work.

### DIFF
--- a/config/initializers/write_freeze.rb
+++ b/config/initializers/write_freeze.rb
@@ -1,0 +1,136 @@
+# Cutover write-freeze (Render -> GCP Cloud Run migration, Phase 5).
+#
+# When the ops env var WRITE_FREEZE is set, the Rack middleware below rejects
+# data-mutating requests with HTTP 503 + Retry-After so that no late or
+# stale-DNS write lands on the soon-to-be-abandoned Render database during the
+# cutover freeze and soak. Reads pass through untouched (users can still view
+# and speak their boards), and a minimal auth allowlist keeps sign-in working so
+# users are not locked out for the duration of the soak.
+#
+# This is an operator-controlled infrastructure maintenance mode, not a
+# user-facing product feature, so it is gated by an ENV var rather than an
+# AVAILABLE_FRONTEND_FEATURES flag (analogous to other env-gated infra behavior
+# in this app). The only user-facing surface is the maintenance response, whose
+# human-readable copy is routed through i18n (config/locales/en.yml).
+#
+# Default (WRITE_FREEZE unset): zero behavior change. The middleware is always
+# in the stack but is inert until the operator sets WRITE_FREEZE at freeze start
+# and unsets it on rollback. See scripts/gcp/PHASE5-CUTOVER-RUNBOOK.md step 1.
+module WriteFreeze
+  # Mutating HTTP verbs. GET / HEAD / OPTIONS always pass through.
+  MUTATING_METHODS = %w[POST PUT PATCH DELETE].freeze
+
+  # Seconds to advertise in the Retry-After header. The offline/sync client
+  # should treat 503 + Retry-After as "retry later" (against current DNS, which
+  # the 60s TTL points at GCP), not "drop the write".
+  RETRY_AFTER_SECONDS = 120
+
+  # Auth/session routes that MUST stay writable during the freeze, so existing
+  # users can still sign in. Each entry is an unanchored regex fragment matched
+  # against the request path (mirrors Throttling::PROTECTED_PATHS), so
+  # 'oauth2/token' also covers 'oauth2/token/login'. Keep this list MINIMAL:
+  # authentication only, never data writes.
+  #
+  # Deliberately NOT allowlisted: forgot_password / password_reset. Those write
+  # a reset token that would be lost on the abandoned Render DB and are not
+  # needed to sign in with existing credentials; a user landing on fresh DNS
+  # resets via GCP. All other data mutations (board saves, button/image
+  # uploads, LogSession creation, profile/goal/settings writes) are rejected by
+  # design - those are exactly the writes that must not diverge from Cloud SQL.
+  ALLOWLIST_PATHS = [
+    'oauth2/token',          # session#oauth_token / oauth_login / oauth_logout
+    '^/token',               # session#token (browser password grant)
+    'wait/token',            # session#token_wait
+    'api/v1/token/refresh',  # session#oauth_token_refresh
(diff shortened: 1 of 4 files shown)


swahlquist:

## Automated PR Review

### Verdict: `COMMENT`

| --- | --- | --- |

**Senior-dev:** The middleware is well-structured and the i18n, ENV-gate, and test coverage are solid. Two issues are worth a comment before merge: an over-broad allowlist regex that can be path-traversal-abused, and a missing Cache-Control/Pragma header that may cause the 503 to be cached by CDN/proxy during the soak.

**Adversary:** Middleware introduces regex-based path matching security risks and potential for missed side-effect routes. Critical for cutover safety.

### Findings (6 unique)
- **[MED]** ALLOWLIST_RE matches substrings, so a crafted path can bypass the freeze - 'oauth2/token' in ALLOWLIST_PATHS is joined with | and matched as an unanchored substring. A request to /api/v1/boards/oauth2/token_id (or any path embedding an allowlist fragment) passes allowlisted? and skips the freeze. Anchor each entry or match only against well-defined path segments (e.g. path... _(flagged by: senior-dev)_
- **[MED]** 503 responses carry no Cache-Control header; CDNs/proxies may cache them - Without 'Cache-Control: no-store' (or at minimum 'no-cache') in the rejection headers, an intervening CDN or reverse proxy (Render's own edge, Cloudflare, nginx) can cache the 503 and continue serving it to users after the freeze is lifted, requiring a manual cache purge. Add 'Cache-Control' => 'no-... _(flagged by: senior-dev)_
- **[MED]** Regex allowlist vulnerable to path traversal bypass - ALLOWLIST_PATHS uses regex fragments that could be bypassed with crafted paths containing allowlist strings as substrings (e.g., '/evil/oauth2/token' or '/token/../boards'). Need anchors or strict path matching. _(flagged by: adversary)_
- **[MED]** Missing coverage for non-api side-effect writes - SIDE_EFFECT_GET_PATHS list may be incomplete; other GET endpoints performing writes (e.g., /admin/impersonate, webhooks) could bypass freeze and corrupt abandoned DB. Relies on manual audit. _(flagged by: adversary)_
- **[LOW]** Potential race condition on ENV read - WriteFreeze.enabled? reads ENV at each request. If operator unsets WRITE_FREEZE mid-flight, some requests may incorrectly 503. Acceptable for maintenance mode but noted. _(flagged by: adversary)_
- **[LOW]** JSON/HTML content negotiation may mismatch Accept - wants_json? uses accept.include?('application/json') which matches 'application/json+something' but also matches 'text/html, application/json' incorrectly for browsers. Could serve JSON to browser navigation. _(flagged by: adversary)_

---

swahlquist:
**Claude adversary (red-team) review — verdict: ship, no Critical/High.**
Used the Claude adversary pass as review-of-record (HIPAA/COPPA-adjacent cutover code stays Claude-only per repo policy; Codex `/review-pr` skipped to avoid its branch-flipping auto-committer in a shared checkout).

Verified SAFE: genuinely inert when `WRITE_FREEZE` unset; `config.middleware.use` placement covers all `lib/json_api` controller writes; no method-override verb-down bypass (Rack::MethodOverride runs first); all 10 existing-user auth POST routes pass the allowlist (sign-in not over-blocked); JSON-vs-HTML negotiation never feeds a sync client HTML.
Non-blocking conditions for the window (not merge blockers): (1) the side-effect-GET denylist is allow-by-default and misses cache-writing GETs (buttonset/protected_image regenerate cache rows) — the real backstop is a read-only Postgres role on the abandoned Render DB during the soak; (2) confirm the runbook enumerates the accepted-loss auth/password-reset writes.

PR #483, opened 2026-06-24
docs(runbook): close maintenance-message checklist item (no prod users) + fix stale merge notes

swahlquist:
## What

Phase 5 cutover runbook (`scripts/gcp/PHASE5-CUTOVER-RUNBOOK.md`) pre-cutover checklist housekeeping. Docs-only, no code.

## Why

Render **prod has no real clients or users** at cutover (only a few internal/fake test accounts; prod will be brought up to staging, then migrated to GCP). So the runbook's *proactive* maintenance-window announcement is unnecessary. The only maintenance surface that matters is the **reactive** WriteFreeze 503 page (PR #472, i18n'd `write_freeze.title`/`write_freeze.body`), which is already built and merged and is sufficient for the fake-account / offline-replay edge cases.

## Changes

- **Maintenance message (i18n) checklist item** → marked `[x]`: satisfied by the #472 503 page; no proactive announcement needed. Records a **re-open condition** if real users are onboarded to prod before the window (no in-app banner mechanism exists yet).
- **Step-1 "announce the window" bullet** → annotated as skipped for this cutover (original copy retained for future use), so the doc is not internally contradictory.
- **Stale merge annotations** → "pending merge to staging" corrected to "merged to staging" on **PR #472** and **PR #473** (both merged). This is the one-line touch-up flagged last session, folded in here.

## Not changed

No code, no i18n strings, no behavior. The reactive 503 copy from #472 is untouched.

PR #484, opened 2026-06-24
feat(gcp): pre-DNS Render-vs-Cloud-SQL delta-check script (Phase 5 runbook step 7)

swahlquist:
## What

Adds `scripts/gcp/phase5-delta-check.sh` - the **pre-DNS delta check** from the Phase 5 cutover runbook (step 7), plus the runbook references that name it. No app code; operator tooling only.

## Why

Immediately before the DNS flip, the only thing that catches a writer that slipped past the write-freeze is comparing the **live Render** DB against the **freshly-restored Cloud SQL** DB. The step-0a/3 reconciliation only compares dump-to-restore, so it cannot see writes that reached Render *after* the fresh dump. **Any non-zero delta is a rollback trigger.**

## Behavior

- **Forward (gate, default):** compares `MAX(id)` + `MAX(updated_at)` (optionally `COUNT(*)` with `--counts`) on `log_sessions`, `boards`, `board_contents` (override via `TABLES=`). Exit `0` = match (safe to flip); exit non-zero = drift (do NOT flip DNS).
- **Reverse (`--since <flip-ts>`):** enumerates Cloud SQL rows newer than the DNS-flip time - the cutover-window writes that exist only on GCP, for replay/merge back to Render on rollback (Rollback step 3). Report-by-default; `--fail-on-rows` to gate.
- **Strictly read-only:** every query runs in `BEGIN READ ONLY` + `default_transaction_read_only=on` - verified the transaction *physically rejects* an INSERT - so it is safe against live prod (HIPAA). Never echoes the DSN/password. Models `phase4-setval-sequences.sh` conventions.

## Review

Claude adversary red-team pass (Claude-only - HIPAA-adjacent cutover tooling, **not** Codex). Verdict **SHIP-WITH-FIXES**; all merge-blockers addressed in commit `d2b4a9b`:
- forward mode refuses identical Render/Cloud SQL endpoints (false-OK) unless `ALLOW_SAME_DB=1`
- `to_char` timestamp normalization (Render PG vs Cloud SQL PG18 format-divergence false-DRIFT)
- absent-table = drift `[ABSENT]`; reverse-mode table-coverage + numeric-count assertions (no silent under-report)
- `TABLES` identifier validation (can't escape the read-only transaction)
- header warns never to run under `bash -x` (DSN/password leak)

Pushed back on the adversary's "reverse should default to fail-on-rows": runbook Rollback step 3 is explicitly replay-or-accept-loss (operator judgment), so the report stays report-by-default; the real risk is closed by the coverage assertion.

## Dry-run evidence (local, dev/test DBs, 2026-06-24)

| Scenario | Result |
|---|---|
| same-DB without override | refused, exit 1 |
| `ALLOW_SAME_DB=1`, identical data (`--counts`) | all `[OK]`, exit 0 |
| distinct DBs | `[DRIFT]`/`ROLLBACK TRIGGER`, exit 1 |
| empty table both sides | `(none)` -> `[OK]` |
| reverse, rows newer (+`--fail-on-rows`) | reports total, exit 2 |
| reverse, none newer | "nothing to reconcile", exit 0 |
| read-only write attempt | `cannot execute INSERT in a read-only transaction` |
| invalid `TABLES` / missing env | refused |

The **live** pre-DNS run against Render + Cloud SQL remains a gated cutover step.

## Note

Touches `PHASE5-CUTOVER-RUNBOOK.md` step 7 + checklist; no line overlap with open PR #483 (runbook housekeeping), so both auto-merge cleanly.

https://claude.ai/code/session_01KiZM2j7J1bYAMTc2Ug9wjL

### DIFF
--- a/scripts/gcp/PHASE5-CUTOVER-RUNBOOK.md
+++ b/scripts/gcp/PHASE5-CUTOVER-RUNBOOK.md
@@ -323,6 +323,22 @@ Then, BEFORE any DNS change:
   trigger** (it means an external writer or offline replay slipped past the freeze; go fix the
   freeze, do not flip DNS).
 
+  **Script: `scripts/gcp/phase5-delta-check.sh`** (built, strictly read-only - every query runs in
+  `BEGIN READ ONLY`, so it is safe against live prod). Forward (gate) mode exits non-zero on any
+  drift:
+
+  ```
+  cloud-sql-proxy --port 5432 lingolinq-prod:us-central1:lingolinq-prod-pg &
+  RENDER_DATABASE_URL='postgres://USER:PASS@RENDER_HOST:5432/lingolinq_production' \
+  CLOUDSQL_DATABASE_URL='postgres://lingolinq_app:PASS@127.0.0.1:5432/lingolinq_production' \
+    ./scripts/gcp/phase5-delta-check.sh           # add --counts to also compare COUNT(*) (slower)
+  ```
+
+  Default tables are `log_sessions boards board_contents` (override with `TABLES=`). On **rollback**
+  (Rollback step 3), the same script in reverse mode enumerates the cutover-window writes that
+  landed only on Cloud SQL, for replay/merge back into Render:
+  `CLOUDSQL_DATABASE_URL=... ./scripts/gcp/phase5-delta-check.sh --since '<DNS-flip timestamp>'`.
+
 ### 8. Front-end decision gate  (tracker 5.3 - DECIDED 2026-06-23)
 
 The web service currently deploys with `--allow-unauthenticated` straight to the `run.app` URL.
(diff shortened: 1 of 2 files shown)


swahlquist:

## Automated PR Review

### Verdict: `COMMENT`

| --- | --- | --- |

**Senior-dev:** Senior-dev (Claude) pass skipped: the diff contained secret-shaped content (credential), which is not sent to the external model endpoint. Rotate any real credential and re-run after removing it from the diff.

**Adversary:** Adversary (DeepSeek) pass skipped: the diff contained secret-shaped content (credential), which is not sent to the no-BAA OpenRouter endpoint. Rotate any real credential and re-run after removing it from the diff.

### Findings (0 unique)
_No findings._

---

swahlquist:
### Adversary review verdict (Claude-only, HIPAA-adjacent)

**SHIP-WITH-FIXES** — all findings addressed in `d2b4a9b`. Disposition:

| # | Sev | Finding | Disposition |
|---|-----|---------|-------------|
| 1 | High | reverse `total` arithmetic could abort/under-report on malformed probe | **Fixed** — numeric guard on COUNT before summing (exit 3 on non-numeric) |
| 2 | High | reverse defaults to non-gating report, no coverage check | **Partially accepted** — added table-coverage assertion (missing table → exit 3); **kept report-by-default** (runbook Rollback step 3 = replay-or-accept-loss, an operator judgment call) |
| 3 | Med | forward both-MISSING == OK (silent dropped table) | **Fixed** — absent table is `[ABSENT]` = drift |
| 4 | Med | `MAX(updated_at)::text` cross-version format → false DRIFT | **Fixed** — `to_char(...,'YYYY-MM-DD HH24:MI:SS.US')` |
| 5 | Med | `TABLES` identifier injection could escape read-only txn | **Fixed** — validate `^[a-z_][a-z0-9_]*$` |
| 6 | Low | `bash -x` leaks DSN password | **Fixed** — header warning |
| 7 | Low | same DSN both sides → false "safe to flip" | **Fixed** — forward refuses identical endpoints unless `ALLOW_SAME_DB=1` |

**Operational note (not a code defect):** `statement_timeout` on a large `MAX(updated_at)` scan aborts cleanly (not a false-OK) but would kill the gate mid-window. Before the live run, confirm `updated_at` is indexed on `log_sessions` (both sides) or raise `STATEMENT_TIMEOUT`.

Core gate is sound: hard connection/SQL failures abort under `set -e` (cannot produce a false-OK); the read-only transaction was verified to physically reject writes.

MelissaOneil (approved):
gtg

PR #486, opened 2026-06-25
Traci/styling/styling updates

traciday:

🏠 5. Home-page tour polish

🌍 9. Translations
<details> <summary>🔧 Technical notes for reviewers</summary>
New code: components/sidebar-editor.{js,hbs} (the panel), utils/tours/board-detail-speak.js (the speak tour), registered in tours/registry.js.
Backend: app/models/user.rb#merge_missing_default_sidebar_boards rewritten to preserve stored order (append-only).
Sidebar default-show: new app-state.js#effective_quick_sidebar is the single source of truth (treats an unset preference as "show"); consumed by application.js, board-detail.js, and app-state.js sidebar computeds.
Loading overlay: app-state.js#arm_board_load_overlay + _debounceBoardLoadOverlay, called from new-board.js, import-from-html.js, import-from-json-bundle.js, create-board-new.js, style-switcher.js.
Symbol fix: new #ll-symbol-white-backing SVG filter in index.html + paired app.scss rules.
Tour hand-offs: new pending-flag watchers in guided-tour.js (board_picker_tour_pending, board_detail_tour_pending_speak); in-body tour buttons via shared.js#tourBodyButton.
Styling: ~770 new lines in app.scss (md-sidebar-editor__*, md-board-detail-inline-sidebar__*, md-tour__body-*).
Cleanup: template event handlers modernized from {{action}} to {{on "click" (action)}} (removes Ember deprecation warnings).
Merge conflicts (app.scss, LEARNINGS.md, 13 locales) were resolved keeping this branch's new content plus staging's additions (room_lc/rooms_lc, goal_status_prompt); all locale JSON validated as parseable.
GitHub will show many commits already in staging by content (earlier squash-merges) — the Files changed tab (43 files) is the true delta.
</details>


swahlquist:

## Automated PR Review

### Verdict: `COMMENT`

| --- | --- | --- |
| Senior-dev (Claude Sonnet 4.6) | `COMMENT` | 6 |

**Senior-dev:** The PR is a large, well-structured feature addition with good i18n hygiene and safety guards. A small number of concrete issues need attention before merging: a missing feature-flag gate on the sidebar-editor surface, a silent dedup false-positive that can misroute users to the wrong board, and a hard page-reload that can break unsaved session state.

**Adversary:** Potential race conditions in board copying logic and missing error handling for duplicate sidebar entries.

### Findings (8 unique)
- **[HIGH]** _findExistingUserCopy trusts key convention without parent lineage — can return wrong board - When parentId and parentKey are both absent, the code returns found unconditionally (the !parentId && !parentKey branch). A different board that happens to share the slug under the user's namespace (e.g. two public boards with the same slug from different orgs) will be silently reused as the home bo... _(flagged by: senior-dev)_
- **[HIGH]** sidebar-editor new UI surface is not behind a feature flag - The new sidebar-editor component and the 'Edit Sidebar' button are unconditionally rendered for all users in speak mode. Per project policy, new user-facing behavior must be feature-flag gated because AAC users find unexpected UI changes disruptive. Wrap with the existing feature_flag / has_feature ... _(flagged by: senior-dev)_
- **[MED]** back action calls window.location.reload() — destroys unsaved session state - In sidebar-editor.js _reloadSpeakPage, a hard reload is used to refresh the sidebar after changes. This silently discards any unsaved work the user had in the speak session (e.g. a partially built utterance). A targeted data refresh (re-fetching user preferences and re-rendering) or a soft route ref... _(flagged by: senior-dev)_
- **[MED]** _paintPreparingOverlay called without a transition — overlay may never be dismissed on cancel path - In board-preview-overlay.js _paintPreparingOverlay, paint_view_switch_overlay is called with no transition function. If _handlePickError fails to remove the overlay DOM node (e.g. the element id differs or the remove throws), users are stranded behind a full-screen overlay with no way to dismiss it.... _(flagged by: senior-dev)_
- **[MED]** sidebar_boards written directly on the user model without FERPA/HIPAA data-isolation check - In _save, user.set('preferences.sidebar_boards', newArray) and user.save() are called using appState.setup_user as the subject when present. If setup_user is a student/patient record in a district account, saving sidebar preferences through it from a non-privileged session could violate data-isolati... _(flagged by: senior-dev)_
- **[MED]** Race condition in board copy deduplication - Both board-preview-overlay.js and sidebar-editor.js have independent _findExistingUserCopy functions that query the store for existing copies. Concurrent copy operations could create duplicates if both checks run before either copy completes. _(flagged by: adversary)_
- **[MED]** Missing conflict resolution for sidebar duplicates - _pushEntry in sidebar-editor.js pushes new entries without checking for duplicates beyond the initial lookup.keys check. A race could add the same board multiple times to the sidebar array. _(flagged by: adversary)_
- **[LOW]** In-body tour action buttons use a mutable expando property (_tourActionWired) on DOM nodes - In guided-tour.js _onTourStepShow, btn._tourActionWired = true is set as a non-standard expando on a DOM element. Shepherd may re-create the step element on re-show, making the guard stale and missing re-wiring, or the property may survive element reuse and skip re-wiring on a legitimately new eleme... _(flagged by: senior-dev)_

---

PR #487, opened 2026-06-25
Traci/styling/styling updates

traciday:

🏠 5. Home-page tour polish

🌍 9. Translations

🔧 Technical notes for reviewers

✅ Fixed (3 real issues)
[HIGH] #1 — wrong-board reuse from slug-trust + [MED] #6 — duplicated dedup logic
The old _findExistingUserCopy returned a found board even when it had no parent lineage (!parentId && !parentKey), so an unrelated board the user happened to own at the same slug could be silently set as their home board. I extracted one shared helper, [board-copy.js](vscode-webview://15dvltuo394spu1l3ka3t79t2gec5meji3e5ggofbe037osvoou5/app/frontend/app/utils/board-copy.js#L25), that reuses a copy only on a positive parent_board_id/parent_board_key match, and fetches with { reload: true } so the parent fields are authoritative (a list-partial could otherwise hide them and cause a needless duplicate). Both [board-preview-overlay.js](vscode-webview://15dvltuo394spu1l3ka3t79t2gec5meji3e5ggofbe037osvoou5/app/frontend/app/components/board-preview-overlay.js) and [sidebar-editor.js](vscode-webview://15dvltuo394spu1l3ka3t79t2gec5meji3e5ggofbe037osvoou5/app/frontend/app/components/sidebar-editor.js) now call it — killing the divergent second implementation that finding #6 flagged.

[HIGH] #2 — sidebar editor not flag-gated
Added an edit_sidebar flag to [feature_flags.rb](vscode-webview://15dvltuo394spu1l3ka3t79t2gec5meji3e5ggofbe037osvoou5/lib/feature_flags.rb#L31) (AVAILABLE = canonical beta opt-in; temporarily ENABLED with a REMOVE-before-merge note, same pattern as home_tour), and gated both the inline Edit button and the <SidebarEditor> panel render in [board-detail.hbs](vscode-webview://15dvltuo394spu1l3ka3t79t2gec5meji3e5ggofbe037osvoou5/app/frontend/app/templates/user/board-detail.hbs) behind feature_flags.edit_sidebar.

[MED] #7 — _pushEntry duplicate race
[_pushEntry](vscode-webview://15dvltuo394spu1l3ka3t79t2gec5meji3e5ggofbe037osvoou5/app/frontend/app/components/sidebar-editor.js#L355) now re-checks the live array for the key before pushing, so a rapid/concurrent add can't seat the same board twice.

🔧 Hardened (1)
[LOW] #8 — DOM expando — Replaced the btn._tourActionWired expando with a module-level WeakSet in [guided-tour.js](vscode-webview://15dvltuo394spu1l3ka3t79t2gec5meji3e5ggofbe037osvoou5/app/frontend/app/components/guided-tour.js#L15) (no DOM mutation; new elements wire, reused elements skip, entries GC with their nodes).

📋 Refuted with evidence (3 — no change needed)
#3 reload loses unsaved message — False. The working utterance is continuously persisted to localStorage as working_vocalization ([utterance.js:314](vscode-webview://15dvltuo394spu1l3ka3t79t2gec5meji3e5ggofbe037osvoou5/app/frontend/app/utils/utterance.js#L314)) and rehydrated on boot ([utterance.js:36](vscode-webview://15dvltuo394spu1l3ka3t79t2gec5meji3e5ggofbe037osvoou5/app/frontend/app/utils/utterance.js#L36)), so window.location.reload() restores it.
#4 overlay never dismisses — False. view_switch_overlay arms an unconditional 8s safety timer + a routeDidChange dismissal, and the error path explicitly removes the id-matched node. Three independent dismissal paths.
#5 FERPA/HIPAA via setup_user — False. _save writes appState.currentUser's own prefs ([sidebar-editor.js:309](vscode-webview://15dvltuo394spu1l3ka3t79t2gec5meji3e5ggofbe037osvoou5/app/frontend/app/components/sidebar-editor.js#L309)); setup_user is only used to query which boards to list, never as the save subject. user.save() is permission-checked server-side.


swahlquist:

## Automated PR Review

### Verdict: `COMMENT`

| --- | --- | --- |

**Senior-dev:** The PR is well-structured and addresses previous review findings, but has three concrete issues: a feature-flag left permanently enabled, a window.location.reload() that discards in-progress unsaved work, and a slug-only dedup match in the sidebar editor that can still surface a false positive.

**Adversary:** Potential race condition in board copy reuse logic and missing authorization check for sidebar editor access.

### Findings (6 unique)
- **[HIGH]** edit_sidebar flag is ENABLED (not just AVAILABLE) — ships to all users - feature_flags.rb adds edit_sidebar to ENABLED, not just AVAILABLE. The PR description explicitly says 'REMOVE-before-merge note' but the diff shows it still enabled. AAC users find unexpected UI changes disruptive; this must be AVAILABLE-only before merge. _(flagged by: senior-dev)_
- **[MED]** _reloadSpeakPage calls window.location.reload() discarding any unsaved utterance state - sidebar-editor.js _reloadSpeakPage does a hard reload after any sidebar change. Although utterance.js persists the working vocalization to localStorage, any other transient speak-page state (e.g. partially-built phrases not yet flushed) could be lost. The PR description refutes #3 for board-preview-... _(flagged by: senior-dev)_
- **[MED]** Slug-only match in _current_lookup can mark an unrelated board as already-on-sidebar - sidebar-editor.js _current_lookup tracks bare slugs (key.split('/').pop()) and ordered_brands hides catalog boards whose slug matches lookup.slugs. Two unrelated boards from different orgs sharing a slug (e.g. 'core') would cause one to be hidden from the Add list even though neither is actually on ... _(flagged by: senior-dev)_
- **[MED]** Potential race condition in board copy lookup - In board-preview-overlay.js, findExistingUserCopy is called but the result is used to decide whether to reuse or copy. If the user creates a copy elsewhere between the lookup and the copy operation, a duplicate could still be created despite dedup intent. _(flagged by: adversary)_
- **[MED]** Missing server-side authorization check for edit_sidebar flag - The sidebar-editor component gates UI behind feature_flags.edit_sidebar, but server-side endpoints (e.g., saving sidebar_boards) should also validate the flag to prevent API abuse if client-side check is bypassed. _(flagged by: adversary)_
- **[LOW]** WeakSet may not prevent all double-binding - guided-tour.js uses a WeakSet to track wired buttons, but if the same element is removed and re-added to the DOM (not just hidden/shown), it could be re-wired because the WeakSet entry might be garbage collected. _(flagged by: adversary)_

---

PR #488, opened 2026-06-25
Traci/styling/styling updates

traciday:

🏠 5. Home-page tour polish

🌍 9. Translations

🔧 Technical notes for reviewers

✅ Fixed (3 real issues)
[HIGH] https://github.com/lingolinq/LingoLinq-AAC/pull/1 — wrong-board reuse from slug-trust + [MED] https://github.com/lingolinq/LingoLinq-AAC/pull/6 — duplicated dedup logic
The old _findExistingUserCopy returned a found board even when it had no parent lineage (!parentId && !parentKey), so an unrelated board the user happened to own at the same slug could be silently set as their home board. I extracted one shared helper, board-copy.js, that reuses a copy only on a positive parent_board_id/parent_board_key match, and fetches with { reload: true } so the parent fields are authoritative (a list-partial could otherwise hide them and cause a needless duplicate). Both board-preview-overlay.js and sidebar-editor.js now call it — killing the divergent second implementation that finding https://github.com/lingolinq/LingoLinq-AAC/pull/6 flagged.

[HIGH] https://github.com/lingolinq/LingoLinq-AAC/pull/2 — sidebar editor not flag-gated
Added an edit_sidebar flag to feature_flags.rb (AVAILABLE = canonical beta opt-in; temporarily ENABLED with a REMOVE-before-merge note, same pattern as home_tour), and gated both the inline Edit button and the panel render in board-detail.hbs behind feature_flags.edit_sidebar.

[MED] https://github.com/lingolinq/LingoLinq-AAC/pull/7 — _pushEntry duplicate race
_pushEntry now re-checks the live array for the key before pushing, so a rapid/concurrent add can't seat the same board twice.

🔧 Hardened (1)
[LOW] https://github.com/lingolinq/LingoLinq-AAC/pull/8 — DOM expando — Replaced the btn._tourActionWired expando with a module-level WeakSet in guided-tour.js (no DOM mutation; new elements wire, reused elements skip, entries GC with their nodes).

📋 Refuted with evidence (3 — no change needed)
https://github.com/lingolinq/LingoLinq-AAC/pull/3 reload loses unsaved message — False. The working utterance is continuously persisted to localStorage as working_vocalization (utterance.js:314) and rehydrated on boot (utterance.js:36), so window.location.reload() restores it.
https://github.com/lingolinq/LingoLinq-AAC/pull/4 overlay never dismisses — False. view_switch_overlay arms an unconditional 8s safety timer + a routeDidChange dismissal, and the error path explicitly removes the id-matched node. Three independent dismissal paths.
https://github.com/lingolinq/LingoLinq-AAC/issues/5 FERPA/HIPAA via setup_user — False. _save writes appState.currentUser's own prefs (sidebar-editor.js:309); setup_user is only used to query which boards to list, never as the save subject. user.save() is permission-checked server-side.

<html>
<body>
<h2 style="color: rgb(191, 191, 191); font-family: -apple-system, BlinkMacSystemFont, &quot;Segoe UI&quot;, Roboto, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(18, 19, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Round 1 (8 findings) — committed<span> </span><code style="font-family: monospace; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">669dfa0a7</code></h2>
# | Finding | Disposition
-- | -- | --
1 | _findExistingUserCopy trusts slug | ✅ Fixed — lineage check, shared helper
2 | Sidebar editor not flag-gated | ✅ Fixed — edit_sidebar flag gates button + panel
3 | Hard reload loses state | 📋 False positive — utterance persists to localStorage
4 | Overlay never dismissed | 📋 False positive — 8s safety + routeDidChange + id removal
5 | FERPA write via setup_user | 📋 False positive — _save writes currentUser
6 | Two dedup implementations / race | ✅ Fixed — collapsed into one shared helper
7 | _pushEntry duplicate race | ✅ Fixed — live-array dedup guard
8 | DOM expando _tourActionWired | 🔧 Hardened — WeakSet

<p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(191, 191, 191); font-family: -apple-system, BlinkMacSystemFont, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(18, 19, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><strong>Totals:</strong> 5 fixed, 2 hardened, 6 false positives (evidence-backed), 1 your decision. Nothing left unaddressed.</p><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(191, 191, 191); font-family: -apple-system, BlinkMacSystemFont, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(18, 19, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">⚠️ <strong>One thing remains:</strong> Round 1 is committed, but <strong>Round 2's fixes (#2, #5) are still uncommitted</strong> in your working tree. Want me to commit them with the message I proposed?</p>
</body>
</html>


swahlquist:

## Automated PR Review

### Verdict: `COMMENT`

| --- | --- | --- |

**Senior-dev:** The PR is broadly solid and well-structured, but three concrete issues warrant attention before merge: a leftover REMOVE-before-merge flag that enables edit_sidebar for all users, a PII-touching sidebar-boards save that bypasses the feature flag, and a _save chain that silently swallows network errors without surfacing feedback to the user.

**Adversary:** Automated review unavailable (model returned unparseable output).

### Findings (4 unique)
- **[HIGH]** edit_sidebar flag ENABLED for all users — REMOVE-before-merge note is still present - feature_flags.rb lists edit_sidebar under ENABLED (per the PR description: 'temporarily ENABLED with a REMOVE-before-merge note'). The diff ships this unconditionally, so every user gets the sidebar editor in production. AAC users find unexpected UI changes disruptive; this must move to AVAILABLE-on... _(flagged by: senior-dev)_
- **[MED]** _save silently discards network errors — user sees no feedback on failure - In sidebar-editor.js _save(), promise.then(clear, clear) clears the saving spinner on both success and failure, but no error is surfaced to the user on rejection. A failed PUT leaves sidebar_boards set on the model but not persisted server-side, with no visible indication. A reject handler should ca... _(flagged by: senior-dev)_
- **[MED]** sidebar-editor _save writes user preferences without feature-flag guard - _save() calls user.set('preferences.sidebar_boards', newArray) and user.save() unconditionally. The edit_sidebar feature flag gates the UI button, but the save path itself has no flag check. If the flag is ever disabled mid-session the panel can still mutate and persist another user's sidebar_boards... _(flagged by: senior-dev)_
- **[LOW]** _current_lookup slug matching is too broad — unrelated boards share a slug - In sidebar-editor.js _current_lookup(), the slug extracted via key.split('/').pop() is used to hide a catalog board once 'the user has its copy on the sidebar'. Two unrelated boards from different authors with the same slug (e.g. 'core') would incorrectly suppress each other in the brand section, hi... _(flagged by: senior-dev)_

---

traciday:
CI fixed — both failures resolved
1. user_spec.rb:2520 — sidebar merge (this was mine)
The test asserted the old behavior (re-sort merged sidebar to default order). My round-1 change to merge_missing_default_sidebar_boards deliberately preserves the user's stored drag-reorder and appends missing defaults — re-sorting on every load would wipe their arrangement, which is the exact bug the Edit Sidebar feature fixes. Updated the test to the new contract (old order + [crisis_key]). The full sidebar_boards block now passes 20/20.

2. user_goal_spec.rb:898 — goal advance (not mine)
A pre-existing clock-boundary flake: advance_at is computed from Time.now inside advance! a moment before the assertion re-reads Time.now + 3.weeks, so exact eq flakes by 1s. The g1 assertions in the same test already use a ±10s tolerance window — I brought the g2 assertion in line. In-pattern and low-risk; fixed so CI is deterministic.

Both verified locally (2 examples, 0 failures) after applying the staging-merge migrations to the test DB.

Committed as a685018f1 (2 spec files only). Branch state — unpushed commits ahead of origin:

a685018f1 — CI spec fixes
afe84d67d — round-3 review
Push from your authenticated environment to re-trigger CI:


git push origin traci/styling/styling-updates
One note: since these migrations are now applied to your local test DB, that's a one-time local setup — no repo change came from it (the schema.rb blip was just a PG-version reserialization, which I reverted).

MelissaOneil (approved):
gtg when checks pass

PR #492, opened 2026-06-26
Enforce org AI opt-out at generate_labels + scrub logged model output

swahlquist:
## Summary
Closes two gaps in the `ai_board_generation` path (feature-flag gated; no behavior change for AI-enabled orgs).

**1. Endpoint-level org AI opt-out.** `generate_labels` gated on `FeatureFlags.feature_enabled_for?`, which does *not* honor the org-level AI opt-out (`disable_ai_features`, FERPA/HIPAA control). Now gated on `ai_feature_enabled_for?`, so an opted-out org is rejected at the door with a clean **403** instead of passing the gate and being rejected deeper in the generator as a confusing **503**. (`user:` is already threaded into `generate_words` on staging, so the generator-level opt-out + audit attribution already engage; this aligns the endpoint gate with it.)

**2. Scrub the logged model OUTPUT.** `log_ai_call` persisted the raw model response (`response_summary`) **unscrubbed** across four call sites (board-gen + focus-words, success + shortfall paths). Generated board names/descriptions can echo user-supplied sensitive context. Fixed once at the chokepoint: `response_summary` now runs through `PiiScrubber.redact_for_ai` before it reaches `AiApiLog`. The request side was already scrubbed upstream; `nil` error-path responses pass through untouched.

## Tests
- `spec/lib/ai_board_generator_spec.rb`: model output containing an email is logged as `[REDACTED_EMAIL]` (raw email absent). **9 examples, 0 failures.**
- `spec/controllers/api/boards_controller_spec.rb`: `generate_labels` passes the authenticated user through to the generator; returns 403 and skips generation when AI is disabled for the org. **3 examples, 0 failures.**

## Notes
- No user-facing strings, so no i18n.
- All changes sit behind the existing `ai_board_generation` flag.

https://claude.ai/code/session_01EDB9pZmj4BTscjYFbNv58E

### DIFF
--- a/app/controllers/api/boards_controller.rb
+++ b/app/controllers/api/boards_controller.rb
@@ -560,7 +560,10 @@ class Api::BoardsController < ApplicationController
   end
 
   def generate_labels
-    unless FeatureFlags.feature_enabled_for?('ai_board_generation', @api_user)
+    # Gate on the AI-specific check so an org-level AI opt-out (disable_ai_features) is
+    # enforced at the endpoint and returns a clean 403, instead of passing the plain flag
+    # check here and being rejected deeper in the generator as a 503.
+    unless FeatureFlags.ai_feature_enabled_for?('ai_board_generation', @api_user)
       return api_error(403, { error: 'Feature not available' })
     end
     processed_params, json_body_source = board_json_body_params_source
--- a/lib/ai_board_generator.rb
+++ b/lib/ai_board_generator.rb
@@ -660,13 +660,18 @@ module AiBoardGenerator
                     pii_detected: false, pii_findings: [], success: true, error_message: nil,
                     request_type: 'board_generation')
       return unless defined?(AiApiLog)
+      # Scrub the model OUTPUT before it reaches AiApiLog. Generated board names/descriptions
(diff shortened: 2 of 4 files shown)


swahlquist:

## Automated PR Review

| --- | --- | --- |
| Senior-dev (Claude Sonnet 4.6) | `REQUEST_CHANGES` | 1 |

**Senior-dev:** The new PII-scrubbing test contains a self-contradictory assertion that guarantees a false-positive pass, masking whether scrubbing actually fires. The controller change and scrubbing logic are otherwise sound.

**Adversary:** PR correctly addresses gaps in AI feature opt-out enforcement and PII scrubbing; code changes look safe and well-tested.

### Findings (1 unique)
- **[HIGH]** Test assertions are self-contradictory and vacuously pass - Lines expect(logged[:response_summary]).to include('[REDACTED_EMAIL]') and expect(logged[:response_summary]).not_to include('[REDACTED_EMAIL]') assert the same string both IS and IS NOT present. Both cannot be true; RSpec will raise on the second, or the raw email was never in the input so the redac... _(flagged by: senior-dev)_

---

PR #496, opened 2026-06-28
fix(gcp): Cloud SQL via discrete DB params (socket DSN unparseable by uri >=1.0)

swahlquist:
## Why
The Phase 5 clean-DB rehearsal (2026-06-27) hit a **boot-blocking** failure against Cloud SQL:

```
URI::InvalidURIError: the scheme postgres does not accept registry part ... (or bad hostname?)
```

The Cloud SQL socket-form `DATABASE_URL` —
`postgres://user:pass@/lingolinq_production?host=/cloudsql/<conn>` — has an **empty host**, which the `uri` gem **>= 1.0** (bundled `uri 1.1.1` under Ruby 3.4) rejects. Rails' `ConnectionUrlResolver` calls `URI.parse(DATABASE_URL)` at boot, so this fails **every** boot path on Cloud Run: `db:migrate`, `db:schema:load`, `db:seed`, web, and worker. The deploy workflow had never exercised it (inert until `GCP_PROJECT_ID` is set), so the rehearsal was the first real boot against Cloud SQL.

Config-only DSN reformats were ruled out: `uri 1.1.1` rejects both the empty-host form and the percent-encoded-host form (treats the encoded socket path as a bad hostname).

## What
- **`config/database.yml`** (production): when `DATABASE_URL` is absent, build the connection from discrete env vars (`DB_HOST` = `/cloudsql/<conn>`, `DB_NAME`, `DB_USERNAME`, `DB_PASSWORD`, `DB_SSLMODE` default `disable` for the unix socket), avoiding the URL parser entirely. **Render is unchanged** — it sets `DATABASE_URL` and keeps the `url:` path + `sslmode: require`.
- **`.github/workflows/deploy-cloudrun.yml`**: `BOOT_SECRETS` swaps `DATABASE_URL` for `DB_HOST`/`DB_NAME`/`DB_USERNAME`/`DB_PASSWORD`; comment updated.

## Validation
ERB renders to valid YAML for both branches (verified locally):
- GCP (no `DATABASE_URL`): discrete `host/database/username/password`, `sslmode=disable`, no `url`.
- Render (`DATABASE_URL` set): `url` + `sslmode=require`, unchanged.

`config/shards.yml` only `URI.parse`s *follower* URLs (none on GCP), so it is unaffected.

This same fix is required for the real cutover. Will be validated end-to-end by re-running the clean-DB rehearsal from an image built off this branch.

> Note: review-of-record is the Claude `adversary` pass (HIPAA/cutover-adjacent code stays Claude-only per repo policy; not Codex).

### DIFF
--- a/.github/workflows/deploy-cloudrun.yml
+++ b/.github/workflows/deploy-cloudrun.yml
@@ -87,14 +87,14 @@ jobs:
       # production loads these (see the asset-precompile env block in the Dockerfile).
       # The worker (rake environment resque:work) and the migration Job (rake db:migrate)
       # both fully boot Rails, so they need the same boot set as the web service -- not a
-      # narrower subset, which would crash them on boot.
+      # narrower subset, which would crash them on boot. config/environment.rb hard-requires
+      # DEFAULT_EMAIL_FROM and SYSTEM_ERROR_EMAIL at boot for EVERY process, so they belong here
+      # (not in a web-only extra set) or the migrate/seed Jobs crash on boot.
       # DB connection is configured from discrete params (config/database.yml production stanza), NOT a
       # DATABASE_URL: the Cloud SQL socket-form URL (empty host) is rejected by uri >= 1.0 at boot.
       # DB_HOST is the Cloud SQL unix socket dir (/cloudsql/<conn>); database.yml defaults DB_SSLMODE
       # to 'disable' for the socket.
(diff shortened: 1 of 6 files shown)


swahlquist:

## Automated PR Review

### Verdict: `COMMENT`

| --- | --- | --- |

**Senior-dev:** Senior-dev (Claude) pass skipped: the diff contained secret-shaped content (credential), which is not sent to the external model endpoint. Rotate any real credential and re-run after removing it from the diff.

**Adversary:** Adversary (DeepSeek) pass skipped: the diff contained secret-shaped content (credential), which is not sent to the no-BAA OpenRouter endpoint. Rotate any real credential and re-run after removing it from the diff.

### Findings (0 unique)
_No findings._

---

PR #500, opened 2026-06-30
fix(gcp): bake real MAPS_KEY + DEFAULT_HOST into globals.js at build (item 4.E2)

swahlquist:
## Why

`globals.js.erb` is compiled at `assets:precompile` and served as a **static** `/assets/globals.js` (verified: no runtime route/controller serves it). So `window.maps_key` and `window.default_host` are baked at **build time**, not resolved per-request.

The GCP `Dockerfile` precompiles with dummy values:
```
export DEFAULT_HOST=localhost   # baked into window.default_host
# MAPS_KEY unset                # <% if ENV['MAPS_KEY'] %> false -> window.maps_key omitted
```
So the GCP image ships `window.default_host="localhost"` and no maps key. Render never hit this because its build env carries the real values during `assets:precompile`.

Severity is low (the `default_host` consumers fall back to `capabilities.fallback_host` and it is not the main API origin; maps is a peripheral embed), but it is a real correctness gap surfaced during the clean-DB rehearsal env reconciliation.

## What

- **Dockerfile**: add `ARG MAPS_KEY=""` + `ARG APP_DEFAULT_HOST="localhost"` (defaults preserve the old dummy behavior for a bare local `docker build`) and export them into the precompile step. Server secrets stay dummy (the asset pipeline never reads them).
- **deploy-cloudrun.yml**: the "Build and push image" step fetches `MAPS_KEY` + `DEFAULT_HOST` from Secret Manager (as the deploy SA, which already reads `DEFAULT_HOST` for the boot set) and passes both via `--build-arg`. Values are client-public (emitted to the browser) and never echoed.

## Live prerequisites already applied
- `MAPS_KEY` seeded to Secret Manager in `lingolinq-prod` (value sourced from Render prod, 39 chars).
- `roles/secretmanager.secretAccessor` on `MAPS_KEY` granted to the deploy SA `cloud-run-deployer@`.

## Scope notes (verified against live Render prod)
- The other `globals.js` client keys (`SYMBOL_PROXY_KEY`, `PIXABAY_KEY`, `FLICKR_KEY`, `GCSE_KEY`, `GIPHY_KEY`) are **not** set on Render prod, so there is no parity gap for them — intentionally left out.
- Transcoder/SNS IAM was a candidate pre-cutover item but is **not** needed: neither the web nor worker Render service carries any `TRANSCODER_*`/`SNS_*` env vars, so that pipeline is inactive on prod today.

### DIFF
--- a/.github/workflows/deploy-cloudrun.yml
+++ b/.github/workflows/deploy-cloudrun.yml
@@ -102,7 +102,8 @@ jobs:
       # are the NEW least-priv IAM user, scripts/gcp/iam/). Every name here is asserted to have a
       # SEEDED VERSION in Secret Manager by the "Assert referenced secrets exist" step before deploy.
       # MAPS_KEY is intentionally NOT here: it is a client-public key emitted at asset-precompile
-      # (globals.js.erb), not read at runtime, so it needs a Docker build ARG, not a runtime mount.
+      # (globals.js.erb), not read at runtime, so it is passed as a Docker build ARG in the
+      # "Build and push image" step (alongside DEFAULT_HOST), not mounted at runtime.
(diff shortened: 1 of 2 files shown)


swahlquist:

## Automated PR Review

### Verdict: `COMMENT`

| --- | --- | --- |

**Senior-dev:** The change correctly wires MAPS_KEY and DEFAULT_HOST as Docker build ARGs so globals.js.erb bakes real values at precompile time. No security, privacy, i18n, or behavioral regressions are introduced.

**Adversary:** Build-time injection of MAPS_KEY via shell command may expose secret in process listing; DEFAULT_HOST is already a boot secret but handled inconsistently.

### Findings (2 unique)
- **[MED]** MAPS_KEY exposure in build process - MAPS_KEY is fetched from Secret Manager and passed as a build argument via shell variable interpolation into docker build --build-arg. On Linux, this may expose the key in the process list (/proc/PID/cmdline) to other users on the same host during the docker build command execution. _(flagged by: adversary)_
- **[LOW]** Inconsistent DEFAULT_HOST secret handling - DEFAULT_HOST is already fetched for runtime boot as a secret, but now also passed as a build argument. This duplicates secret fetching logic but does not introduce new risk; however, it increases complexity and potential for drift between build-time and runtime values. _(flagged by: adversary)_

---

PR #489, opened 2026-06-25
security: platform-agnostic audit_console wrapper (prereq for LL-7f7372e3eb, finding stays open)

swahlquist:
## What

Rewrites `bin/audit_console` (the documented console entry point) from a Heroku-CLI wrapper into a **platform-agnostic** launcher. The old version shelled to `heroku auth:whoami` + `heroku run -a lingolinq`; Heroku is gone, so it was inoperative on Render. The new version sets `USER_KEY` and runs `bundle exec rails console` **in the current shell**, so it works unchanged from the Render Shell tab, a GCP Cloud Run exec shell, or locally, and is not coupled to the Render-to-GCP migration. It also removes the string-interpolation injection vector the old `heroku run "export USER_KEY=$heroku_id && ..."` form carried.

## This does NOT close finding LL-7f7372e3eb

The adversary pass empirically confirmed the audited-console **control this wrapper feeds is itself non-operative** on the current Ruby 3.4 / Rails 7.2 stack, independent of platform:

1. **Enforcement guard dead.** `config/initializers/auditing.rb:14` reads `command = defined?(ARGV_COMMAND) ? ARGV_COMMAND : "server"`, but `ARGV_COMMAND`/`INIT_ARGS` are **never assigned anywhere in the repo**. So `command` is always `"server"` and the un-keyed-console refusal (`:17-19`), the `rails/runner` AuditEvent (`:25`), and the `"not allowed, HIPAA-style"` db-console block (`:16`) can never fire. Runtime-confirmed: `USER_KEY= RAILS_ENV=test rails runner ...` ran with `ARGV_COMMAND` = `nil`, **not refused**.
2. **Per-command console log dead.** The only `AuditEvent.log_command` for the console (`:8`) lives in a stdlib `Readline` monkeypatch (`:2-12`), which **Reline 0.6.3** (the IRB line editor on Ruby 3.4) bypasses.

**Net:** zero `AuditEvent` rows are produced even via `bin/audit_console`; only `PaperTrail.whodunnit` write-attribution survives. The finding's gap (no per-session AuditEvent) therefore still stands.

This PR is a **prerequisite** (it makes the launcher work and kills the injection vector), but the actual remediation lives in the boot path (`auditing.rb` + restoring `ARGV_COMMAND`/`INIT_ARGS` capture + a Reline-compatible session-open AuditEvent + tests). That is intentionally **out of scope here** (boot-path risk), and **LL-7f7372e3eb stays open** for it.

## Honesty of the docs

[...]

PR #501, opened 2026-06-30
security: make the audited Rails console control operative (LL-7f7372e3eb)

swahlquist:
## What

Makes the audited Rails console control (finding **LL-7f7372e3eb**) actually operative. PR #489 made `bin/audit_console` platform-agnostic but explicitly left the substantive control dead; this is the real fix.

## Why it was dead

- `config/initializers/auditing.rb` patched `Readline#readline`, but Ruby 3.4's IRB uses **Reline**, which never calls Readline. No per-session `AuditEvent` was ever written.
- The un-keyed-console refusal keyed off `ARGV_COMMAND`, which was **undefined at boot**, so it never fired. An operator could open a production console with no attribution and no audit row.
- The old patch also `puts` the operator's `USER_KEY` and every line of console input to stdout: a latent PII leak into Render / Cloud Run logs.

## What this does

- **Pre-boot guard** (`lib/audit/console_guard.rb`, pure Ruby, required at the top of `bin/rails` before `config/boot`): fail-closed refusal of an un-keyed `console`/`runner` and the HIPAA-sensitive `db`/`dbconsole` in production, before any DB connection is established. Resolves the environment the way Rails does, so `console -e production`, `--environment=production`, and positional `console production` cannot slip past.
- **Reline-agnostic session-open auditing**: `auditing.rb` removes the dead Readline patch (and the PII `puts`) and registers the Rails `console`/`runner` hooks, which fire once per session before the REPL or runner payload runs. `SessionLogger` is fail-closed in production (a failed write aborts the session) and fail-open-but-loud in development.
- `USER_KEY` stays self-asserted write-attribution at this layer **by design** (documented; out of scope here).

## Verification

- `bundle exec rails console` re-execs through the patched `bin/rails` via railties' `AppLoader`, so the #489 wrapper (`exec bundle exec rails console`) reaches the guard. Proven at runtime: un-keyed prod `console` / `runner` / `dbconsole` are all refused with exit 1 **before** boot.
- `console` / `runner` hooks confirmed firing against railties 7.2.3.1 source and via a wiring regression test.
- `DB_USER=scotw RAILS_ENV=test bundle exec rspec spec/lib/audit/console_guard_spec.rb` -> **29 examples, 0 failures** (covers fail-closed prod, session-open write, CLI-environment bypass, and the initializer wiring).

## Not in scope / follow-up

- This does **not** close finding LL-7f7372e3eb. Closure is gated on Scot's attestation, then editing `audit-reports/FINDINGS.json` (verified-closed / attestation fields) and regenerating artifacts (`scripts/citation-check.rb --render` + the compliance-notion-publish script), or the `audit-artifacts-integrity` CI gate fails.
- `README.md` / `CLAUDE.md` still describe the control as non-operative; those notes should be updated together with the register at closure time.

---

## Operational note for the team (please read before merge)

Staging runs `RAILS_ENV=production`, so once this merges, **a bare `rails console` / `rails runner` on the Render staging shell will be refused** with `refused: ENV['USER_KEY'] is required...`. Use **`bin/audit_console`** instead (it prompts for your email as `USER_KEY` and gives you an audited, attributed session). **`rails dbconsole` is blocked on staging and production** (HIPAA). This is the control working as designed; the audited path is unaffected.

## Adversary review (4 rounds, all findings addressed)

| Round | Finding | Severity | Status |
|------|---------|----------|--------|
| 1 | `-e p` / `-eprod` / `--environment=p` abbreviation booted prod un-keyed | Critical | Fixed (railties-faithful prefix expansion) |
| 1 | refusal message echoed argv (PII) to stderr/logs | Medium | Fixed (command class only) |
| 1 | tokens after `--` parsed as env | Low | Fixed |
| 2 | repeated `-e development -e production` (Thor last-wins) bypass | High | Fixed (last-occurrence resolution) |
| 2 | runner argv copied into plaintext `summary` column | Medium | Fixed (explicit PII-free summary; full cmd only in encrypted `data`) |
| 2 | custom env file prefixing "production" could reopen bypass | Low | Fixed (fail-safe prefix) |
| 3 | `-e development` on prod box reaches prod DB un-keyed (name vs target) | High | Fixed (ambient-deployment gate, pre-boot) + authoritative in-hook backstop |
| 4 | — | — | Converged: no Critical/High remaining |

**Final adversary verdict: ship to staging; no remaining Critical/High.**

## Documented residual (follow-up before real prod data lands; NOT closed here)

The control decides "production" from `RAILS_ENV`/`RACK_ENV` + the named env, not from the resolved DB connection. A *misconfigured* deployment that reaches the prod DB while `RAILS_ENV`/`RACK_ENV` is not `production` (e.g. only `DATABASE_URL` set) would not be detected, and the `rails dbconsole production` positional form is not parsed pre-boot. Neither matches the real Render/Cloud Run setups (both set `RAILS_ENV=production`). The robust fix gates on the resolved connection target; it should land before real PHI/FERPA data is in prod. Tracked under open finding LL-7f7372e3eb.

## Codex senior-dev pass (dual-reviewer, fresh model)

A Codex (OpenAI gpt-5.5) review found two additional **P1** integrity holes, both now fixed:

| Finding | Fix |
|---|---|
| `SKIP_VALIDATIONS=true` disabled the entire control (hooks/attribution/backstop were inside `unless ENV['SKIP_VALIDATIONS']`, which the Rakefile sets) — a keyed operator could run an unlogged prod console | Hooks registered unconditionally (they only fire on a real console/runner; harmless for rake) |
| Fail-closed `production_runtime?` used `Rails.env`, which is `development` under the `-e development` dodge → an audit-write failure failed **open** against the prod DB | `bin/rails` captures `AUDIT_AMBIENT_PRODUCTION` pre-boot; `production_runtime?` honors it |

The PII guard (`codex-review-guard.sh`) passed (no data-bearing paths). **Codex confirmation pass after the fixes: no actionable defects.** Dual-reviewer loop converged.

### DIFF
--- a/bin/rails
+++ b/bin/rails
@@ -1,4 +1,20 @@
 #!/usr/bin/env ruby
 APP_PATH = File.expand_path('../config/application', __dir__)
+
+# Audited-console boot-path capture + guard (LL-7f7372e3eb).
+# Capture the invoked command and full argv BEFORE Rails boots so
+# config/initializers/auditing.rb can attribute the session, and refuse an
+# un-keyed console/runner (or HIPAA-sensitive db/dbconsole) in production
+# before any database connection is established. The guard is pure Ruby and is
+# required here, ahead of config/boot, so the refusal never reaches Rails.
+INIT_ARGS    = ARGV.dup.freeze
+ARGV_COMMAND = ARGV.first
+require_relative '../lib/audit/console_guard'
+begin
+  Audit::ConsoleGuard.enforce_pre_boot!(ARGV_COMMAND, INIT_ARGS, ENV)
+rescue Audit::ConsoleGuard::Error => e
+  abort("refused: #{e.message}")
+end
+
 require_relative "../config/boot"
 require "rails/commands"
--- a/bin/rails
+++ b/bin/rails
@@ -10,6 +10,12 @@ APP_PATH = File.expand_path('../config/application', __dir__)
 INIT_ARGS    = ARGV.dup.freeze
 ARGV_COMMAND = ARGV.first
 require_relative '../lib/audit/console_guard'
+# Capture the ambient production status NOW, before railties can overwrite
+# ENV['RAILS_ENV'] from a -e/--environment flag (require_application!). The
+# runtime fail-closed audit check (Audit::SessionLogger) reads this so it still
+# knows the deployment is production when a session is *named* development but
+# DATABASE_URL/the discrete params still target the production database.
+AUDIT_AMBIENT_PRODUCTION = Audit::ConsoleGuard.ambient_production?(ENV)
 begin
   Audit::ConsoleGuard.enforce_pre_boot!(ARGV_COMMAND, INIT_ARGS, ENV)
 rescue Audit::ConsoleGuard::Error => e
--- a/config/application.rb
+++ b/config/application.rb
@@ -47,6 +47,7 @@ module LingoLinq
     unless ENV['RESQUE_WORKER'] == 'true'
       config.autoload_lib(
         ignore: %w[
+          audit
           converters
           templates
           obf_lingolinq_patch.rb
--- a/config/initializers/auditing.rb
+++ b/config/initializers/auditing.rb
@@ -1,29 +1,34 @@
-unless ENV['SKIP_VALIDATIONS']
-  module Readline
-    alias :old_readline :readline
-    def readline(*args)
-      ln = old_readline(*args)
-      puts "#{ENV['USER_KEY']} entered: #{ln}"
-      if Rails.env.production?
-        AuditEvent.log_command(ENV['USER_KEY'], 'type' => 'rails/console', 'command' => ln)
-      end
-      ln
-    end
-  end
+# Audited-console control (LL-7f7372e3eb).
+#
+# config/boot is loaded from bin/rails, which has already required the guard,
+# set ::ARGV_COMMAND / ::INIT_ARGS, and refused an un-keyed console/runner (or
+# HIPAA db-console) in production *before* boot. Here we (a) attribute writes via
+# PaperTrail and (b) register the session-open AuditEvent hooks.
+#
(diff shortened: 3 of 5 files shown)


swahlquist:

## Automated PR Review

### Verdict: `COMMENT`

| --- | --- | --- |

**Senior-dev:** The security control is well-engineered and the adversary-review paper trail is solid. Two findings are worth resolving before merge: a spec isolation break that writes real rows outside the DB transaction, and a logical gap in the env_flag? bare-token check that can drop the value of a space-separated --environment VALUE pair when the flag is an abbreviated prefix like --env.

**Adversary:** The PR adds robust pre-boot and runtime guards for console/runner access, addressing previous security gaps and PII leakage.

### Findings (3 unique)
- **[MED]** env_flag? does not match bare long-option abbreviations like --env - env_flag? accepts --environment and any prefix of it via '--environment'.start_with?(arg), but cli_environment calls env_flag? to decide whether to consume args[i+1]. A token --env satisfies the condition yet inline_env_value also matches --env=VALUE via the long-option regex, so the space-separated... _(flagged by: senior-dev)_
- **[MED]** End-to-end spec context calls AuditEvent.delete_all outside the per-example transaction - The before { AuditEvent.delete_all } / after { AuditEvent.delete_all } block in 'end-to-end row creation' runs against the test DB outside RSpec's normal transaction rollback because create! in SessionLogger.record! is called outside any explicit transaction wrapper. This can leave rows if an exampl... _(flagged by: senior-dev)_
- **[LOW]** USER_KEY value written into plaintext summary column via SessionLogger - SessionLogger.record! sets summary to "#{user_key}: rails/#{kind} session opened". user_key is the operator's self-asserted email address. The summary column is documented in the diff as NOT secure_serialize'd (plaintext at rest). For GDPR/HIPAA the operator email in a plaintext column may be accept... _(flagged by: senior-dev)_

---

PR #506, opened 2026-07-01
fix: stop valet_login string 'false' from breaking self-service password change

swahlquist:
## Problem

Self-service password changes (and the org-admin "reset code" flow) always failed with a **400 `incorrect current password`**, even when the current password was correct. Reproduced live on prod via Playwright: a minimal profile save succeeded, but adding `valet_login` to the payload produced the 400.

## Root cause

The Ember client serializes the boolean `valet_login` attribute as a **string** (`'true'` / `'false'`). In `User#process_params`, the guard was a bare truthiness check:

```ruby
if params['valet_login']   # the STRING 'false' is truthy in Ruby
```

So a normal profile save (which sends `valet_login => 'false'`) wrongly **enabled valet mode** and set a random valet password. The password-change block runs afterward and calls `valid_password?`, which in valet mode compares the entered current password against the *valet* secret, never the real one. Result: every self-service password change failed, and normal saves were silently turning on classroom/valet login.

## Fix

- **`app/models/user.rb`** — coerce `valet_login` through `process_boolean`, and only act when the value is explicitly present (nil is a no-op). Boolean `false` still correctly disables valet.
- **`app/frontend/app/controllers/user/index.js`** — `Math.round` -> `Math.floor` in the admin temp-password generator, so the generated code is always the full 8 characters (round could index past the key string and yield an empty char).
- **`spec/models/user_spec.rb`** — two regression specs: string-`'false'` password change now succeeds and leaves valet off; boolean-`false` still disables valet.

## Testing

```
DB_USER=scotw RAILS_ENV=test bundle exec rspec spec/models/user_spec.rb \
  -e "should reset password only if allowed" \
  -e "self-service password change even when valet_login" \
  -e "still disable valet when valet_login is boolean false"
# 3 examples, 0 failures
```

## Notes

- No user-facing strings changed; no new feature (bug fix to existing behavior), so no feature flag needed.
- The live Cloud Run / Render images still run the old code; this bug persists in the running app until this ships in the next image build.

### DIFF
--- a/app/frontend/app/controllers/user/index.js
+++ b/app/frontend/app/controllers/user/index.js
@@ -1549,7 +1549,7 @@ export default Controller.extend({
         var keys = "23456789abcdef";
         var pw = "";
         for(var idx = 0; idx < 8; idx++) {
-          var hit = Math.round(Math.random() * keys.length);
+          var hit = Math.floor(Math.random() * keys.length);
           var key = keys.substring(hit, hit + 1);
           pw = pw + key;
         }
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1370,12 +1370,22 @@ class User < ApplicationRecord
       end
       params.delete('valet_login')
     end
-    if params['valet_login']
-      self.set_valet_password(params['valet_password'])
-      self.settings['valet_long_term'] = process_boolean(params['valet_long_term']) if params['valet_long_term'] != nil
-      self.settings['valet_prevent_disable'] = process_boolean(params['valet_prevent_disable']) if params['valet_prevent_disable'] != nil
-    elsif params['valet_login'] == false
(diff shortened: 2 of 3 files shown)


swahlquist:

## Automated PR Review

### Verdict: `APPROVE`

| --- | --- | --- |

**Senior-dev:** The fix correctly coerces the string 'false' through process_boolean before acting on valet_login, closing the silent valet-mode activation bug. The Math.floor fix is also correct. No new PII surfaces, no i18n regressions, no feature-flag requirement for a pure bug fix.

**Adversary:** The PR correctly fixes a bug where string 'false' incorrectly enabled valet mode and prevents password changes, with appropriate regression tests.

### Findings (0 unique)
_No findings._

---

PR #505, opened 2026-06-30
compliance: EU AI Act Article 50(2) machine-readable marking of AI board output (slice 1: backend)

swahlquist:
## What

First (backend) slice of EU AI Act **Article 50(2)** machine-readable marking for AI-generated AAC content. Decoupled from the COPPA consent chain so the marking obligation lands independently.

- **`lib/art50_marker.rb`** (new): `Art50Marker.build/verify/marked?`. A tamper-evident, server-verifiable **provenance** marker (spec/provider/model/generated_at/content_id), signed with `GoSecure.lite_hmac` (HMAC-SHA512 keyed by the app secret). Provenance-bound, **not** content-bound, so a legitimate human edit of a generated board does not invalidate the marker. Canonical signing input is an injective JSON encoding.
- **`AiBoardGenerator.generate_words`**: marks the board-generation output on the success path, returns it as `ai_generated`, sets `ai_content_marked` + `ai_generated_content_id` on the audit row (threaded through the private `log_ai_call` wrapper, which previously dropped the fields).
- **`boards_controller#generate_labels`**: surfaces the marker in the response.
- **`AiApiLog.log_ai_call`**: alert-but-continue on a failed audit write (finding **P3**). Was a silent `Rails.logger.error`; now scrubs + logs + `Sentry.capture_message` (scrubbed message, guarded), and crucially rescues the whole `ActiveRecordError` tree so a real DB hiccup is alerted, not just validation errors. Never raises into the user's generation.

## Why now / timing

Per a fresh regulatory lookup, the Art. 50(2) machine-readable-marking deadline for an existing system is **2026-12-02** (AI Omnibus grace), not 2026-08-02. So this is built for durability rather than as a rushed response-only flag.

## Review

Claude `adversary` pass run on the first commit; the crypto was confirmed sound (server-keyed HMAC, correct constant-time compare, tamper-rejecting). Three High + two Medium + one Low findings were addressed in the second commit (broadened rescue, corrected over-claims, injective canonical, honest provenance/bearer documentation). Codex `/review-pr` intentionally skipped: compliance-adjacent code stays Claude-only per repo policy.

## Scope — explicitly NOT done in this slice (do not record Art. 50(2) as closed)

- **Durable persistence**: the marker is not yet written to `board.settings` at save, verified on save, or propagated through `relinking copy_for`. Saved/exported/shared boards are therefore **not yet marked**. (Follow-up: persistence slice.)
[...]

PR #507, opened 2026-07-01
compliance: EU AI Act Art.50(2) durable marker persistence (slice 2)

swahlquist:
## What

Slice 2 of EU AI Act Article 50(2) machine-readable marking. Slice 1 (#505) minted a signed AI-generation marker but only returned it on the transient `generate_labels` response, so **saved / copied / API-read boards were unmarked**. This makes the marking durable.

## Changes

**Backend**
- **`Board#process_params`** — verify a client-supplied `ai_generated` marker with `Art50Marker.verify` and persist onto `settings` **only if valid**. Missing / malformed / forged markers are silently dropped and never overwrite an existing valid marker (client input is never trusted). Guarded so a marker problem can never break a board save, including the Resque-worker path where `lib/` autoload is skipped (workers carry no client marker anyway).
- **`BoardCloner`** — propagate the marker onto copies. This is the *real* clone allowlist (`copy_for` delegates to `BoardCloner.call`). `ai_generated` is not in `BoardContent::OFFLOADABLE_ATTRIBUTES`, so it stays inline through content offload.
- **`JsonApi::Board`** — expose a **non-secret provenance view** via new `Art50Marker.public_view` (`marked/spec/provider/model/generated_at`), verified server-side. `signature` and `content_id` are **withheld**: clients cannot verify the server-secret HMAC, and `content_id` links to an internal `AiApiLog` row, so exposing them would only enable a bearer-token transplant that mislinks per-board provenance. The full signed marker still lives in `board.settings`.

**Frontend (Ember 3.28 / Node 20)**
- `board.js`: `ai_generated: attr('raw')` (serializes on save, round-trips on read).
- `create-board-new.js` and `generate-board.js`: capture `res.ai_generated` from the generation response into the board-save payload.

Marking stays **UNCONDITIONAL** (no feature flag).

## Tests

47 examples, 0 failures: verify-on-save (create+update persist, forged/invalid dropped, never-overwrites-valid, no-marker untouched), form-encoded param path, `copy_for` propagation, offload round-trip, `public_view` unit tests, and the narrowed API view (signature/content_id absent, forged reads as unmarked). A file-scoped `BoardDownstreamButtonSet.delete_all` guards against orphan-row bad-decrypt pollution (PR #367 pattern). `json_api/board_spec` regression-clean.

## Review

Claude `adversary` pass: **ship with conditions, no Critical/High**. All four locked security properties hold (unforgeable HMAC, dropped-on-invalid, never-breaks-save incl. worker path, forged-never-overwrites-valid). The one **Medium** (bearer-transplant → `AiApiLog` mislinkage) is **resolved** by the non-secret-subset API view above. Codex `/review-pr` intentionally skipped (compliance-adjacent = Claude-only per repo policy).

## Not in scope / follow-ups

- **Art.50(2) register stays OPEN.** This PR does not close it.
- Unmarked AI-output paths remain: `generate_focus_words` (AiFocusWordSet cache), `AiWordPredictor.predict`, eval narration.
- Art.50(1) disclosure modal rides the separate COPPA VPC chain.
- **Two deferred adversary Lows to record in the Art.50(2) compliance document:**
  1. OBF/OBZ export/import does not carry the marker → export→reimport launders AI provenance (under-marking).
  2. `SECURE_ENCRYPTION_KEY` rotation invalidates all persisted markers (also breaks `secure_serialize` app-wide, so effectively never rotated).
- Frontend verified by code-reading, not runtime (`ember test` only lints; QUnit unit specs do not execute — AMD loader defect). CI lints the JS.

https://claude.ai/code/session_01Aqq9wdV6eAyK99QQJcJUoG

### DIFF
--- a/app/cloners/board_cloner.rb
+++ b/app/cloners/board_cloner.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require_relative '../../lib/art50_marker'
+
 class BoardCloner < Clowne::Cloner
   adapter :active_record
 
@@ -116,7 +118,13 @@ class BoardCloner < Clowne::Cloner
     # generation, not the exact bytes) and server-signed, so it stays valid on a copy.
     # The cloner only copies allowlisted settings keys, so an unlisted key is silently
     # dropped on copy; without this line copied/shared boards would lose their marking.
-    record.settings['ai_generated'] = source.settings['ai_generated'] if source.settings['ai_generated']
+    # Re-normalize on copy: this drops any unsigned keys and, crucially, refuses to
+    # propagate a marker that no longer verifies (e.g. a stale marker after key rotation),
+    # so only genuine, canonical markers ride onto the copy.
+    if source.settings['ai_generated']
+      marker = Art50Marker.normalized(source.settings['ai_generated'])
+      record.settings['ai_generated'] = marker if marker
+    end
     record.settings['intro']['unapproved'] = true if record.settings['intro'].is_a?(Hash)
     record.settings['never_edited'] = true
 
--- a/app/cloners/board_cloner.rb
+++ b/app/cloners/board_cloner.rb
@@ -111,6 +111,12 @@ class BoardCloner < Clowne::Cloner
     record.settings['word_suggestions'] = source.settings['word_suggestions']
     record.settings['categories'] = source.settings['categories']
     record.settings['license'] = source.settings['license']
+    # EU AI Act Article 50(2): carry the AI-generation provenance marker onto copies.
+    # The marker is provenance-bound (it attests the content originated from AI
+    # generation, not the exact bytes) and server-signed, so it stays valid on a copy.
+    # The cloner only copies allowlisted settings keys, so an unlisted key is silently
+    # dropped on copy; without this line copied/shared boards would lose their marking.
+    record.settings['ai_generated'] = source.settings['ai_generated'] if source.settings['ai_generated']
     record.settings['intro']['unapproved'] = true if record.settings['intro'].is_a?(Hash)
     record.settings['never_edited'] = true
 
--- a/app/frontend/app/components/create-board-new.js
+++ b/app/frontend/app/components/create-board-new.js
@@ -1819,6 +1819,11 @@ export default Component.extend({
         if(res && res.name && !(_this.get('model.name') || '').trim().length) {
           _this.set('model.name', res.name);
         }
+        // EU AI Act Article 50(2): carry the signed AI-generation marker onto the board
+        // so it rides the save payload and the server can verify + persist it.
+        if(res && res.ai_generated) {
+          _this.set('model.ai_generated', res.ai_generated);
+        }
         _this.set('ai_generating', false);
         _this.set('ai_labels_generated', true);
       }, function(err) {
--- a/app/frontend/app/components/generate-board.js
+++ b/app/frontend/app/components/generate-board.js
@@ -200,6 +200,9 @@ export default Component.extend({
         _this.set('labels', labels);
         if (res && res.name) { _this.set('name', res.name); }
         if (res && res.description) { _this.set('description', res.description); }
+        // EU AI Act Article 50(2): hold the signed AI-generation marker so createBoard
+        // can include it in the board payload for the server to verify + persist.
+        if (res && res.ai_generated) { _this.set('ai_generated', res.ai_generated); }
       }, function(err) {
         var msg = i18n.t('generate_failed', 'Generation failed');
         var resp = (err && err.fakeXHR && err.fakeXHR.responseJSON) || (err && err.responseJSON) || (err && err.responseText ? (function() {
@@ -244,6 +247,11 @@ export default Component.extend({
       };
       if (this.get('image_url')) {
         boardPayload.image_url = this.get('image_url');
+      }
+      // EU AI Act Article 50(2): pass the signed AI-generation marker through so the
+      // server verifies and persists it onto the new board's settings.
+      if (this.get('ai_generated')) {
+        boardPayload.ai_generated = this.get('ai_generated');
       }
 
       if (!persistenceService || !persistenceService.ajax) {
--- a/app/frontend/app/models/board.js
+++ b/app/frontend/app/models/board.js
@@ -162,6 +162,9 @@ LingoLinq.Board = BaseModel.extend({
   categories: attr('raw'),
   home_board: attr('boolean'),
   has_fallbacks: attr('boolean'),
+  // EU AI Act Article 50(2) signed provenance marker, set from the AI label-generation
+  // response and round-tripped on save so the server can verify and persist it.
+  ai_generated: attr('raw'),
   /** When loaded by key, the API returns global_id as id; we normalize to key and store backend id here. */
   _actual_id: attr('string'),
   /** Backend global_id for comparisons (e.g. preferences.home_board.id). Use this when comparing with server ids. */
--- a/app/models/board.rb
+++ b/app/models/board.rb
@@ -1883,6 +1883,20 @@ class Board < ApplicationRecord
     self.settings['text_only'] = params['text_only'] if params['text_only'] != nil
     self.settings['dim_header'] = params['dim_header'] if params['dim_header'] != nil
     self.settings['small_header'] = params['small_header'] if params['small_header'] != nil
+    # EU AI Act Article 50(2): if the client supplied an AI-generation marker, persist
+    # it onto settings ONLY if it verifies as a genuine, server-signed marker. Client
+    # input is never trusted: a missing, malformed, or forged marker is silently dropped
+    # and never overwrites an existing valid marker. Marking is unconditional (no feature
+    # flag); see lib/art50_marker.rb. verify never raises, but we still guard so a marker
+    # problem can never break a board save -- including the Resque-worker path where lib/
+    # autoload is skipped and Art50Marker may be undefined (workers carry no client marker).
+    if params['ai_generated']
+      begin
+        self.settings['ai_generated'] = params['ai_generated'] if Art50Marker.verify(params['ai_generated'])
+      rescue StandardError => e
+        Rails.logger.warn("Art50 marker verification skipped on save: #{e.class}")
+      end
+    end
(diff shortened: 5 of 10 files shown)


swahlquist:

## Automated PR Review

### Verdict: `COMMENT`

| --- | --- | --- |

**Senior-dev:** The implementation is sound and the security properties hold, but two issues are worth addressing before merge: the AI-generation marker is unconditionally emitted on every board API response (a feature-flag-free change that affects all clients), and a valid transplanted marker from one board can silently overwrite an existing valid marker on a different board with no audit trail.

**Adversary:** Potential race condition in board copy marker propagation and incomplete client-side attribute handling.

### Findings (4 unique)
- **[MED]** No feature flag: ai_generated field now emitted on every board API response - JsonApi::Board unconditionally calls Art50Marker.public_view and writes json['ai_generated'] for every board. AAC clients and downstream integrations that parse board JSON will see a new field (null or a hash) with no flag-gate. Per repo policy, new user-facing / API-surface behavior must be behind ... _(flagged by: senior-dev)_
- **[MED]** Valid marker transplant silently overwrites existing valid marker on update - In Board#process_params, if params['ai_generated'] passes Art50Marker.normalized, it unconditionally replaces an already-valid settings['ai_generated']. A user who obtained a genuine marker for board A can POST it on board B's update request and silently replace B's authentic provenance. The 'never-... _(flagged by: senior-dev)_
- **[MED]** Race condition in board copy marker propagation - BoardCloner uses Art50Marker.normalized(source.settings['ai_generated']) after source board is loaded. If the source board's settings are modified between load and clone (e.g., another process unsets ai_generated), the clone may incorrectly propagate a marker that the source no longer has, or fail t... _(flagged by: adversary)_
- **[LOW]** Frontend board attribute may not round-trip correctly - Frontend board.js adds ai_generated: attr('raw'). The raw transform may not guarantee the nested object structure is preserved exactly for Ember Data's serialization, potentially causing client-side truncation or type coercion before server verification. Tests only cover server-side. _(flagged by: adversary)_

---

swahlquist:
## Review follow-up (canonicalization + observability)

Addressed the senior-dev review. Verdict on each finding, evaluated against the real code:

- **Markers not canonicalized (agreed, fixed):** `Art50Marker.sign` ignores unsigned keys, so `verify` alone accepted a valid marker padded with arbitrary client keys and both the save path and the cloner persisted them verbatim. Added `Art50Marker.normalized` (verify, then slice to the canonical signed field set) and routed both the save path (`board.rb`) and the clone path (`board_cloner.rb`) through it. Stored markers are now exactly what the server signed; the smuggled-key vector (`evil`/`admin:true`) is closed and tested. The cloner additionally refuses to propagate an at-rest marker that no longer verifies, aligning copy behavior with the `public_view` render path.
- **verify swallows exceptions (agreed, adjusted impl):** the naive 'log inside verify' would flood, since `public_view` calls `verify` on every board render. But forged/invalid markers return false without raising; the rescue only fires on genuine crypto/config faults (e.g. a missing/rotated signing secret). Added a guarded `Rails.logger.error` there (exception class only, no marker contents) so that failure is observable with no flood path and no secret leak.
- **No Ember behavior tests (disagreed on actionability):** this repo's `ember test` only runs lint; QUnit unit specs do not execute (AMD loader defect, issue #314), so an added Ember test would be dead in CI. Both call sites verified by code path review instead.
- **Reviewer attribution (no code change):** the automated PR-bot layer (Claude senior + adversary) is distinct from the manual Claude `adversary` review; no contradiction.

**Verification:** `RAILS_ENV=production zeitwerk:check` clean (the new app to lib `require_relative` has no eager-load conflict). Specs: art50_marker + persistence 36/0, json_api/board 15/0 (no regression), controller Art.50 3/0. Fresh Claude `adversary` re-review of the delta: ship, no Critical/High.

**Documented residuals (no change needed for this PR):** copies drop markers invalidated by a key rotation (consistent with the render path; key rotation already breaks secure_serialize app-wide); the worker-path Art50Marker reference in `process_params` stays rescue-guarded. Both recorded for the Art.50(2) compliance note. The Art.50(2) register entry stays OPEN.

swahlquist:
## Review reconciliation (Codex senior pass + follow-up state)

The Codex senior review ran against `f3d99523`. The current pushed head is `9b3d3faa1` ("canonicalize Art.50(2) marker at persist + clone boundaries"), which landed after that review. Status of each Codex finding against the live head:

**Medium 1 (markers not canonicalized) — RESOLVED in `9b3d3faa1`.** `Art50Marker.normalized(marker)` verifies then slices to the canonical `PERSIST_KEYS`, and both boundaries route through it: the save path (`app/models/board.rb:1895`) and the cloner (`app/cloners/board_cloner.rb:125`). The cloner additionally refuses to propagate an at-rest marker that no longer verifies. Coverage added in `spec/lib/art50_marker_spec.rb` (+35) and `spec/models/board_art50_marking_spec.rb` (+35) proving extra/unsigned keys are stripped and invalid stored markers are not copied.

**Low 1 (rescue converts every exception to false) — RESOLVED in `9b3d3faa1`.** `Art50Marker.verify`'s rescue now logs `"Art50Marker.verify errored: #{e.class}"` (exception class only, no marker contents) so a missing/rotated signing secret is observable. Forged/invalid markers fail via `secure_compare` and never reach the rescue, so there is no log-flood path.

**Medium 2 (frontend marker transport lacks focused tests) — coverage boundary, deferred by decision.** The enforceable server-side contract is tested in `spec/controllers/api/boards_controller_spec.rb:1356-1396`: the marker reaches the save payload and persists (JSON-body path), survives the form-encoded param path, and a forged marker is dropped while the board still saves. A QUnit/Ember behavior test for `create-board-new.js` / `generate-board.js` is deliberately deferred: per issue #314 (AMD loader defect), `ember test` runs lint only and does not execute unit/behavioral specs, so such a test would pass without ever running and would create false assurance on a compliance PR. The server-side controller spec is the enforceable gate for marker transport until #314 is resolved.

**Low 2 (reviewer attribution) — reconciling.** The follow-up commit `9b3d3faa1` records a Claude adversary re-review ("ship, no Critical/High"). Attaching/confirming that result here so the Article 50 review provenance is coherent for the audit trail.

Article 50(2) remains open (slice 2 of the marking work); OBF-export and key-rotation limitations to be recorded per the milestone plan.

swahlquist:
## Claude adversary review (canonical artifact for the Art.50(2) audit trail)

**Reviewer:** Claude (Opus 4.8), `adversary` red-team pass. **Head reviewed:** `9b3d3faa1`. Attack surface: HMAC signing/verification, canonicalization boundary, normalize/verify gating, log/PII hygiene, form-param coercion path, JSON exposure, frontend transport.

**Verdict: SHIP WITH CONDITIONS.** No Critical. No High in the cryptographic/forgery sense: `GoSecure.lite_hmac` keys the outer HMAC-SHA512 with the server `encryption_key`, so a marker cannot be minted or altered without the server secret; `secure_compare` is constant-time; `normalized`/`public_view` correctly gate on `verify` before slicing to `PERSIST_KEYS`. The specs (`spec/lib/art50_marker_spec.rb`, `spec/models/board_art50_marking_spec.rb`) genuinely cover tamper, forgery, unsigned-key stripping, offload round-trip, and bearer overwrite. The persistence and propagation this slice adds are sound.

The findings below have compliance teeth, not forgeability teeth:

**[High] Marking is client-cooperative, not server-authoritative.** The server generates the marker (`lib/ai_board_generator.rb:157`) but does not persist it; it returns `ai_generated` to the client and relies on the browser to echo it back into the board-save payload (`generate-board.js:203,253`, `create-board-new.js:1824`, `models/board.js:165`). Persist only fires if `params['ai_generated']` is present (`app/models/board.rb:1893`). Any client that does not echo the marker (stale cached bundle, JS error mid-flow, a direct `POST /api/v1/boards` API client) persists an AI board **unmarked**, and no server-side record reconciles the board against its `AiApiLog` generation. Marking fails open to "unmarked." **Counter-measure: do NOT record Art.50(2) as satisfied on this slice; track server-authoritative stamping as a required follow-up. Merge the plumbing, keep the obligation open.**

**[Medium] Save-path marking silently no-ops if `Art50Marker` is not loaded.** `board.rb:1895` calls `Art50Marker.normalized` without a `require_relative`, unlike `board_cloner.rb:3` and `json_api/board.rb:2`. `lib/` autoload is skipped when `ENV['RESQUE_WORKER']=='true'` (`config/application.rb:47`). The save rescue converts a `NameError` into a `warn` and drops the marker. The in-code comment argues workers carry no client marker, so the guard should not fire in a worker; still, the load-order dependency is nondeterministic. **Counter-measure: add `require_relative '../../lib/art50_marker'` to `board.rb` for parity; add a spec exercising the persist path with `Art50Marker` guaranteed-undefined.**

**[Medium] Bearer/transplant.** The marker binds provenance, not content (by design, documented `lib/art50_marker.rb:14-27`). A user who runs one AI generation can attach that valid marker to a hand-authored board and the server will verify+persist it (over-marking). Low-harm direction for Art.50(2); ensure the compliance record states the marker is a server-issued provenance/bearer attestation, not a content-integrity claim.

**[Medium] Key rotation silently de-marks all boards.** HMAC keyed by the single app `encryption_key`; the marker carries `sig_alg` but no key-id. Rotating `ENCRYPTION_KEY` makes every existing marker read as unmarked (and drop on next copy). Document that marker durability is bounded by the app-secret lifetime; add key-id + verify-against-previous + re-sign migration if rotation is ever planned. (The app secret also gates all `secure_serialize` data, so it is effectively never rotated.)

**[Low] Log path and canonicalization verified clean (no action).** Both rescues log `e.class` only (no marker contents, board id, or user data); the rescue is unreachable by forged markers (structural checks + `secure_compare` do not raise), so no log-flood path. Canonical signing is a fixed-order `[key,value]` JSON array; `verify` forces every signed field to a non-empty String, so no non-string smuggling. JSON serializer is allowlist and withholds `signature`/`content_id`. One doc nit: the `board.rb:1888` comment says it "never overwrites an existing valid marker," but a valid incoming marker does overwrite an existing valid one (harmless) — tighten the comment.

**Bottom line:** merge the plumbing; keep Article 50(2) OPEN pending server-authoritative stamping; the `require_relative` parity fix is the one cheap code hardening worth landing.

PR #512, opened 2026-07-02
chore: add project-scoped postgres-dev MCP server

swahlquist:
## Summary

Companion to lingolinq/ai-company-brain#110, which removes `postgres-dev` (a direct-SQL Postgres MCP server) from the global WSL Claude Code / Gemini CLI configs — it was reachable from every session, not just this repo. This re-adds it here as a project-scoped `.mcp.json` entry, so it's still available while working in LingoLinq-AAC and nowhere else.

## Review trail (dual-reviewer pattern)

- **Claude adversary**: no Critical/High.
- **Codex senior-dev, round 1** (`Request changes`): flagged the description's "write-capable" claim as false — verified `@modelcontextprotocol/server-postgres` wraps every query in `BEGIN TRANSACTION READ ONLY` + `ROLLBACK`; also flagged the unpinned package and that `metadata.description` isn't a documented Claude Code approval-surface.
- **Codex senior-dev, round 2** (after fixing the above — pinned to `0.6.2`, corrected description): overturned the "read-only" framing again. `node-postgres`'s simple query protocol (`client.query(sql)`, no bound params) permits multi-statement strings, so a query containing `COMMIT` mid-string can end the read-only transaction early and run writes in autocommit mode before the wrapper's own `ROLLBACK`. The description now says so plainly rather than claiming enforced read-only.

## Known open follow-up (not done in this PR)

The real fix is a dedicated Postgres role with enforced read-only grants, not client-side wrapper behavior. Not implemented here.

## Security note

Connection string (`postgresql:///lingolinq-development?host=/var/run/postgresql`) has no embedded credential — local Unix-socket peer auth only. Confirmed safe to commit. Claude Code requires a one-time approval on first load of a project `.mcp.json` server; per the adversary review, that approval doesn't re-gate subagents/worktrees/headless runs after the first approval, and the "synthetic dev DB" assumption in the description is per-machine, not something this committed file can guarantee for any other checkout.


swahlquist:

## Automated PR Review

### Verdict: `COMMENT`

| --- | --- | --- |
| Senior-dev (Claude Sonnet 4.6) | `COMMENT` | 1 |

**Senior-dev:** The entry is honestly documented and correctly scoped, but the connection-string arg is passed via npx args array and will be visible to any process that can read /proc or ps output on a multi-user machine; that is a low-severity note worth capturing. No high-severity findings.

**Adversary:** No new vulnerabilities introduced; PR transparently documents known limitations.

### Findings (1 unique)
- **[LOW]** Connection string exposed in process args (ps/proc-visible) - The DSN is an element of the args array, so it appears in the process argument list. On a shared Linux host any local user can read it via ps aux or /proc//cmdline. Unix-socket peer auth means no password leaks, but the DB name and socket path are visible. Consider passing via env var instead. _(flagged by: senior-dev)_

---

PR #514, opened 2026-07-03
compliance(flusher): implement Flusher.flush_leftovers retention cleanup (LL-991d259b2a)

swahlquist:
## Summary
- Implements `Flusher.flush_leftovers`, previously an empty stub with a 7-item TODO describing retention cleanup that was never built (compliance finding LL-991d259b2a).
- Cleans up orphaned `board_button_images`/`board_button_sounds` join rows, orphaned `log_session_boards`/`user_board_connections` join rows, and stale `Progress` records (>1 month), all via ID-batched scoped deletes (never unbounded `.destroy_all`).
- Reports (does not delete) stale-`item_type` `PaperTrail::Version` rows: `DATA_RETENTION.md` requires 6-year retention + cold-storage archival, and a renamed/removed model class isn't grounds to treat that audit trail as disposable. Filed as follow-up finding `LL-5d7197fa7d`.
- Does **not** delete `board_images`/`board_sounds` themselves via join-table orphan checks — dropped after review found the join tables can be transiently (`board_button_sounds`, async `map_images` deferral) or permanently (`board_button_images`, no longer populated) stale, which would risk deleting live media on a false-positive orphan read.
- Does not implement developer-key expiry: `DeveloperKey` has no expiration concept in this codebase; documented inline rather than invented.
- Each deletion category writes its own `AuditEvent` (`event_type: 'retention_flush'`) atomically with its delete (one transaction per category), preceded by a "planned" event and followed by a final aggregate "completed" event.
- Scheduler now dispatches this job via `Worker.schedule_for(:slow, ...)` instead of the default queue.
- New migration adds indexes (`board_button_sounds.button_sound_id`, `progresses.created_at`, `log_session_boards.log_session_id`) backing the new anti-join/age scans, added `algorithm: :concurrently`.

## Test plan
- [x] `bundle exec rspec spec/lib/flusher_spec.rb spec/lib/tasks/scheduler_spec.rb` — 53 examples, all green except 3 pre-existing unrelated failures from shared test-DB row pollution (tracked separately, PR #367)
- [x] Regression specs for the button_image/button_sound "never touched" race (via real `process_buttons`/`@map_later`, not synthetic joins)
- [x] Orphan removal + non-removal specs for each join-table category
- [x] Stale `PaperTrail::Version` item_type report-only detection (real class survives, renamed/removed class is reported not deleted, non-model constant like `File` handled)
- [x] Per-category `AuditEvent` count verification
- [x] Migration applied locally, `schema.rb` updated
[...]

PR #444, opened 2026-06-19
security(coppa): close org-authored consent bypass + record school-authorization basis

swahlquist:
## What
Phase 0 of the org-authored COPPA fix. **Narrows** a consent-bypass hole to validated org authorizations and makes those authorizations **auditable**. It does NOT, on its own, require parental involvement or a verified school contract for org-authored under-13 accounts; that is Phase 1 (see Honest scope below).

## The bug (closed)
`User#process_params` skipped the under-13 parental-consent gate whenever `authored_organization_id` was a **non-blank string** (the skip keyed on the raw param), while the org/author validation ran later. So a present-but-**invalid or unauthorized** `authored_organization_id` bypassed the COPPA gate **and recorded nothing**. A client could present any non-blank string and skip the under-13 gate.

## The fix
- Compute the validated org authorization **once** (org resolves AND author is a manager with `edit`) before the COPPA block, and gate the skip on that. Invalid/unauthorized org ids now fall through to the parental-consent gate.
- Record `settings['school_authorization']` (basis, org, authorizing manager, timestamp, `record_id`).
- Emit an immutable `AuditEvent` (`event_type: school_authorization`) post-save in `api/users#create`, **fail-open** so a failed audit insert can never orphan or 500 the child account.

## Adversary review (pre-merge red-team) — addressed
Verdict was MERGE-WITH-FIXES. Applied:
- **High #2 (must-fix):** audit emission is now fail-open with guarded PII-scrubbed loud logging (mirrors `AuditEvent.log_command`). Spec stubs the audit to raise and asserts the account is still created.
- **High #1 (decision flagged):** `allows?('edit')` admits **assistant-level** managers, not only full managers (`organization.rb:43` vs `:44`). This PR **preserves the pre-existing authoring scope** and documents it with a code comment + a spec. **Open decision for Scot:** restrict the school-official exception to full managers (`'manage'`) and/or a signed-contract flag? Deferred to Phase 1.
- **Medium #3:** comment noting any path that sets `school_authorization` must also emit the audit (today only `api/users#create` creates org-authored users).

## Honest scope
With the signed-contract precondition deferred to Phase 1, any org manager or assistant with `edit` can still author under-13 accounts with no parental involvement. This PR makes that **audited and limited to validated authorizations**, not **authorized** in the full COPPA sense. Phase 1 (org contract flag, full-manager scope decision, re-consent on contract change) is tracked in `outputs/plans/2026-06-19-org-coppa-bypass-fix-scope.md`.

## Tests
[...]

PR #528, opened 2026-07-04
compliance(users): emit AuditEvent for every account creation (LL-d35cbdb313)

swahlquist:
## Summary
- Compliance finding LL-d35cbdb313 (FERPA, accepted 2026-06-23): `api/users#create` persisted new accounts with no audit trail, except the narrow `school_authorization` event (#444) that only fires for org-authored under-13 accounts. Plain self-registration and ordinary org start-code signups left no creation record.
- Adds a second, general-purpose `user_creation` AuditEvent that fires for **every** created account, recording the authenticated author (nil for self-registration), whether a start code was used, and the authoring organization's global_id when the code resolves to one.
- Additive, not a replacement: the existing `school_authorization` event is untouched and still separately records the COPPA authorization basis for org-authored minors. Same fail-open pattern as that event (audit-insert failure logs loudly, PII-scrubbed, never 500s or orphans the already-persisted account).

## Test plan
- [x] 4 new controller specs: plain signup, start-code signup records the org, org-authored minor records both events, fail-open on audit-insert error
- [x] Full `create` block: 43 examples, 0 failures, 1 pre-existing unrelated pending
- [x] `ruby -c` + rubocop on both changed files — no new offenses introduced

## Follow-up
Register attestation (marking LL-d35cbdb313 `verified-closed`) is a separate PR after this merges, matching the pattern used for LL-991d259b2a (#514 shipped the fix, #527 attested it).

### DIFF
--- a/app/controllers/api/users_controller.rb
+++ b/app/controllers/api/users_controller.rb
@@ -252,6 +252,7 @@ class Api::UsersController < ApplicationController
     end
     user = User.process_new(user_data, {:pending => true, :author => @api_user})
     start_progress = nil
+    start_code_org = nil
     if !user || user.errored?
       return api_error(400, {error: "user creation failed", errors: user && user.processing_errors})
     end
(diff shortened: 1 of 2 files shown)


swahlquist:
## Compliance PR Checklist

PR #531, opened 2026-07-05
fix(uploads): set UPLOADS_S3_NO_ACL=1 for the private uploads bucket

swahlquist:
## What

Adds `UPLOADS_S3_NO_ACL=1` to `APP_ENV_VARS_STATIC` in the Cloud Run deploy workflow (covers web + worker).

## Why

Continuation of finding **LL-705b10bcd7** (S3 SigV4/KMS-SSE upload failure on the GCP rehearsal stack). The SigV4 rewrite (77233fb19, afe03d2de) is deployed and correct; the remaining failures are AWS-side, and there are two stacked blockers:

1. **KMS (admin action, not this PR):** `lingolinq-prod-uploads` has SSE-KMS default encryption (CMK `d5a1a377-...`) and the runtime principal is denied `kms:GenerateDataKey`, so every write fails. The needed statement already exists in `scripts/gcp/iam/lingolinq-cloudrun-s3-ses-policy.json`; it has to be attached with admin credentials.
2. **ACL (this PR):** the bucket is `BucketOwnerEnforced` + full Block Public Access, so once KMS is fixed the presigned POST would still be rejected for carrying `acl=public-read`. `lib/uploader.rb` already supports `UPLOADS_S3_NO_ACL` but nothing set it.

Access model decision (Scot, 2026-07-05): keep the bucket private (HIPAA/FERPA posture) rather than restoring legacy public-read parity. Client-facing reads will need presigned GETs / CDN with OAC; that read-path audit is a follow-up, not this PR.

## Verification

- Both blockers confirmed with read-only checks plus a synthetic `put-object` (denied at KMS; nothing written; test key deleted).
- The same env var was applied live to the rehearsal stack alongside this PR so the next synthetic ButtonImage test can exercise it once KMS perms land.

https://claude.ai/code/session_011i7vwUmphP4z5sa8BubCsy

### DIFF
--- a/.github/workflows/deploy-cloudrun.yml
+++ b/.github/workflows/deploy-cloudrun.yml
@@ -115,7 +115,12 @@ jobs:
       # Block Public Access, per the private-bucket model (Scot, 2026-07-05). A presigned POST
       # carrying acl=public-read is rejected by such a bucket, so lib/uploader.rb must omit the
       # ACL field. Client-facing reads go through presigned GETs / CDN, not public object URLs.
-      APP_ENV_VARS_STATIC: AWS_REGION=us-west-2,SES_REGION=us-west-2,UPLOADS_S3_BUCKET=lingolinq-prod-uploads,STATIC_S3_BUCKET=lingolinq-prod-static,UPLOADS_S3_NO_ACL=1,SENTRY_ENVIRONMENT=production,RAILS_LOG_TO_STDOUT=true,RAILS_SERVE_STATIC_FILES=true,RUBY_YJIT_ENABLE=1,WEB_CONCURRENCY=2
+      # UPLOADS_S3_CDN: CloudFront distribution E2X2HAS6Y1L2MI fronts the private bucket via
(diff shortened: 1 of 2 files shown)


swahlquist:
Scope grew after the read-path audit: commit cb1ee57f5 adds UPLOADS_S3_CDN pointing at a new CloudFront distribution (E2X2HAS6Y1L2MI) that fronts the private bucket via Origin Access Control. Infra applied 2026-07-05: OAC, distribution (GET/HEAD/OPTIONS, CachingOptimized, CORS-S3Origin origin-request policy, SimpleCORS response headers), bucket policy scoped to the distribution ARN (replacing a dead legacy public-read statement), KMS key-policy grant for cloudfront.amazonaws.com scoped to the same ARN, and S3 CORS (GET/HEAD, any origin). Verified: KMS-encrypted test object served 200 via the CDN with access-control-allow-origin while direct anonymous bucket access stays 403. Full consumer inventory + sequencing: ai-company-brain outputs/plans/2026-07-05-s3-private-bucket-read-path-scope.md

PR #535, opened 2026-07-05
fix(compliance): narrow DataPolicyEnforcer's session-only retention gap

swahlquist:
## Summary
- `DataPolicyEnforcer.enforce_retention!` filtered `LogSession` to `log_type: 'session'` only, so notes, quick assessments, evals, and journal entries never expired under an org's `retention_months` policy, even though `docs/legal/DATA_RETENTION.md` documents the whole "Communication logs (LogSession)" category as governed by this nightly job (GDPR 5(1)(e); FERPA).
- Extends the retention filter to `session`, `note`, `assessment`, `eval`, `journal` -- the same discrete, timestamped log types `api/logs_controller.rb` already exposes as browsable communication logs, minus `profile` (see below).
- Deliberately excludes `daily_use` and `modeling_activities`: verified via `LogSession.process_daily_use`/`process_modeling_event` that these are per-user singleton trackers (`find_or_create_by` on `log_type`+`user_id`) whose `started_at` freezes at first creation and never advances, so an age-based purge would delete a still-in-use record for any long-tenured active user rather than a stale one.
- Deliberately excludes `profile`: Codex's review of this PR found `UserExtra#process_profile` caches each profile session's identifiable summary + `global_id` into `UserExtra.settings['recent_profiles']` and `UserLink.data['state']['profile_history']`, and nothing refreshes those caches when the source log is destroyed -- purging `profile` today would delete the source record while leaving identifiable content behind elsewhere. Filed **LL-caf2528468** to track the cache-invalidation work needed before `profile` can be added back.
- Boards, button images/sounds, and `AiApiLog` are out of scope: each has its own documented, separate retention mechanism in `DATA_RETENTION.md` unrelated to org `retention_months`.

**This narrows, but does not close, LL-1890f6a922.** Adversary review found three more pre-existing gaps on axes this PR doesn't touch, each real enough that framing this as a full fix would overclaim:
- **LL-14edf1a801** (medium): `enforce_retention!`'s candidate-org query only matches orgs whose *own* `data_policy_version` is set, so a child org that merely *inherits* `retention_months` from a parent (`Organization#effective_data_policy`) is skipped entirely -- its users' logs never expire.
- **LL-3bb2e2eaad** (medium): `Flusher.flush_record` destroys the `LogSession` and immediately deletes its own PaperTrail version, and this job writes no `AuditEvent` -- contradicts `DATA_RETENTION.md`'s row committing PaperTrail versions to 6 years (HIPAA), and leaves no record of what was purged or when.
- **LL-de9c94bf36** (low): the purge is scoped to `user_id` only, so one sponsoring org's policy reaches a user's entire log history, including logs outside that org's context -- flagged as a design question for Scot, not asserted as a defect.

All three are pre-existing (this PR only changed the `log_type` filter) and are left open/untriaged in the register rather than fixed here -- same reasoning as the `profile` deferral: scope-creeping this PR into the org-iteration and audit-trail logic would be a materially larger, riskier change deserving its own dedicated review.

## Test plan
- [x] `spec/lib/data_policy_enforcer_spec.rb`: no-op with no data policy / zero retention_months, purge of stale `session`/`note`/`assessment`/`eval`/`journal` logs, `daily_use`/`modeling_activities`/`profile` never purged regardless of age, logs outside the org's sponsorship untouched, near-boundary retention window, `eval` with nil `started_at` never purged.
- [x] `bundle exec rspec spec/lib/data_policy_enforcer_spec.rb` -- 8 examples, 0 failures.
- [x] `bundle exec rspec spec/models/organization_spec.rb spec/models/log_session_spec.rb` green (no regressions); `spec/lib/flusher_spec.rb` has 3 pre-existing failures unrelated to this change (stale `ButtonImage` row count in the shared test DB, a known issue).
- [x] Codex senior-dev review (`/review-pr`) and adversary review both run; findings addressed or filed as follow-ups above.

### DIFF
--- a/audit-reports/DOCUMENT-REGISTER.json
+++ b/audit-reports/DOCUMENT-REGISTER.json
@@ -605,7 +605,7 @@
       "lastReviewed": "2026-06-19",
       "nextReviewDue": "2026-09-19",
       "attestation": {},
-      "contentHash": "99e071130588610e9a34d388691f6d771a1774dbb10f2a72eac46c855654825c",
+      "contentHash": "92f126ee2288e6ddfb4e169b71419c8d2029886ac11c81c18ca278029104f4fe",
       "bundles": [],
       "mirrors": [
         {
--- a/audit-reports/DOCUMENT-REGISTER.json
+++ b/audit-reports/DOCUMENT-REGISTER.json
@@ -605,7 +605,7 @@
       "lastReviewed": "2026-06-19",
       "nextReviewDue": "2026-09-19",
       "attestation": {},
-      "contentHash": "92f126ee2288e6ddfb4e169b71419c8d2029886ac11c81c18ca278029104f4fe",
+      "contentHash": "7171d953aaa17beb8ba9323f6f47986d2f0dac11b2c37f4922ce6fd93b4ed86e",
       "bundles": [],
       "mirrors": [
         {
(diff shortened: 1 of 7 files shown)


swahlquist:
Codex's senior-dev review flagged a real P1: extending retention to `profile` logs would delete the source `LogSession` while `UserExtra#process_profile` (app/models/user_extra.rb:40-97) leaves stale identifiable summaries behind in `UserExtra.settings['recent_profiles']` and `UserLink.data['state']['profile_history']`, since nothing refreshes those caches on destroy. Verified this against the code directly.

Response: dropped `profile` from `RETAINABLE_LOG_TYPES` (same treatment as `daily_use`/`modeling_activities`, just for a different reason) rather than shipping a half-fixed purge, and filed **LL-caf2528468** to track the cache-invalidation work needed before `profile` can be added back. Checked `note`/`assessment`/`eval`/`journal` for the same pattern — none found (verified against `user_badge.rb` and `weekly_stats_summary.rb`).

Updated PR now excludes `profile`; spec updated to assert it's never purged, same as `daily_use`/`modeling_activities`.

PR #540, opened 2026-07-06
fix(security): protected_image no longer leaks a permanent user_token

swahlquist:
## Summary
- Plan E of the Fable 5 Blitz sprint (LL-310b464be4): `protected_image` (`api/users_controller.rb`) accepted a **permanent, non-expiring** `user_token` via URL query param -- once captured (access logs, browser history, or a cross-origin Referer header on the fallback/CDN redirect), it stays valid forever.
- Adds `User#protected_image_token` / `User.find_by_protected_image_token`: a 30-day signed, expiring token, modeled on the existing `GoSecure` token-signing primitive already used elsewhere in this codebase (`GoSecure.browser_token`/`valid_browser_token?`), but using epoch-timestamp math rather than that helper's day-of-year arithmetic.
- 30-day expiry deliberately outlasts the 12-day CDN `expires_in` window `protected_image` already sets on successful redirects, so it won't expire while a synced board is still relying on the cached copy.
- Consolidates the token-append logic, previously duplicated identically in three places (`Uploadable#url_for`, `ButtonImage#settings_for` x2), into one shared `Uploadable.tokenize_protected_image_url` helper.
- The client independently reconstructs `protected_image` URLs in several places (`LingoLinq.Image.personalize_url` and its ~7 callers) using the raw `currentUser.user_token` value transmitted at login. Added a dedicated `protected_image_token` field (server: `lib/json_api/user.rb`; client: `models/user.js`) so the client uses the new expiring token for this purpose instead of the permanent one, without changing the semantics of `user_token` itself (still used, out of scope, for public board-by-token rendering and lesson viewing).

## This narrows, but does not close, LL-310b464be4

Adversary review of this PR correctly flagged that the fix is incomplete as a security remediation, not just as a compatibility nicety:

- `find_by_protected_image_token` still accepts the **legacy permanent format** indefinitely (2-segment token -> `find_by_token`), by deliberate design -- so boards/images already cached client-side (offline/IndexedDB) before this ships keep resolving after deploy. `User#global_id` is confirmed hyphen-free for User records (the `@sub_id` suffix mechanism is Board-only), so the segment count safely disambiguates old vs. new format.
- But `user_token` itself is **unchanged and still broadly exposed**: it's still serialized to every client on login (`lib/json_api/user.rb:41`), still embedded in lesson-share links and iframe `data-user_token` attributes, and still valid forever. Anyone who captures a `user_token` through any of those surfaces can still hand-build a `protected_image` URL that authenticates permanently -- this PR only stops *newly generated* image URLs from carrying that permanent token; it does not revoke or time-box the token itself.
- Net effect: this PR closes the "new image URLs leak a forever-valid credential" gap and adds observability for the deferred one, but the underlying permanent-token exposure this finding named is still live via the fallback and via `user_token`'s other uses. **Do not mark LL-310b464be4 closed on this PR** -- it should stay open, narrowed, with this residual documented (matching how PR #535 narrowed rather than closed LL-1890f6a922).
- Added: the legacy-fallback branch now logs each time it's hit (`[protected_image_legacy_token]`, includes the resolved user's global_id, not the token) so a future decision to sunset it has usage evidence instead of a guess.
- Also surfaced by adversary review: the 30-day expiry means an exported `.obf`/`.obz` file containing protected images (`lib/converters/lingo_linq.rb:192,268` embed the token via the same `url_for`/`settings_for` path) will have those specific image links stop working 30 days after export, with no refresh mechanism (unlike a live device, which reconnects and gets a fresh token). This is a real usability tradeoff of the 30-day window, not a defect -- flagging it since it wasn't visible when the window was chosen. Now pinned by a regression test (see below).
- Out of scope (per the register finding, which cites only `protected_image`): `User#user_token`/`User.find_by_token` themselves, and their other callers (`boards_controller.rb` public board rendering, `api/lessons_controller.rb` lesson viewing).

## Round 3: Copilot review + final adversary pass (commit 0da283509)

Copilot reviewed the PR after the logging commit and found two more Medium issues, both fixed:

- **Constant-time signature compare**: `find_by_protected_image_token`'s signature check used `==` (variable-time). Switched to `ActiveSupport::SecurityUtils.secure_compare` to remove the timing side-channel. (The legacy `find_by_token` path still uses `==`, called out by the follow-up adversary pass as a pre-existing Low, deferred to the eventual `user_token` retirement rather than fixed here.)
- **OBZ export regression test**: added a test (`spec/lib/converters/lingo_linq_spec.rb`) that exports a board with a protected image, confirms the embedded token resolves to the exporting user immediately, then time-travels 31 days and confirms the same URL now correctly fails to resolve -- pinning the export-consequence noted above as tested behavior, not just prose.
- The automated compliance-PR-checklist COPPA checkbox on this PR is checked: the only `users_controller.rb` change is the token-lookup swap inside `protected_image`, no consent/registration/data-collection logic is touched.
- A fresh adversary review was run against this exact commit (not the earlier one from before the logging/Copilot fixes). Verdict: **Approve**, no Critical/High findings. Three Low/informational notes, none requiring further action this PR: (1) the legacy fallback's `find_by_token` still isn't constant-time -- accepted, tracked with the eventual `user_token` retirement; (2) the legacy-fallback log line is per-request rather than a sampled metric -- fine for now, only logs a global_id, no PII; (3) confirmed the OBZ-embedded token really is a 30-day bearer credential for the exporter's own protected symbol images (not student PII) -- a strict improvement over the prior permanent token, already covered by the new test.

## Test plan
- [x] `spec/models/user_spec.rb`: new `protected_image_token`/`find_by_protected_image_token` coverage -- valid round-trip, custom lifespan, expired token, tampered signature, tampered expiry, non-numeric expiry, garbage/nil input, legacy-format fallback, legacy-fallback logging (present and absent).
- [x] `spec/controllers/api/users_controller_spec.rb`: new cases for the expiring-token format and an expired-token rejection; all pre-existing legacy-`user_token` cases left unchanged (they pass unmodified, validating the backward-compat fallback).
- [x] `spec/models/button_image_spec.rb` / `spec/models/concerns/uploadable_spec.rb`: updated to assert the token round-trips to the correct user via `find_by_protected_image_token` rather than a hardcoded string (avoids flakiness from the token's embedded timestamp).
- [x] `spec/lib/converters/lingo_linq_spec.rb`: new OBZ export test (see Round 3 above).
- [x] `bundle exec rspec spec/models/user_spec.rb spec/models/button_image_spec.rb spec/models/concerns/uploadable_spec.rb spec/controllers/api/users_controller_spec.rb spec/lib/converters/lingo_linq_spec.rb` -- 861 examples, 0 failures attributable to this change (11 pre-existing `AuditEvent.count` baseline failures reproduce identically on an unrelated, unmodified branch -- known shared-test-DB drift, not caused by this PR).
- [x] Three review passes complete: Codex senior-dev (`/review-pr`), Claude adversary (twice -- pre- and post-Copilot-fixes), and GitHub Copilot. All findings addressed or documented above as accepted residuals for Scot's triage.

### DIFF
--- a/app/controllers/api/users_controller.rb
+++ b/app/controllers/api/users_controller.rb
@@ -942,7 +942,7 @@ class Api::UsersController < ApplicationController
 
   def protected_image
     user = User.find_by_path(params['user_id'])
-    api_user = User.find_by_token(params['user_token'])
+    api_user = User.find_by_protected_image_token(params['user_token'])
     valid_result = nil
     if !api_user
       expires_in 30.minutes, :public => true
--- a/app/frontend/app/controllers/application.js
+++ b/app/frontend/app/controllers/application.js
@@ -1692,7 +1692,7 @@ export default Controller.extend({
     if(image && image.get('personalized_url') && !button.no_skin) {
       image_url = image.get('personalized_url');
     } else if(button.get('original_image_url') && LingoLinqImage.personalize_url) {
-      image_url = LingoLinqImage.personalize_url(button.get('original_image_url'), _this.appState.get('currentUser.user_token'), _this.appState.get('referenced_user.preferences.skin'), button.no_skin);
+      image_url = LingoLinqImage.personalize_url(button.get('original_image_url'), _this.appState.get('currentUser.protected_image_token'), _this.appState.get('referenced_user.preferences.skin'), button.no_skin);
     }
     var obj = {
       label: button.label,
--- a/app/frontend/app/models/buttonset.js
+++ b/app/frontend/app/models/buttonset.js
@@ -779,7 +779,7 @@ LingoLinq.Buttonset = BaseModel.extend({
             }
             emberSet(button, 'image', emberGet(button, 'image') || templateHelpers.path('images/blank.gif'));
             if(emberGet(button, 'image') && LingoLinqImage.personalize_url) {
-              emberSet(button, 'image', LingoLinqImage.personalize_url(button.image, _this.appState.get('currentUser.user_token'), _this.appState.get('referenced_user.preferences.skin'), button.no_skin));
+              emberSet(button, 'image', LingoLinqImage.personalize_url(button.image, _this.appState.get('currentUser.protected_image_token'), _this.appState.get('referenced_user.preferences.skin'), button.no_skin));
             }
             emberSet(button, 'on_same_board', emberGet(button, 'steps') === 0);
   
@@ -1188,7 +1188,7 @@ LingoLinq.Buttonset = BaseModel.extend({
 
 LingoLinq.Buttonset.fix_image = function(button, images) {
   if(button.image && LingoLinqImage.personalize_url) {
-    button.image = LingoLinqImage.personalize_url(button.image, app_state.get('currentUser.user_token'), app_state.get('referenced_user.preferences.skin'), button.no_skin);
+    button.image = LingoLinqImage.personalize_url(button.image, app_state.get('currentUser.protected_image_token'), app_state.get('referenced_user.preferences.skin'), button.no_skin);
   }
   var image = images.find(function(img) { return img.get('id') === button.image_id; });
   if(image) {
--- a/app/frontend/app/models/image.js
+++ b/app/frontend/app/models/image.js
@@ -118,12 +118,12 @@ LingoLinq.Image = BaseModel.extend({
       var alternate = (this.get('alternates') || []).find(function(a) { return a.library == preferred_symbols; });
       if(alternate) { url = rewriteBrokenSymbolUrl(alternate.url); }
     }
-    return LingoLinq.Image.personalize_url(url, this.get('appState.currentUser.user_token'), this.get('appState.referenced_user.preferences.skin'), LingoLinq.Image.unskins[this.get('id')]);
+    return LingoLinq.Image.personalize_url(url, this.get('appState.currentUser.protected_image_token'), this.get('appState.referenced_user.preferences.skin'), LingoLinq.Image.unskins[this.get('id')]);
   },
-  personalized_url: computed('url', 'appState.currentUser.user_token', 'appState.referenced_user.preferences.skin', 'appState.referenced_user.preferences.preferred_symbols', 'appState.edit_mode', function() {
+  personalized_url: computed('url', 'appState.currentUser.protected_image_token', 'appState.referenced_user.preferences.skin', 'appState.referenced_user.preferences.preferred_symbols', 'appState.edit_mode', function() {
     return this.personalizing_url();
   }),
-  personalized_url_without_preferred_symbols: computed('url', 'appState.currentUser.user_token', 'appState.referenced_user.preferences.skin', 'appState.referenced_user.preferences.preferred_symbols', 'appState.edit_mode', function() {
+  personalized_url_without_preferred_symbols: computed('url', 'appState.currentUser.protected_image_token', 'appState.referenced_user.preferences.skin', 'appState.referenced_user.preferences.preferred_symbols', 'appState.edit_mode', function() {
     return this.personalizing_url(true);
   }),
   best_url: computed('personalized_url', 'appState.referenced_user.preferences.preferred_symbols', 'data_url', function() {
--- a/app/frontend/app/models/user.js
+++ b/app/frontend/app/models/user.js
@@ -34,6 +34,7 @@ LingoLinq.User = BaseModel.extend({
   },
   user_name: attr('string'),
   user_token: attr('string'),
+  protected_image_token: attr('string'),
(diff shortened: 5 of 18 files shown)


swahlquist:
## Compliance PR Checklist

- [x] COPPA consent flow reviewed (users_controller changed) - the only change in `users_controller.rb` is swapping the token lookup used by `protected_image` (`find_by_token` -> `find_by_protected_image_token`); no consent, registration, or data-collection logic is touched. (Rechecked at commit 0da283509 after the checklist bot reset this box on the last push.)

swahlquist:

## Automated PR Review

### Verdict: `COMMENT`

| --- | --- | --- |

**Senior-dev:** The security fix is well-structured and the test coverage is thorough. One correctness issue exists in the token-parsing logic: User global_ids contain hyphens (format N_M with an underscore, but the split is on -), so the 3-part check may be fragile; more critically, the legacy-fallback log line will raise if find_by_token returns nil (no guard on the nil case before calling .global_id). The .obz...

**Adversary:** PR reduces exposure of permanent user_token in protected_image URLs but leaves a residual security gap: the legacy permanent token is still accepted with logging but not blocked, and user_token remains valid forever via other channels.

### Findings (5 unique)
- **[HIGH]** find_by_protected_image_token calls .global_id on potentially-nil user before guard - Line Rails.logger.info("...") if user is correct, but if find_by_token raises before returning nil (e.g. DB error) the guard is bypassed. More concretely: the parts-length guard (unless parts.length == 3) sends ANY 3-segment string down the new path, but a legacy token that happens to contain exactl... _(flagged by: senior-dev)_
- **[MED]** Protected-image token baked into exported .obz files expires after 30 days with no refresh path - As acknowledged in the PR description, url_for/settings_for embed the token at export time via tokenize_protected_image_url. An .obz opened 31+ days after export will have broken protected images. This is a real regression for any offline/archive workflow. Consider embedding the legacy permanent tok... _(flagged by: senior-dev)_
- **[MED]** Legacy token fallback remains an indefinite exposure - find_by_protected_image_token accepts the old permanent user_token indefinitely with only logging. An attacker who captured a permanent token before this change can still use it forever; the fallback should have a sunset plan. _(flagged by: adversary)_
- **[LOW]** url_for now returns self.url unchanged when user is nil (behaviour change from prior code) - Prior code explicitly return self.url if !token before the regex check. The new tokenize_protected_image_url also returns url when user is nil (first guard), so the nil-user path is preserved. However, the prior code also returned early when the URL did NOT match the protected_image pattern, whereas... _(flagged by: senior-dev)_
- **[LOW]** Export usability regression for protected images - Exported .obf/.obz files embed the 30-day expiring token; after expiration, images in exported files will break with no refresh mechanism. This is acknowledged but may impact long-term archive usability. _(flagged by: adversary)_

---

swahlquist:
Verified the [HIGH] and one [LOW] finding from the automated review above (comment 4895058516) against the code at 0da283509; both are false positives from an incorrect premise about `User#global_id`.

**[HIGH] "calls .global_id on potentially-nil user" / 3-part-collision concern -- false.**
- `Rails.logger.info(...) if user` already guards the nil case; `.global_id` is never called on nil.
- The deeper concern (a legacy 2-part token could theoretically split into 3 parts and get misrouted into the new-format verifier) requires either `User#global_id` or the legacy hash segment to contain a `-`. Neither can: `global_id.rb:12` builds User ids as `"1_#{id}"` (no hyphen), and the `@sub_id`-appends-a-hyphen mechanism (`global_id.rb:14-16`) is gated `self == Board` everywhere it's set (verified all 7 call sites) -- it never fires for a `User` instance calling its own `.global_id`. The legacy token's hash half is `GoSecure.sha512(...)` = `Digest::SHA512.hexdigest(...)` (go_secure-0.70/lib/go_secure.rb:6-7) -- pure hex, never a hyphen. So a legacy token always splits into exactly 2 parts, a new-format token always splits into exactly 3; there's no collision case.

**[LOW] "url_for now returns self.url unchanged when user is nil (behaviour change)" -- false, and actually a robustness improvement.**
- Old: `token = user && user.user_token; return self.url if !token` -- if `user` is nil this returns `self.url` unchanged, identical to new behavior.
- Actually the old code had a latent bug the new code fixes: if `self.url` were nil while `user` was present, `self.url.match(...)` would raise `NoMethodError`. The new `Uploadable.tokenize_protected_image_url` guards `url &&` before matching, so a nil `url` now returns nil instead of crashing.

The two [MED] findings (legacy-token indefinite exposure, 30-day OBZ export expiry) aren't new -- both are the intentional, already-documented residuals from the "This narrows, but does not close, LL-310b464be4" section of the PR description, and the export one is now covered by a regression test in `spec/lib/converters/lingo_linq_spec.rb`.

PR #543, opened 2026-07-06
fix(scripts): fail closed in sync-render-env when env-var read fails (hourly --apply wipe guard)

swahlquist:
## Why this is urgent

This repo's `scripts/sync-render-env.js` runs **hourly with `--apply`, unattended**, via `.github/workflows/sync-render-secrets.yml`. It carries the same wipe bug found as a near-miss in ai-company-brain during the 2026-07-06 Render key rotation (fixed there in lingolinq/ai-company-brain#123): a failed read of a service's current env vars degraded to `[]`, and Render's env-vars PUT replaces the **entire** list.

Failure chain under the old code: the `RENDER_API_KEY` repo secret gets rotated/revoked → the hourly dry-run's reads 401 and read as empty → every managed key prints as `NEW (24)` → the workflow's `grep` change detector sets `has_changes=true` → the apply step PUTs only the managed keys → **every unmanaged var on dev/staging/prod is wiped** (`UPLOADS_S3_NO_ACL`, `UPLOADS_S3_CDN`, `LEADER_POSTGRES_URL`, ...), with a green "changes applied" notification.

## Changes

- **Fail-closed reads**: `getRenderEnvVars` throws on any error or non-array response instead of returning `[]`
- **Pagination**: follows Render's cursor so >100-var services are never silently truncated; repeated-cursor + 50-page guards throw
- **Apply guards**: skip a service whose read fails; refuse to apply against a zero-var read (deliberately no override flag); failed PUTs recorded; `main()` exits 1 when anything was skipped
- **Workflow**: dry-run step propagates the script's exit code (was `|| true`), so a failed read fails the run and fires the existing ❌ notification instead of reading as "no changes"
- **audit**: reports "Comparison INCOMPLETE" instead of a false all-clear on unreadable envs
- **exportToOp**: fix `OP_VAULT` ReferenceError + plaintext-output warning
- **Tests**: `scripts/sync-render-env.test.js`, 25 tests over a mocked `https.request` (no real network), run under Node 20 — including "failed dry-run output contains no NEW/CHANGED lines", which pins the exact input the workflow's change detector greps

## Review lineage

The fix design had a red-team (adversary) pass on the brain PR; this port was the adversary's one **Critical** finding (unattended hourly apply runs the unfixed copy). Script diff touches `scripts/` + `.github/workflows/` only — no fixtures, seeds, or data-bearing paths.

## Verification

- 25/25 tests pass under Node 20 (`node scripts/sync-render-env.test.js`)
- Same guard verified live in the brain repo against the real Render API: revoked key → all services refused with explicit errors, exit 1; valid key → clean reads, exit 0

### DIFF
--- a/.github/workflows/sync-render-secrets.yml
+++ b/.github/workflows/sync-render-secrets.yml
@@ -58,7 +58,13 @@ jobs:
           OP_RENDER_SYNC_TOKEN: ${{ secrets.OP_RENDER_SYNC_TOKEN }}
           RENDER_API_KEY: ${{ secrets.RENDER_API_KEY }}
         run: |
-          OUTPUT=$(node scripts/sync-render-env.js --source op 2>&1 || true)
+          # Capture output for the change detector, but PROPAGATE a non-zero
+          # exit: the script exits 1 when it could not read a service's
+          # current env vars (fail-closed wipe guard), and that must fail
+          # this run (triggering the failure notification) rather than be
+          # swallowed and read as "no changes".
+          DRYRUN_EXIT=0
+          OUTPUT=$(node scripts/sync-render-env.js --source op 2>&1) || DRYRUN_EXIT=$?
           echo "$OUTPUT"
           echo "output<<EOF" >> $GITHUB_OUTPUT
           echo "$OUTPUT" >> $GITHUB_OUTPUT
@@ -69,6 +75,11 @@ jobs:
             echo "has_changes=true" >> $GITHUB_OUTPUT
           else
             echo "has_changes=false" >> $GITHUB_OUTPUT
+          fi
+
+          if [ "$DRYRUN_EXIT" -ne 0 ]; then
+            echo "::error::sync-render-env dry-run failed (exit $DRYRUN_EXIT) -- see log; not applying."
+            exit "$DRYRUN_EXIT"
           fi
 
       - name: Apply changes to Render
--- a/scripts/sync-render-env.js
+++ b/scripts/sync-render-env.js
@@ -198,16 +198,43 @@ function renderApiRequest(method, reqPath, body = null) {
 }
 
 async function getRenderEnvVars(serviceId) {
-  try {
-    // Render API paginates at 20 by default -- use limit=100 to get all
-    const result = await renderApiRequest('GET', `/services/${serviceId}/env-vars?limit=100`);
-    if (!Array.isArray(result)) return [];
+  // Fail CLOSED: any failure here must THROW, never degrade to []. Render's
+  // env-vars PUT replaces the service's ENTIRE list, and the apply path
+  // builds its payload from this read -- so a swallowed error (e.g. a 401
+  // from a revoked API key) that returned [] would turn --apply into "wipe
+  // every unmanaged var on prod" (2026-07-06 near-miss during the Render
+  // key rotation; this script runs hourly with --apply via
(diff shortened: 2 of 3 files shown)


swahlquist:
**Update: this is not just latent — the failure chain was executing live.**

Run [28816364644](https://github.com/lingolinq/LingoLinq-AAC/actions/runs/28816364644) (2026-07-06 19:06 UTC, hourly cron) shows the `RENDER_API_KEY` repo secret was never updated in today's rotation:

- dry-run: 401 on all three services → each read as empty → `NEW (23)` → `has_changes=true`
- **the apply step ran** and attempted the wipe-PUT on dev, staging, and prod
- the only reason nothing was wiped is that the PUTs 401'd with the same revoked key
- the workflow then posted `✅ Render secrets sync applied changes` to Google Chat — a false success

Interim measures taken:
1. `sync-render-secrets.yml` schedule **disabled** (it was syncing nothing and sending false success notifications)
2. Repo secret `RENDER_API_KEY` still needs updating to the rotated key (blocked from doing this autonomously; command below)

```
op read "op://LingoLinq Admin/Render API/RENDER_API_KEY" | gh secret set RENDER_API_KEY --repo lingolinq/LingoLinq-AAC
```

Re-enable order after merge: update secret → `gh workflow enable sync-render-secrets.yml` → `gh workflow run sync-render-secrets.yml -f dry_run=true` and confirm clean reads.

swahlquist:

## Automated PR Review

### Verdict: `COMMENT`

| --- | --- | --- |
| Senior-dev (Claude Sonnet 4.6) | `APPROVE` | 1 |

**Senior-dev:** The PR correctly hardens the fail-closed wipe guard, adds pagination, propagates exit codes in the workflow, and ships tests. No i18n, Ember, PII, or feature-flag concerns apply to a Node.js ops script.

**Adversary:** The PR introduces robust fail-closed mechanisms to prevent catastrophic env var wiping, but the change to GitHub Actions workflow may break notifications or create a race condition on partial failure detection.

### Findings (5 unique)
- **[MED]** Workflow change may break failure notification flow - The workflow now fails the dry-run step on any read error instead of swallowing the exit code. This triggers the existing failure notification, but if that notification was originally designed for apply failures only, it may now fire too frequently or with inappropriate content for read errors. _(flagged by: adversary)_
- **[MED]** Race condition in workflow change detection - The workflow sets has_changes=true based on grep for NEW/CHANGED BEFORE checking the exit code. If the script outputs diff lines then fails with exit 1, the workflow might incorrectly think there are changes even though the run failed. _(flagged by: adversary)_
- **[LOW]** syncFailures is a module-level mutable array shared across test resets - resetState() does syncFailures.length=0, which works because the export is the live array reference. This is fragile: if a future test imports a second instance or the array is replaced rather than truncated, resets silently stop working. Consider exporting a reset helper or returning failures from ... _(flagged by: senior-dev)_
- **[LOW]** Export function still vulnerable to sensitive data leakage - The exportToOp function now warns about plaintext output but still prints secret values to console. If someone accidentally pipes output or runs in an insecure environment, secrets could still leak despite the warning. _(flagged by: adversary)_
- **[LOW]** Test mocking may not cover all real API edge cases - The test uses a simple https.request mock that may not accurately simulate all real-world API failure modes like network timeouts, partial responses, or malformed JSON that could occur in production. _(flagged by: adversary)_

---

PR #550, opened 2026-07-08
Feat/melissa signup boards sidebar copies

MelissaOneil:


swahlquist:

## Automated PR Review

| --- | --- | --- |
| Adversary (DeepSeek-V3.2, ZDR) | `BLOCK` | 6 |

**Senior-dev:** The PR is generally well-structured with good i18n coverage and reasonable security improvements. A few medium-confidence issues warrant attention before merge.

**Adversary:** Critical Google OAuth missing terms check blocks registration; multiple security and logic gaps in user board copying and sidebar resolution.

### Findings (10 unique)
- **[HIGH]** Google OAuth signup lacks terms check for stored link; registration silently blocked - google_signup_complete now checks link['terms_agree'] but google_start may not store it (terms_agree in config only for flow='register'). If user starts OAuth with flow='register' but config not stored/passed, the server returns error, but frontend only shows generic error, no restart guidance. _(flagged by: adversary)_
- **[HIGH]** Assign vocal flair home uses copy_board_to_library which may trigger unrestricted copy - assignVocalFlair84AsHome calls user.copy_board_to_library with user-supplied board id but lacks permission check that board is public/copyable, potentially allowing copying of private boards if frontend passes arbitrary board id. _(flagged by: adversary)_
- **[MED]** assignVocalFlair84AsHome silently swallows errors on success path - In setup.js and tour-board-picker.js the .catch only resets assigning_home_board but does NOT call modal.error. The old code did. When saveHomeBoard rejects (user.save fails), the spinner clears silently with no user feedback; onError is only called from copyBoardAndSaveHome, not from the saveHomeBo... _(flagged by: senior-dev)_
- **[MED]** New showEmailStep computed skips feature-flag gating - The new two-step email registration flow (method-chooser → dedicated email step) is user-visible UI restructuring with no feature flag wrapping showEmailStep or continue_with_email. AAC users are sensitive to unexpected UI changes; this should be behind a feature flag per the project policy. _(flagged by: senior-dev)_
- **[MED]** User.resolve_sidebar_boards_for issues N+1 Board queries per user - Board.find_all_by_path(system_keys.uniq) batches catalog lookups, but user.boards.where(parent_board_id: parent_ids) is a per-user query emitted every time sidebar_boards is called (on every board render). At scale this could degrade performance noticeably; at minimum it should be noted that this pa... _(flagged by: senior-dev)_
- **[MED]** User.sidebar_boards deduplication fails to deduplicate crisis vocabulary entries - dedupe_resolved_sidebar_boards only dedupes by slug for crisis-vocabulary, but default_sidebar_boards may include multiple entries with same slug but different keys (e.g., lingolinq/crisis-vocabulary, username/crisis-vocabulary) causing duplicates. _(flagged by: adversary)_
- **[MED]** Sidebar board resolution may expose private board keys via parent_board_id - resolve_sidebar_entry fetches board by key without checking user access; if user has copy of private board, sidebar may leak parent board key (system key) even if user shouldn't see it. _(flagged by: adversary)_
- **[LOW]** restart_google_signup clears age_attested before router transition completes - Setting age_attested: false then calling router.transitionTo resets the terms attestation synchronously, but if the transition is aborted (e.g. an unsaved model check) the UI is left in a half-reset state where googleSignupProfile is null and registrationStep is 'account' but age_attested is already... _(flagged by: senior-dev)_
- **[LOW]** Missing error handling in assignVocalFlair84AsHome may leave assigning_home_board true - If assignVocalFlair84AsHome fails, .catch sets assigning_home_board false, but if onSuccess not called and onError not defined, UI may stay in loading state. _(flagged by: adversary)_
- **[LOW]** Gemfile lock update includes websocket-driver but may miss transitive dependency audit - Added websocket-driver >=0.8.2 for CVE but lock update may not force re-resolution of transitive dependencies (actioncable), leaving older vulnerable versions if present. _(flagged by: adversary)_

---

PR #552, opened 2026-07-08
security/compliance: fix the two High audit findings (backtrace leak + missing consent AuditEvent)

swahlquist:
Remediates the two adversary-confirmed **High** findings from the 2026-07-08 audit refresh. Code + regression tests only; the findings register is left untouched (their status is Scot's to change).

## `LL-85038c0a7b` — debug_sync backtrace leak (`app/controllers/api/button_sets_controller.rb`)
The `buttonsets#generate` `debug_sync=1` rescue returned `{error: e.message, backtrace: e.backtrace}` to the client, exposing internal file paths and gem versions to any caller with board-view permission.
- **Fix:** drop the `backtrace` key (it is already written to `Rails.logger` two lines above); standardize on the `{error: message}` shape used elsewhere in the API. No behavior change beyond removing the leak.
- **Test:** asserts the error body has `error` but no `backtrace` key.

## `LL-47117a3443` — no immutable AuditEvent on parental consent (`app/controllers/parental_consents_controller.rb`)
COPPA accountability needs a tamper-evident record that verifiable parental consent was obtained. The grant path recorded consent in user settings but wrote no audit row.
- **Fix:** emit `AuditEvent.log_command` on a successful grant (`type`, `user_id`, `method`, IP, timestamp). It is immutable (`attr_readonly`) and fail-open, so it can never block a parent from completing consent; `data` is `secure_serialized`. Fires only on a new grant, not on idempotent re-visits.
- **Tests:** an AuditEvent is written on grant; a repeat visit does not double-log.

## Verification
Both affected specs pass locally (20 examples, 0 failures):
- `spec/controllers/parental_consents_controller_spec.rb`
- `spec/controllers/api/button_sets_controller_spec.rb`

## Register note (for Scot)
After this merges, `LL-47117a3443` and `LL-85038c0a7b` can move to **remediated-unverified** with this PR's SHA as closure evidence, pending your verification. I did not change their status — per the audit rules, only you triage.

https://claude.ai/code/session_0178kTWqKrbGNf6GsM2sd2b4

### DIFF
--- a/app/controllers/api/button_sets_controller.rb
+++ b/app/controllers/api/button_sets_controller.rb
@@ -80,7 +80,10 @@ class Api::ButtonSetsController < ApplicationController
         rescue => e
           Rails.logger.error "SYNC DEBUG ERROR: #{e.message}"
           Rails.logger.error e.backtrace.join("\n")
-          render json: {error: e.message, backtrace: e.backtrace}, status: 500
+          # The backtrace is already in the server log above; never return it to
+          # the client. Standardize on the {error: message} shape the rest of the
+          # API uses so this debug path stops leaking internal paths/gem versions.
+          render json: {error: e.message}, status: 500
           return # Stop execution here for sync mode
         end
       end
--- a/app/controllers/parental_consents_controller.rb
+++ b/app/controllers/parental_consents_controller.rb
@@ -11,19 +11,10 @@ class ParentalConsentsController < ApplicationController
     if user && c.is_a?(Hash) && c['parent_consent_granted_at'].present? && !user.coppa_parental_consent_pending?
       @already_granted = true
       @success = true
-    elsif user && user.grant_parental_consent!(token)
+    elsif user && user.grant_parental_consent!(token, ip: request.remote_ip, user_agent: request.headers['User-Agent'])
+      # grant_parental_consent! writes the immutable COPPA AuditEvent atomically
+      # with the consent grant (see User#grant_parental_consent!).
       @success = true
-      # COPPA accounting: record an immutable, tamper-evident audit row proving
-      # verifiable parental consent was obtained, for whom, when, and from where.
-      # log_command is fail-open (best-effort) so a persistence hiccup can never
-      # block the parent from completing consent; data is secure_serialized.
-      AuditEvent.log_command(user.global_id, {
-        'type' => 'parental_consent_granted',
-        'user_id' => user.global_id,
-        'method' => 'email_token_link',
-        'ip_address' => request.remote_ip,
-        'granted_at' => Time.now.utc.iso8601
-      })
       UserMailer.schedule_delivery(:confirm_registration, user.global_id)
       UserMailer.schedule_delivery(:new_user_registration, user.global_id)
       ExternalTracker.track_new_user(user)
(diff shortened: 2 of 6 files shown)


swahlquist:

## Automated PR Review

### Verdict: `COMMENT`

| --- | --- | --- |

**Senior-dev:** Both security fixes are correct and well-tested. One medium finding: the controller comment says AuditEvent uses fail-open semantics, but the model implementation is actually fail-closed (AuditEvent.create! inside the lock rolls back the grant on failure), which is a documentation contradiction that could mislead future maintainers about the reliability contract.

**Adversary:** Backtrace leak is fixed; COPPA audit logic introduces a potential deadlock and transactional inconsistency.

### Findings (4 unique)
- **[MED]** Controller comment contradicts model's fail-closed semantics - parental_consents_controller.rb line comment says 'fail-open, so it can never block a parent from completing consent', but User#grant_parental_consent! uses AuditEvent.create! inside with_lock, meaning an audit insert failure raises and rolls back the grant. The PR description also says 'fail-open'.... _(flagged by: senior-dev)_
- **[MED]** Deadlock risk in nested transaction with explicit lock - grant_parental_consent! uses with_lock(requires_new: true) inside a controller action likely already in a transaction, creating nested savepoints. Combined with AuditEvent.create! (which also creates a transaction), this can cause deadlocks under concurrency. _(flagged by: adversary)_
- **[LOW]** devices.each called outside the lock after with_lock block completes - devices.each(&:invalidate_cached_keys) is called after the with_lock block returns. If the lock block sets res=true but the user is reloaded by another thread before cache invalidation, stale keys may persist briefly. Minor race, low impact, but the old code ran invalidation inside the implicit succ... _(flagged by: senior-dev)_
- **[LOW]** Potential PII leakage into AuditEvent.data - The audit event stores raw IP and user agent from the request without sanitization. While stored securely, it increases PII scope beyond COPPA's required record-keeping. _(flagged by: adversary)_

---

swahlquist:
Addressed the review in `eb8f6dbee`. Both points were valid — the finding itself calls for mirroring `grant_ai_consent!` and for capturing user-agent + policy version, which the first pass didn't do.

**High (atomicity + concurrency):** refactored `User#grant_parental_consent!` to mirror the canonical `grant_ai_consent!` pattern — the settings write, Privacy Policy acknowledgment, `save!`, and `AuditEvent.create!` now run inside one `with_lock(requires_new: true)` (SELECT FOR UPDATE + SAVEPOINT). So:
- an audit-insert failure rolls back the consent grant, **including the token invalidation**, so the parent just retries the same link (no consent-without-record, and no lockout);
- concurrent token requests serialize — the second reloads, sees the committed grant, and no-ops.

Moved the fail-open controller `log_command` out; request context (`ip`, `user_agent`) is passed into the model.

**Medium (payload):** the immutable record now carries `method`, `ip`, `user_agent`, `privacy_policy_version`, the exact persisted `granted_at`, and a stable `record_id` + `event_type`.

**Tests:** added a model spec proving (a) one immutable event on grant carrying the persisted timestamp, and (b) full rollback with the token retained when `AuditEvent.create!` raises; controller specs updated to assert the richer payload and the timestamp match. Local run: 20 controller + consent model examples, 0 failures.

Backtrace fix left unchanged per your approval.

PR #556, opened 2026-07-09
compliance(eu): jurisdiction primitive + jurisdiction-aware COPPA/GDPR age gate

swahlquist:
## EU launch readiness (Poland): the two engineering gates before the first EU user signs up

Poland is sold through a distributor we host, which makes GDPR active. GDPR Art. 8 sets the digital-consent age at up to 16 (Poland uses 16); today the registration age gate assumes the US COPPA age-13 threshold. This PR adds a jurisdiction primitive and makes the consent-age threshold jurisdiction-aware, behind a feature flag that is OFF by default.

Two atomic commits, each with RSpec.

### Commit 1 — `LingoLinq::Jurisdiction` primitive
`app/models/lingo_linq/jurisdiction.rb`, exposing `LingoLinq::Jurisdiction.eu?(signal)` and `.country_code(signal)`.

- Derives an ISO 3166-1 alpha-2 country from signals that **already exist** on a user or request: a locale (`pl-PL`), an explicit country (`PL`), an `Accept-Language` header, a User-like object (`settings` country/region/locale), or a Hash. **No IP geolocation, no new gem.**
- Ambiguity policy: an explicit `country`/`region` field is trusted; a bare lowercase language subtag (`pl`, `es`, `fr`) is treated as **unknown** so non-EU speakers of EU languages are never misclassified as EU. Unknown ⇒ `eu? == false` ⇒ today's behavior.
- EU-27 set; EEA-not-EU (IS/LI/NO) and the UK are intentionally excluded.
- File is namespaced under `lingo_linq/` because Zeitwerk requires the `LingoLinq::Jurisdiction` constant to live there; a bare `app/models/jurisdiction.rb` defining that constant breaks eager loading. `rails zeitwerk:check` passes.

### Commit 2 — jurisdiction-aware age gate
The age threshold lives **only** on the frontend (`register.js#_classifyCommunicatorAge`, `getFullYear() - 13`); the backend trusts the boolean `coppa_under_13` and never sees the birthdate, so its consent-triggering gate needs **no** change. The EU→16 value is computed server-side and delivered to the anonymous register UI via `domain_settings`.

- New frontend feature flag **`eu_consent_age`** in `AVAILABLE_FRONTEND_FEATURES` only ⇒ **OFF for everyone by default**. `FeatureFlags.eu_consent_age_enabled?` mirrors how anonymous registration reads flags (`ENABLED_FRONTEND_FEATURES`).
- `JsonApi::Json.coppa_consent_age(signal)` ⇒ 16 (EU) / 13 (else) via the primitive.
- `ApplicationController#coppa_consent_age_injection` (a `helper_method`) injects `coppa_consent_age` **only when the flag is on**, using a per-request jurisdiction signal (org/domain country/region/locale → `?locale=` → `Accept-Language`). Merged into a **fresh** copy of the settings hash in `application.html.erb` and `api/integrations#domain_settings` so the **cached per-host blob is never mutated**.
- `register.js`: new `coppaConsentAge` computed (flag- and value-guarded, 13 fallback) used as the classifier cutoff. The `under_13`/`over_13` labels are kept as semantic "under/over threshold", so nothing downstream ripples.

### Flag OFF = identical to today
Flag in `AVAILABLE_` only ⇒ backend injects nothing (byte-identical `domain_settings`) **and** the frontend `coppaConsentAge` returns 13, so the cutoff is exactly the original `getFullYear() - 13`.

### Verification
- `DB_USER=scotw RAILS_ENV=test bundle exec rspec` on the new/changed specs: **73 examples, 0 failures**. Covers EU (Poland + others), non-EU (US/UK/NO/CA), ambiguous/unknown, User/Hash signals; `coppa_consent_age` EU/non-EU/unknown; flag registration + gate; and `domain_settings` injection present only when ON (16 EU / 13 non-EU), **absent when OFF**, with **no cache mutation**.
- `rails zeitwerk:check` passes.
- Frontend JS syntax-checked under Node 20. Ember unit specs do not execute here (AMD loader defect), so the register step-machine still needs a **manual browser drive**: `foreman start`, temporarily add `eu_consent_age` to `ENABLED_FRONTEND_FEATURES`, set browser `Accept-Language` to `pl-PL` ⇒ under-16 routes to the parental-consent step; `en-US` or flag OFF ⇒ under-13. (Build check in progress.)

### Scope notes
- Additive and reversible: no migrations, flag OFF by default.
- Stayed out of `docs/legal/**` and the EU AI Act Article 50 / AI-consent code (separate thread). This PR creates the `Jurisdiction` primitive the Art. 50 thread will reuse, and must merge first to avoid a collision on `jurisdiction.rb` + `feature_flags.rb`.

### Known follow-up (not shipped, keeps flag-OFF byte-identical)
Under-threshold copy `coppa_parent_email_required` says "under 13" literally. Correct while the flag is OFF (the gate IS 13). Making it age-dynamic needs an i18n regeneration pass (`i18n_generator.rb`) and is deliberately deferred.

https://claude.ai/code/session_0165vszi8WGZsg5hQaRxQFT2

---

## Dual-review outcome (Codex + Claude adversary)

Codex senior-dev pass: no findings. Claude adversary pass: 2 High, 2 Medium, 2 Low (all verified). Both confirm the core guarantees hold: **flag-OFF is byte-identical** and **the cached per-host domain blob is never poisoned**. The dark merge (flag OFF) is regression-free.

Addressed in commit `dce829ea3` (in-scope for the dark merge): #2 removed dead jurisdiction-signal branches + honest best-effort labeling; #3 documented the two-flag coupling; #4 added specs for the layout data source; #5 fixed the BCP-47 region-subtag parse and documented the ISO-country contract; #6 restored null-safety.

**Documented follow-ups (do NOT enable `eu_consent_age` in production as a GDPR Art. 8 control until these land):**
- **#1 (High):** the age gate trusts a client-supplied `coppa_under_13` boolean; the backend never recomputes age (the birthdate is not sent). An EU 14/15-year-old can bypass consent by sending `coppa_under_13=false`. This is pre-existing (same trust model as COPPA-13) and was explicitly out of scope here (reuse the existing consent gates, do not reimplement). Server-side age enforcement is required before relying on this as an enforceable control.
- **Jurisdiction detection is best-effort browser-locale only** (bare `pl` resolves to age 13). An authoritative org-configured EU-host country should be wired, and the region-present rate measured, before this is represented to EU customers/auditors as a compliance control.

As shipped (flag OFF), this is safe infrastructure; as a flag-ON control it is a more-protective-when-it-fires UI nudge, not yet an enforceable consent gate.

### DIFF
--- a/app/controllers/api/integrations_controller.rb
+++ b/app/controllers/api/integrations_controller.rb
@@ -75,7 +75,13 @@ class Api::IntegrationsController < ApplicationController
   end
 
   def domain_settings
-    render json: @domain_overrides.to_json
+    # Merge the per-request jurisdiction-aware consent age into a FRESH copy so
+    # the cached per-host blob (@domain_overrides) is never mutated. With the
+    # eu_consent_age feature OFF the injection is {}, so the payload is
+    # byte-identical to today.
+    overrides = @domain_overrides.dup
+    overrides['settings'] = (overrides['settings'] || {}).merge(coppa_consent_age_injection)
+    render json: overrides.to_json
   end
 
   def focus_usage
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -35,6 +35,28 @@ class ApplicationController < ActionController::Base
     true
   end
 
+  # EU launch (GDPR Art. 8): per-request jurisdiction-aware parental-consent age,
+  # delivered to the anonymous registration UI via domain_settings. Returns {}
+  # unless the eu_consent_age feature is enabled, so with the flag OFF the
+  # injected domain_settings are byte-identical to today. Callers MUST merge this
+  # into a fresh copy of the settings hash (never mutate @domain_overrides, which
+  # is the cached per-host blob from JsonApi::Json.load_domain).
+  def coppa_consent_age_injection
+    return {} unless FeatureFlags.eu_consent_age_enabled?
+    { 'coppa_consent_age' => JsonApi::Json.coppa_consent_age(jurisdiction_signal_for_request) }
+  end
+  helper_method :coppa_consent_age_injection
+
+  # Best jurisdiction signal available for THIS request, using only signals that
+  # already exist (no IP geolocation). Priority: a configured org/domain country
+  # or region, then locale, then an explicit ?locale= param, then the browser's
+  # Accept-Language header. LingoLinq::Jurisdiction resolves any of these.
+  def jurisdiction_signal_for_request
+    ds = (@domain_overrides && @domain_overrides['settings']) || {}
+    ds['country'] || ds['region'] || ds['locale'] || ds['default_locale'] ||
+      params[:locale].presence || request.headers['Accept-Language'].presence
+  end
+
   def log_api_call
     time = @time ? (Time.now - @time) : nil
     ApiCall.log(@token, @api_user, request, response, time)
--- a/app/frontend/app/controllers/register.js
+++ b/app/frontend/app/controllers/register.js
@@ -152,6 +152,19 @@ export default Controller.extend({
     return !!(ds && ds.coppa_parental_consent);
   }),
   coppa_age_group: null,
+  // EU launch (GDPR Art. 8): the age below which registration requires
+  // verifiable parental consent. Gated by the eu_consent_age feature flag; with
+  // the flag OFF this is always 13 and the age gate is identical to today. When
+  // ON, the EU value (16) is computed server-side (LingoLinq::Jurisdiction) and
+  // delivered through domain_settings, so the single source of EU truth stays
+  // on the backend and this stays a dumb number consumer.
+  coppaConsentAge: computed('appState.feature_flags.eu_consent_age', 'appState.domain_settings.coppa_consent_age', function() {
+    var fallback = 13;
+    if(!this.get('appState.feature_flags.eu_consent_age')) { return fallback; }
+    var age = parseInt(this.get('appState.domain_settings.coppa_consent_age'), 10);
+    if(!age || age < 13 || age > 18) { return fallback; }
+    return age;
+  }),
   parent_consent_email: '',
   coppaParentEmailMissing: computed('triedToSave', 'coppa_age_group', 'parent_consent_email', function() {
     if(this.get('coppa_age_group') !== 'under_13') { return false; }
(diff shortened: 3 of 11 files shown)


swahlquist:

## Automated PR Review

### Verdict: `COMMENT`

| --- | --- | --- |

**Senior-dev:** The implementation is well-structured, feature-flagged, and cache-safe, but two issues need attention before merge: a shallow .dup in the integrations controller fails to isolate a nested settings hash mutation, and the application.html.erb path mutates the cached @​domain_overrides['settings'] hash in-place for requests where coppa_consent_age_injection is non-empty.

**Adversary:** Jurisdiction detection via Accept-Language header is trivial to forge; backend uses age-derived boolean but user-visible labels unchanged; hardcoded age validation range may be insufficient.

### Findings (5 unique)
- **[HIGH]** application.html.erb mutates cached @​domain_overrides['settings'] in-place - (@​domain_overrides['settings'] || {}).merge(coppa_consent_age_injection) returns a new hash but the original @​domain_overrides['settings'] object is untouched; however the caller is directly referencing the cached blob's nested hash. The PR description says callers MUST merge into a fresh copy, bu... _(flagged by: senior-dev)_
- **[MED]** jurisdiction_signal_for_request: user-supplied ?locale= param trusted as a country signal - params[:locale].presence is included in the jurisdiction signal priority chain. A user can craft a request with ?locale=PL to force the EU age gate (16) even from a US context. While this only changes which consent age the anonymous register UI shows (the backend coppa_under_13 boolean is still user... _(flagged by: senior-dev)_
- **[MED]** Jurisdiction determined by client-controlled Accept-Language header - Jurisdiction signal includes Accept-Language header which is client-controlled. An attacker can set Accept-Language to an EU locale to potentially affect age-gating, though frontend validation remains. _(flagged by: adversary)_
- **[MED]** Age validation range does not align with known consent ages - Frontend coppaConsentAge validation (13-18) excludes possible future jurisdictions with ages 18 (e.g., some countries set <13). This may cause validation to incorrectly fallback. _(flagged by: adversary)_
- **[LOW]** coppaConsentAge computed: parseInt with radix 10 on a server-injected integer is safe, but the range guard silently rejects age 13 - The guard age  18 with an explicit 13-allowed lower bound (i.e., age < 13 should be age < 1 or removed) would make intent clearer; currently a server-returned 13 is silently treated as invalid and falls back to 13, which is correct but accidental. _(flagged by: senior-dev)_

---

MelissaOneil (approved):

PR #567, opened 2026-07-09
fix(frontend): bind password-field visibility toggle for Ember 5 {{on}}

MelissaOneil:
The Ember 5.12 action migration left togglePasswordVisibility unbound, so clicking the show/hide control threw this.toggleProperty is not a function.

### DIFF
--- a/app/frontend/app/components/password-field.hbs
+++ b/app/frontend/app/components/password-field.hbs
@@ -1,6 +1,6 @@
 <div class="password-field">
   <Input @type={{this.inputType}} id={{this.id}} class={{this.class}} @value={{this.value}} placeholder={{this.placeholder}} autocomplete={{this.autocomplete}} />
-  <button type="button" class="password-field__toggle" aria-pressed={{this.showPassword}} aria-label={{this.toggleAriaLabel}} {{on "click" this.togglePasswordVisibility}}>
+  <button type="button" class="password-field__toggle" aria-pressed={{this.showPassword}} aria-label={{this.toggleAriaLabel}} {{on "click" (this.ctrlAction "togglePasswordVisibility")}}>
     {{#if this.showPassword}}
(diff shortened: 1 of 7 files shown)


swahlquist:

## Automated PR Review

### Verdict: `COMMENT`

| --- | --- | --- |

**Senior-dev:** The fix correctly routes the click through actions but the ctrlAction factory unconditionally calls evt.preventDefault() on every native click, which will suppress expected browser behaviour (e.g. form submission from sibling elements if the pattern spreads). One low-severity observation about the guard logic is also worth noting.

**Adversary:** The diff introduces a potential memory leak in the ctrlAction helper by creating a new closure on every template invocation without cleanup; also modifies test utilities in ways that could affect unrelated tests.

### Findings (4 unique)
- **[MED]** ctrlAction strips the event and calls preventDefault on every click - In init(), ctrlAction pops the last argument and calls evt.preventDefault() whenever it looks like a DOM event. A password-field toggle button should never need preventDefault (it is type='button'), but codifying this side-effect in a shared helper means any future caller that does need the event (e... _(flagged by: senior-dev)_
- **[MED]** ctrlAction may cause memory leaks - The ctrlAction method returns a new closure each time it is called, but no cleanup is performed when the component is destroyed. In templates where the action is bound to {{on}}, a new closure is created on every render, which may accumulate over time. _(flagged by: adversary)_
- **[LOW]** ctrlAction helper assigned in init() creates a per-instance function closure - Assigning ctrlAction as a plain function on the instance in init() (not on the prototype) allocates a new closure per component instance. For a rarely-used component this is fine, but the pattern contradicts the LEARNINGS.md note ('do not assign per-handler closures on the instance in init()') added... _(flagged by: senior-dev)_
- **[LOW]** Test helper modifications may affect other tests - Changes to sync-test-cleanup.js and persistence-sync-test.js modify global stub behavior and persistence references, potentially introducing side effects for other tests that rely on these utilities. _(flagged by: adversary)_

---

PR #568, opened 2026-07-10
security(user): expiring lesson_share_token for share links (LL-90045bb29c option b)

swahlquist:
## What

Remediation **option (b)** of finding **LL-90045bb29c** (high): stop embedding the permanent, non-revocable `User#user_token` in navigable lesson/board share URLs (path segments that leak into browser history, access logs, `Referer` headers, and forwarded emails as a forever-valid bearer credential).

Introduces `User#lesson_share_token` / `User.find_by_lesson_share_token`, a purpose-scoped **expiring** token mirroring the proven `protected_image_token` pattern (3-part `id-expires-sig`, constant-time compare, 30-day lifespan, distinct verifier purpose).

Does **not** close the finding. Options (a) shipped in #563; (c) (`user_token` rotation + embed-frame surface) remains. Closure is Scot's attestation.

## Design

- **Mint kill-switch:** `FeatureFlags.expiring_lesson_share_tokens_enabled?` (`EXPIRING_LESSON_SHARE_TOKENS` env var, default on) gates only the *mint*, so construction reverts to the legacy token in one switch, no deploy. The **finder always accepts both** new and legacy formats, so flipping the switch never breaks an issued link.
- **Legacy fallback** (2-part `user_token`) still resolves, logged as `[lesson_share_legacy_token]` so the residual can be measured before sunset (option (c)/LL-310b464be4).
- **Id stability:** the lesson serializer **echoes** the requested token (`extra_user_token`) into the response id rather than minting a fresh one, so Ember Data `findRecord` resolves the primary record (fixes the Codex P1 below).

## Construction sites swapped
- Lesson serializer composite id (`lib/json_api/lesson.rb`)
- User serializer (`lib/json_api/user.rb`, new `lesson_share_token` attr; `user_token` retained for option (c))
- **Both lesson email templates** (`app/views/user_mailer/lesson_assigned.*`) — highest-risk surface, *missed by the original scope, caught during build*
- 4 frontend `window.open` sites (authenticated-view, organization/lessons, organization/room, user/lessons)
- Accept points (`lessons#show`, `lessons#complete`, `boards#lesson`) now use `find_by_lesson_share_token`

## Reviews (dual-reviewer, both ran)
- **Codex senior-dev:** found P1 Ember-Data findRecord id-instability (serializer minted a fresh token, breaking the requested id). **Fixed** in `b94abd3` (echo the requested token; regression-guarded by `#complete` specs asserting id equality).
- **Claude adversary:** verdict *Ship with conditions*. Crypto/auth core clean — no forgeable widening, no injection, no kill-switch asymmetry. Open items for triage:
  - **[Medium]** Emailed links 404 after 30 days with no "expired link" UX (product decision on TTL + expired-link experience).
  - **[Low]** Ember id-mismatch on an expired `#show` token (same nil-user path that already existed for malformed tokens; dead-link UX, not a leak).
  - **[Medium]** Pre-existing leaked permanent tokens stay valid via the legacy fallback — **by design**; finding stays open for option (c).

## Tests
Model: mint (flag on/off), finder valid/expired/tampered-sig/tampered-expiry/non-numeric/legacy-fallback/constant-time, purpose-distinctness. Controller: `#show`/`#complete` for both token formats + id-stability guard. Mailer: link resolves. All green in isolated runs.

## Notes
Behavior-preserving hardening; no new user-facing strings (no i18n/em-dash concern); no PII in diff. Follow-ups for triage: expired-link UX (frontend route + i18n) and the option (b) email-TTL product decision.

https://claude.ai/code/session_013fYhA53ubbtwPnpbn8qani

### DIFF
--- a/app/controllers/api/lessons_controller.rb
+++ b/app/controllers/api/lessons_controller.rb
@@ -76,7 +76,7 @@ class Api::LessonsController < ApplicationController
     return unless exists?(lesson, lesson_id)
     return unless lesson.nonce == lesson_code || allowed?(lesson, 'view')
     user = User.find_by_lesson_share_token(user_token)
-    render json: JsonApi::Lesson.as_json(lesson, {wrapper: true, permissions: @api_user, extra_user: user})
+    render json: JsonApi::Lesson.as_json(lesson, {wrapper: true, permissions: @api_user, extra_user: user, extra_user_token: user_token})
   end
 
   def update
@@ -101,7 +101,7 @@ class Api::LessonsController < ApplicationController
 
     Lesson.complete(lesson, user, params['rating'].to_i, nil, params['duration'].to_i)
 
-    render json: JsonApi::Lesson.as_json(lesson, {wrapper: true, extra_user: user, permissions: user})
+    render json: JsonApi::Lesson.as_json(lesson, {wrapper: true, extra_user: user, permissions: user, extra_user_token: user_token})
   end
 
   def assign
--- a/app/controllers/api/lessons_controller.rb
+++ b/app/controllers/api/lessons_controller.rb
@@ -75,7 +75,7 @@ class Api::LessonsController < ApplicationController
     lesson = Lesson.find_by_path(lesson_id)
     return unless exists?(lesson, lesson_id)
     return unless lesson.nonce == lesson_code || allowed?(lesson, 'view')
-    user = User.find_by_token(user_token)
+    user = User.find_by_lesson_share_token(user_token)
     render json: JsonApi::Lesson.as_json(lesson, {wrapper: true, permissions: @api_user, extra_user: user})
   end
 
@@ -96,7 +96,7 @@ class Api::LessonsController < ApplicationController
     if lesson.settings['nonce'] != lesson_code
       return allowed?(lesson, 'never_allow')
     end
-    user = User.find_by_token(user_token)
+    user = User.find_by_lesson_share_token(user_token)
     return unless exists?(user, user_token)
 
     Lesson.complete(lesson, user, params['rating'].to_i, nil, params['duration'].to_i)
--- a/app/controllers/boards_controller.rb
+++ b/app/controllers/boards_controller.rb
@@ -98,7 +98,7 @@ class BoardsController < ApplicationController
   def lesson
     @lesson = Lesson.find_by_path(params['lesson_id'])
     @lesson = nil unless @lesson.nonce == params['lesson_code']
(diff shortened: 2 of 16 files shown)


swahlquist:

## Automated PR Review

### Verdict: `COMMENT`

| --- | --- | --- |

**Senior-dev:** The security hardening is well-designed and the crypto core is sound. One concrete issue: the token split on /-/ is too broad — global_id values contain hyphens, so splitting a 3-part expiring token (global_id-expires-sig) on every hyphen will produce more than 3 parts and the finder silently falls through to the legacy path for any user whose global_id contains a hyphen.

**Adversary:** Security improvement introduces expiring tokens but retains legacy fallback permanently, allowing previously leaked credentials to remain valid indefinitely. Potential for expired token id mismatches in certain error paths.

### Findings (5 unique)
- **[HIGH]** split(/-/) on global_id containing hyphens causes expiring tokens to always fall back to legacy - global_id values like '1_2-abc123' or any multi-segment id with hyphens produce >3 parts when split on /-/, so parts.length == 3 is false and find_by_lesson_share_token silently accepts the permanent user_token instead. The mint produces the correct 3-segment string, but the finder never validates i... _(flagged by: senior-dev)_
- **[MED]** lesson_share_token minted per-call: each user serializer response emits a fresh token - User#lesson_share_token is called in build_json unconditionally (json_api/user.rb line). Each API response mints a different expires_at/sig pair. The lesson id-stability fix (echo extra_user_token) only covers the lesson serializer path; the user serializer's lesson_share_token attribute will drift ... _(flagged by: senior-dev)_
- **[MED]** Legacy token fallback persists indefinitely - The find_by_lesson_share_token method permanently accepts legacy permanent tokens, meaning previously leaked user_tokens remain valid forever. This undermines the security benefit of expiring tokens and creates permanent residual risk. _(flagged by: adversary)_
- **[LOW]** Legacy-fallback log line fires even for nil user (guard is if user) - The log is correctly guarded, but if find_by_token returns a user and the log fires, the global_id is exposed in Rails logs for every legacy link click. Confirm log level (info vs debug) matches your log-shipping/retention policy for PII-adjacent data (user identifiers). _(flagged by: senior-dev)_
- **[LOW]** Potential expired token id mismatch in lesson#show - When an expired token is used in lessons#show, the controller returns nil user but still renders the lesson JSON with extra_user: nil. This could cause Ember Data id mismatch issues similar to those fixed in complete action. _(flagged by: adversary)_

---

PR #571, opened 2026-07-10
fix(eval): persist tiered-eval log data through the real create path

swahlquist:
## Summary
- `EvalSession#persist` posts `{log_type: 'eval', data: {eval_mode, ai_narrative, recommendation, sett, slp_notes, ...}}` to `/api/v1/logs`, but `LogSession#process_params` only ever handled the legacy flat `events`/`note`/`assessment`/singular-`eval` shape — the real create path fell into the events-only background-job branch and silently returned an empty, unpersisted log. Independent of any other work, this means the Comprehensive Eval "save report" feature has likely never actually persisted `eval_mode`/`recommendation`/`ai_narrative`/`sett`/`slp_notes` for any real user via the live API.
- Discovered while pre-flighting the EU AI Act Art.50(2) eval-narration content-marking work (separate branch), which depends on this data actually reaching the database.
- Fix is minimal and targeted:
  - `process_as_follow_on`: route the tiered-eval shape synchronously (mirrors the existing note/assessment/journal fast paths) instead of the events-required stash/background-job path.
  - `process_params`: actually copy `eval_mode`/`intake`/`recommendation`/`events`/`duration_s`/`slp_notes`/`sett`/`ai_narrative` into `self.data` and set `log_type`.
  - `generate_defaults`: its own `log_type` dispatch cascade only recognized the legacy `data['eval']` key and was silently overwriting `log_type` back to `'session'` for the new shape.
  - `json_api/log.rb#extra_includes`: guarded a legacy-only stats assignment that crashed (`NoMethodError` on nil) for the new shape's absent `'evaluation'` key — found via the new regression spec actually failing.

## Test plan
- [x] Reproduced the bug live pre-fix (real payload through `LogSession.process_as_follow_on` returned an empty, unpersisted log)
- [x] Confirmed live post-fix (full `data` hash round-trips through create -> reload, `log_type == 'eval'`)
- [x] New regression spec: `spec/controllers/api/logs_controller_spec.rb` — POSTs the real payload shape, asserts persistence AND round-trip through `JsonApi::Log`'s `tiered_eval` view
- [x] Full regression run: `log_session_spec.rb`, `log_session_board_spec.rb`, `logs_controller_spec.rb`, `json_api/log_spec.rb` — 283 examples, 0 failures, 5 pre-existing unrelated pending
- [x] `eval_narrator_spec.rb`, `eval_sessions_controller_spec.rb`, `eval_recommend_spec.rb` — 55 examples, 0 failures

https://claude.ai/code/session_01HeBiAwADoxVxc92QdcVtvZ

### DIFF
--- a/app/models/log_session.rb
+++ b/app/models/log_session.rb
@@ -281,19 +281,34 @@ class LogSession < ApplicationRecord
       self.data['duration'] = (self.ended_at - self.started_at).to_i rescue nil
     elsif self.data['eval_mode']
       # Tiered eval report (Quick Screen / Targeted / Comprehensive) -- distinct from the
-      # legacy singular data['eval'] shape above. Override (not ||=) rather than trust
-      # started_at/ended_at from the events-timestamp derivation earlier in this method:
-      # that code is designed for realtime button-press 'session' logs keyed on
-      # event['timestamp'] in SECONDS, and would silently corrupt these fields with a
-      # bogus far-future date if a tiered-eval event ever carried a 'timestamp' key in JS
-      # milliseconds. EvalSession#recordEvent currently keys events on 'ts', not
-      # 'timestamp' (app/frontend/app/utils/eval_session.js), so that derivation is a
-      # no-op today -- but this branch must not depend on that key-naming accident for
-      # correctness, so it always derives from the authoritative duration_s instead.
+      # legacy singular data['eval'] shape above.
+      #
+      # started_at/ended_at must be derived idempotently (same result every save), for two
+      # reasons that rule out both a plain ||= and a plain "only set once" guard:
+      #   1. The events-timestamp derivation earlier in this method is designed for
+      #      realtime button-press 'session' logs keyed on event['timestamp'] in SECONDS,
+      #      and runs UNCONDITIONALLY on every save, before this branch. For tiered eval it
+      #      always resolves to nil (first/last['timestamp'] is nil, since
+      #      EvalSession#recordEvent keys real events on 'ts', not 'timestamp' --
+      #      app/frontend/app/utils/eval_session.js) -- so a guard like `if new_record?`
+      #      here is not enough: it would leave started_at/ended_at wiped to nil on every
+      #      LATER save (an admin toggling `highlighted`, attaching a note, any unrelated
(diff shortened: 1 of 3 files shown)


swahlquist:

## Automated PR Review

### Verdict: `COMMENT`

| --- | --- | --- |

**Senior-dev:** The fix is correct and well-reasoned, but two issues in the test file need attention before merging: a reference to an undefined constant REDACTED_ID that will cause a NameError at test runtime, and a sleep 1 in the idempotency test that will make the suite fragile and slow on loaded CI runners.

**Adversary:** The fix introduces a race condition in timestamp generation and has PII exposure in logs, but correctly addresses the core persistence issue.

### Findings (5 unique)
- **[HIGH]** Undefined constant REDACTED_ID in spec causes NameError at runtime - Line (Time.now.to_i * 1000) - [REDACTED_ID] references REDACTED_ID as a bare constant. Ruby will raise NameError when this spec runs, causing the test to fail rather than passing. Replace with a literal integer (e.g. 30_000). _(flagged by: senior-dev)_
- **[MED]** sleep 1 in idempotency spec is fragile on loaded CI runners - The idempotency test relies on a real 1-second wall-clock sleep to detect timestamp drift. Under CPU contention the save after sleep may still fall within be_within(1.minute) but the test intent — exact equality of started_at/ended_at — doesn't need sleep at all; the anchor is data['completed_at'], ... _(flagged by: senior-dev)_
- **[MED]** Race condition in completed_at timestamp generation - self.data['completed_at'] ||= Time.now.to_i in generate_defaults is not atomic - concurrent creation requests for the same eval could see different timestamps, leading to inconsistent started_at/ended_at across replicas. _(flagged by: adversary)_
- **[MED]** PII exposure in production logs - Rails.logger.warn logs full eval payload including slp_notes, ai_narrative, and sett form data (potentially containing FERPA/HIPAA sensitive info) to production logs without redaction. _(flagged by: adversary)_
- **[LOW]** Incomplete data validation for eval_mode - process_params accepts any eval_mode string value without validation against known modes ('quick', 'targeted', 'comprehensive'), potentially allowing invalid modes into persistence. _(flagged by: adversary)_

---

PR #578, opened 2026-07-11
ci: make security-scan blocking, least-priv permissions on secrets-sync

swahlquist:
## Summary
Salvaged from the stale `open-findings-quick-wins` branch (cut 2026-06-22, never PR'd). Everything else in that branch has since shipped a different way and is dropped here:
- The license `external_reference` gate-by-permission approach is superseded — staging fixed that finding (LL-55baae6d40) on 2026-07-03 by removing the field entirely (#517/#519), not gating it.
- The general user-creation audit logging is superseded — staging already ships this via #528 (LL-d35cbdb313), with a more complete implementation than this branch's version.

What's left, and genuinely still missing from current staging:
- `security-scan` job drops `continue-on-error: true` — Brakeman/npm audit failures now actually block CI instead of being purely advisory.
- `sync-render-secrets.yml` gets an explicit `permissions: contents: read` (least-privilege default; the workflow has no need for write access).

## Test plan
- [x] Both modified workflow YAML files parse correctly
- [ ] CI green

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -138,7 +138,6 @@ jobs:
 
   security-scan:
     runs-on: ubuntu-latest
-    continue-on-error: true
 
     steps:
       - name: Checkout repository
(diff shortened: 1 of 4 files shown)


swahlquist:

## Automated PR Review

### Verdict: `APPROVE`

| --- | --- | --- |

**Senior-dev:** Both changes are correct and safe: removing continue-on-error: true from the job level makes security scans properly blocking while the per-step continue-on-error: true on npm audit is intentionally preserved, and adding permissions: contents: read follows least-privilege best practice for a secrets-sync workflow that needs no write access.

**Adversary:** Security scan now blocks on failure, and secrets sync uses read-only permissions, improving security posture.

### Findings (0 unique)
_No findings._

---

PR #580, opened 2026-07-11
Graceful expired-link UX for lesson_share_token (LL-90045bb29c option b follow-up)

swahlquist:
## Summary

Follow-up to PR #568 (LL-90045bb29c option (b), already on staging), filed from Claude's adversary review of that PR. When a `lesson_share_token` fails to resolve (expired, malformed, or wrong-signature), the app previously showed a bare `/404` on a fresh navigation or an unhandled Ember error on a client-side transition. This adds an explicit, reason-agnostic "this link has expired, request a new one" experience on both paths.

- `app/controllers/boards_controller.rb#lesson`: 404 now depends only on whether the lesson exists and its nonce matches, not on whether the share token resolved. A valid lesson with an unresolved token boots the Ember shell instead of 404ing, so the frontend can show the expired-link panel.
- `app/frontend/app/routes/lesson.js`: hardened to handle both possible Ember Data outcomes for an unresolved token (resolves-with-no-user, the empirically observed path; rejects, a defensive fallback), converting either into a `link_expired` state — while still letting a genuine not-found (missing/nonce-mismatched lesson) or a transient 5xx/offline error bubble to the normal error path, not the expired-link message.
- New i18n strings (double-quoted, via `{{t}}`), en.json + es.json placeholder.
- `spec/controllers/boards_controller_spec.rb` + `app/frontend/tests/acceptance/lesson_expired_test.js`: 5 new backend examples, 8 new frontend tests.

**UX-06 runtime finding:** Ember Data 5.3.8 (Ember 5.12) *resolves* (does not reject) `findRecord` for the id-mismatched, user-less payload an unresolved token produces — confirmed empirically via a real store/serializer test, not assumed. Both outcomes are handled regardless.

**Locked/untouched (do not relitigate):** token lifespan (30d), `FeatureFlags.expiring_lesson_share_tokens_enabled?`, `User.find_by_lesson_share_token`'s dual-format accept-side logic, `lesson_assigned.*` mailer views. No feature flag added for this change (explicit decision — error-state polish on an already-broken path, not a new capability). Finding LL-90045bb29c stays **open**; closure is Scot's attestation, not part of this PR.

## Review history

1. Plan reviewed by `gsd-plan-checker` (goal-backward verification): PASS, no blockers.
2. Plan adversary-reviewed (auth-adjacent): ship-with-conditions, all 3 conditions applied before execution.
3. Diff reviewed by Codex (senior-dev pass): 1 finding — a genuinely-missing lesson was being mislabeled "link expired" instead of a real not-found. Fixed and tested.
4. Diff adversary-reviewed a second time: 2 more findings on top of the Codex fix — (a) the Codex fix's own test covered a synthetic error shape this app doesn't actually produce in production (fixed by adding a test for the real `fakeXHR` shape), (b) 5xx/offline errors were still being silently relabeled "link expired" (fixed with a matching transient-error check + test).

## Test plan

- [x] `bundle exec rspec spec/controllers/boards_controller_spec.rb -e lesson` — 26/26 pass
- [x] `ember test --filter="lesson expired"` — 8/8 pass
- [x] Full `ember test` regression — 1708/1708 pass, 0 fail
- [x] `eslint` clean on all touched frontend files


swahlquist:

## Automated PR Review

### Verdict: `COMMENT`

| --- | --- | --- |

**Senior-dev:** The diff's code changes are well-structured and the feature logic is sound, but three concrete issues are worth flagging before merge: a feature-flag bypass on a new user-facing UI state, a plain-object this hazard in a promise chain, and a missing catch on transient errors that may still silently convert 5xx/offline failures into the expired-link message.

**Adversary:** The PR introduces logic to handle expired lesson share tokens gracefully but contains a subtle race condition and potential for data isolation issues between district accounts.

### Findings (5 unique)
- **[MED]** New link-expired UI state ships without a feature flag - PR description explicitly documents the decision to skip a feature flag ('error-state polish on an already-broken path'). However, the expired-link panel is a net-new UI surface visible to AAC users on both the fresh-navigation and SPA-transition paths. Per CLAUDE.md's hard rule, all new user-facing... _(flagged by: senior-dev)_
- **[MED]** Plain-object this hazard in promise executor callbacks in routes/lesson.js - The plan mandates 'use var _this = this inside the promise callbacks; do not use this.get in the executor/callbacks' (01-02-PLAN Task 2). The diff-truncated route code uses .then(model => ...) arrow functions. In Ember 3.28 arrow functions inside a findRecord .then inherit the outer this, but if any... _(flagged by: senior-dev)_
- **[MED]** Transient 5xx / offline errors may still surface as 'link expired' despite stated fix - PR description claims a fix for 5xx/offline errors being relabeled 'link expired' (adversary-review round 2). However the truncated diff does not show the transient-error discriminator in the .catch handler of routes/lesson.js. Without seeing that check, it is not possible to confirm the fix landed.... _(flagged by: senior-dev)_
- **[MED]** Race condition in lesson nonce lazy generation - Lesson#nonce lazily generates and saves a nonce on first read if missing. The controller's nil-safe guard now calls .nonce after find_by_path. Concurrent requests for a new lesson could trigger multiple nonce generations and saves, risking duplicate write conflicts or nonce inconsistency across repl... _(flagged by: adversary)_
- **[LOW]** Data isolation risk via lesson.global_id exposure - Api::LessonsController#show returns lesson content (title, description) when lesson exists and nonce matches, even with an invalid token. Lesson.global_id encodes organization scope. A user with knowledge of a nonce from another district could access lesson metadata across organizational boundaries,... _(flagged by: adversary)_

---

swahlquist:
## Response to Codex review (Request changes)

Both findings addressed, pushed in `196ba18e3..c55e66a7d`.

### High: lesson content fetched regardless of token validity — fixed in `2738f8358`

Confirmed the finding: booting the Ember shell unconditionally (this PR's own change to `boards_controller#lesson`) meant the browser always called `findRecord` -> `Api::LessonsController#show`, which renders full lesson content (title/url/description via `lib/json_api/lesson.rb`) regardless of whether the share token resolved — only the `user` block was gated. The new copy ("for your security, links stop working") didn't match that reality.

Fix: `app/views/boards/index.html.erb` now embeds `window.lesson_share_token_valid` (guarded by `@lesson`, computed server-side from the same `User.find_by_lesson_share_token` result the controller already has). `routes/lesson.js` reads and consumes this one-shot flag *before* calling `findRecord` at all — when the token is known invalid, no content-fetching API call happens for this entry point. The flag is cleared after one read so a later same-session transition to a different lesson isn't affected by a stale value.

Verification: full Rails view rendering (`render_views`) isn't feasible in this test env (confirmed via a scratch spec attempt that hit a Sprockets `vendor.js` asset-pipeline error — this repo already avoids `render_views` everywhere except one unrelated spec, for the same reason). Verified the ERB snippet's logic instead via an isolated `ERB.new(...).result(binding)` render covering all three cases (`@lesson`+`@user` nil -> `false`, `@lesson`+`@user` present -> `true`, `@lesson` nil -> nothing emitted, confirming no interference with `board`/`user` actions). 3 new frontend tests assert `findRecord` is never called when the flag is false, that it's consumed one-shot, and that the existing detection is unaffected when no flag is present.

**Explicitly not fixed (flagged separately, out of scope for this PR):** `Api::LessonsController#show`'s guard is `lesson.nonce == lesson_code || allowed?(lesson, 'view')` — content has always been reachable by anyone who knows a lesson's nonce directly via the API, independent of share-token validity. That's a separate, pre-existing architectural question (should the nonce alone permanently grant content visibility?) that predates this PR and exceeds its scope — surfacing it for a follow-up decision, not deciding it here.

### Medium: dead unused variable — fixed in `2f6ef07fd`

Removed the unused `uid` in `routes/lesson.js`.

### Test status after both fixes

- `ember test --filter="lesson expired"`: 11/11 pass (was 8/8)
- Full `ember test`: 1713 tests, 1673 pass, 40 skip, 0 fail
- `bundle exec rspec spec/controllers/boards_controller_spec.rb`: 26/26 pass (unchanged — this spec suite doesn't enable `render_views`, so the view-rendering path is verified separately as noted above)
- `eslint` clean on all touched files

swahlquist:
## Full closure of the content-disclosure finding (round 2)

Correct — the previous fix (`c55e66a7d`) only closed the fresh-navigation entry point. Extended it now to close the client-side transition path too, per Scot's direction (this touches shared serializer/controller code beyond the original phase scope, so it needed an explicit decision rather than a unilateral expansion).

**Root cause of the residual:** `Api::LessonsController#show`'s guard is `lesson.nonce == lesson_code || allowed?(lesson, 'view')` — token validity was never part of that guard, so any client-side `findRecord` call (no server-rendered page to carry a flag through) still got full content whenever the nonce matched.

**Fix, closed at the source (`ec88b8b0a`):**
- `Api::LessonsController#show` now computes `independently_authorized = lesson.allows?(@api_user, 'view', api_permission_scopes)` — calling the model's pure permission predicate directly, **not** the controller's `allowed?` helper, which renders a 400 on failure and would double-render for the common anonymous nonce-match case (verified via the `permissable-coughdrop` gem source: `allows?` has no side effects, `allowed?` does).
- Passes `withhold_content: !independently_authorized && !user` into `JsonApi::Lesson.as_json`.
- `lib/json_api/lesson.rb`'s `build_json` now gates `title`/`url`/`original_url`/`description`/`due_at`/`due_ts`/`time_estimate`/`past_cutoff`/`badge`/`noframe` (and the youtube-URL rewrite) behind `unless args[:withhold_content]`. `id`/`lesson_code`/`required`/`completed_users` stay unconditional (metadata, not content).

**Blast radius checked:** grepped every other `JsonApi::Lesson.as_json` call site (index/paginate, create, update, complete, `lib/json_api/user.rb`, `lib/json_api/unit.rb`) — none pass `withhold_content`, so it defaults falsy there. Zero behavior change for any other caller.

**New tests** (`spec/controllers/api/lessons_controller_spec.rb`): content withheld for the anonymous+unresolved-token case; content still shown when the token resolves (regression); content still shown for an independently-authorized viewer (e.g. lesson owner) with no resolved token at all — the legitimate admin-preview case (regression).

**Test status:** `lessons_controller_spec.rb` + `json_api/lesson_spec.rb` + `models/lesson_spec.rb` + `boards_controller_spec.rb`: 134 examples, 0 failures. `ember test --filter="lesson expired"`: 11/11. Full `ember test`: 1713 tests, 0 fail.

**Also fixed:** the merge conflict / dirty state from `origin/staging` moving forward (rebased, resolved a real collision in `controllers/lesson.js` against another PR's `appState` service-injection addition — kept both).

No remaining known content-disclosure gap for the unresolved-share-token case, on either entry point.

PR #585, opened 2026-07-12
chore(register): one-command regenerate-register.sh wrapper

swahlquist:
## What

Adds `scripts/regenerate-register.sh`: a single, ordered, verified command to regenerate every derived compliance artifact after editing `FINDINGS.json` / `DOCUMENT-REGISTER.json`.

## Why

Today the derived artifacts are rebuilt by five separate render scripts that must run in the right order, and CI's `audit-artifacts-integrity` job fails the merge if any one drifts. Running them by hand (and remembering all of them) is a recurring footgun — and it makes concurrent register-touching branches risky to reconcile. This wrapper collapses that into one step.

## Behavior

- **Gate:** runs `citation-check` first and refuses to render onto a register whose evidence does not resolve.
- **Renders in dependency order:** document-register (normalizes JSON hashes) before publication-status; plus FINDINGS.md, compliance calendar, local Notion mirror, publication-status.
- **Idempotent output:** suppresses the Notion mirror's timestamp-only churn, so the caller's diff shows real changes exclusively.
- **Re-verifies:** re-runs the four `audit-artifacts-integrity` `--check` commands **plus** a `citation-check` evidence gate. Note `citation-check` is intentionally *not* part of the CI integrity job (see `ci.yml`); gating on it locally is a stricter, correct pre-render safeguard. Green here == green integrity job.
- **`--check` mode:** verify without writing (mirror CI locally).
- **Local-only:** never pushes to Notion/Drive (those sync in their own CI workflows), so it is safe to run anytime with no secrets.

Also wires the wrapper into the `promote-finding` skill (Step 5) so a promotion regenerates *every* derived artifact, not just FINDINGS.md.

## Testing

- `--check` against the current register: all four integrity checks pass (97 PASS / 0 FAIL on citation-check).
- Full write mode against an unchanged register: exits 0 with an empty artifact diff (fully idempotent, no timestamp noise).
- Runs directly via the executable bit (committed `100755`); `bash -n` syntax check clean.

## Review fixes (Codex senior-dev pass)

- Set the script executable in the git index (`100755`). `/mnt/c` is a Windows mount so `chmod` alone did not carry the mode into the commit; the docs invoke it directly, which would be permission-denied on a fresh checkout.
- Corrected wording: the wrapper runs the CI artifact checks *plus* a stricter local citation gate (citation-check is not itself a CI job).

Tooling/dev-experience only; no product code, no runtime behavior change. Does not touch any finding's status (only Scot closes/downgrades).

### DIFF
--- a/.claude/skills/promote-finding/SKILL.md
+++ b/.claude/skills/promote-finding/SKILL.md
@@ -121,16 +121,38 @@ for this reason, cite a different (cleaner) line of the same issue, or paraphras
 line; do not fight the gate. This in-script gate is redundant with the n8n bot's upstream
 pre-scrub, so conservative is correct.
 
-## Step 5: Validate + render
+## Step 5: Validate + regenerate all derived artifacts
+
+Promoting a finding can change severity counts and bundle membership, so it is not enough to
+rebuild `FINDINGS.md` alone — the Notion mirror, compliance calendar, document register, and
+publication-status report all derive from the register and CI's `audit-artifacts-integrity` job
+fails the merge if any drifts. Regenerate them all in one governed, ordered step:
 
 ```
-ruby scripts/citation-check.rb audit-reports/FINDINGS.json          # expect exit 0
-ruby scripts/citation-check.rb --render audit-reports/FINDINGS.json # rebuild FINDINGS.md
+scripts/regenerate-register.sh          # gate on citation-check, render all artifacts, re-verify
 ```
+
+This gates on `citation-check` first (refuses to render onto a register whose evidence does not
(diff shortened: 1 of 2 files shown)


swahlquist:

## Automated PR Review

### Verdict: `APPROVE`

| --- | --- | --- |

**Senior-dev:** Pure developer-tooling and documentation PR with no product code, no PII surface, and no runtime behavior changes. The shell script is safe, idempotent, and well-commented.

**Adversary:** Wrapper script correctly gates on citation-check, renders artifacts in dependency order, and re-verifies CI integrity checks; no security or runtime issues introduced.

### Findings (0 unique)
_No findings._

---

PR #604, opened 2026-07-14
docs: add PR Preflight gate (P1-P6) to CLAUDE.md

swahlquist:
Deliverable 1 of 2 from the 2026-07-12 "why Codex catches so many issues" thread. Adds a mandatory `## PR Preflight` section to `CLAUDE.md` that shifts the highest-frequency reviewer-caught miss classes left of CI and review: claim-vs-HEAD verification (P1), entry-point enumeration for auth/access changes (P2), generated-artifact + git-metadata checks that mirror the `audit-artifacts-integrity` CI job (P3), cross-doc consistency sweep (P4), an honest status block required in every PR body (P5), and a behavioral definition of done for UI flows (P6).

Docs-only change to an agent-instruction file. No feature flag, no i18n, no runtime code touched.

## Fix status
| Item | Status | Evidence |
| --- | --- | --- |
| P1-P6 block inserted after `### Security` | Fixed | `CLAUDE.md` new `## PR Preflight` section |
| P3 lists every check the CI gate runs | Fixed | matches all 5 in `.github/workflows/ci.yml:115` audit-artifacts-integrity (calendar, notion, document-register, publication-status, capability-check) |
| All cited scripts exist and accept `--check` | Fixed | verified at HEAD for all 5 |
| exec-bit check scoped to directly-invoked changed scripts (not all of scripts/) | Fixed | P3 comment block |

## Not covered by this PR
- The Codex-review pipeline itself (Deliverable 2) is a separate build, tracked in `ai-company-brain outputs/docs/2026-07-12-codex-review-pipeline-build-spec.md`.
- P6 behavioral coverage is a documentation rule only; no automated UI smoke suite is added here.

## Review note
A Codex senior-dev pass on the first commit caught that P3 omitted the 5th CI check (`capability-check.rb --check`, ci.yml:139), which would have let a green preflight still fail the CI gate on capability-ledger drift. Verified against HEAD and amended before this PR was opened. That miss-then-catch is the exact pattern this gate exists to prevent.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #607, opened 2026-07-14
feat(codex-review): scaffold Codex review pipeline (Actions side)

swahlquist:
## Summary
Implements the GitHub Actions side of the Codex review pipeline per `outputs/docs/2026-07-12-codex-review-pipeline-build-spec.md` (build order steps 2, 3, 4, 6, 8 of 8). Companion to #604 (CLAUDE.md PR Preflight P1-P6 gate) and the two-tier AI data-routing policy (#127/#129, already merged).

- `.github/codex/review-prompt.md`, `review-schema.json`, `REVIEW-MEMORY.md`
- `scripts/codex-review-path-classifier.sh` -- vendored data-bearing-path guard (Guard A) + Tier 2 compliance-path routing (Guard B), `CODEX_COMPLIANCE_PATHS` default `allow`
- `.github/workflows/codex-review.yml` -- fail-closed pending-status-first review job; `codex exec` / `claude-deep` / `blocked` reviewer routes; verified POST+GET success condition so the job cannot report false success
- `.github/workflows/codex-watchdog.yml` -- dual-trigger (`workflow_run` + 30-min `schedule` sweep) fail-closed backstop
- `.github/workflows/claude-fix.yml` -- headless Claude fix pass, shipped **DISABLED** (`vars.CLAUDE_FIX_ENABLED != 'true'`) per the spec's week-1 review-only posture

## Fix status
| Item | Status | Evidence |
|---|---|---|
| Prompt/schema/memory files | Fixed | `.github/codex/` |
| Path classifier (Guard A + B) | Fixed | `scripts/codex-review-path-classifier.sh`, self-tested |
| codex-review.yml | Fixed | fail-closed pending-first step, input validation before `ref:` use |
| codex-watchdog.yml | Fixed | dual trigger, self-tested `jq` extraction |
| claude-fix.yml | Fixed, disabled | gated behind `vars.CLAUDE_FIX_ENABLED` |

## Not covered by this PR (explicit, out of scope here)
- **Branch protection is NOT made required yet.** Flipping `codex-review/deep-pass` to required before this workflow is proven live via `workflow_dispatch` would leave the status permanently unresolved and block every merge to `staging`. This needs a deliberate follow-up once this workflow is tested.
- **CI secrets are not installed** (`CODEX_OPENAI_API_KEY`, an Anthropic key for the `claude-deep`/`claude-fix` routes, and the eventual `N8N_CODEX_RESULTS_WEBHOOK_URL`/`HMAC_SECRET` once W2 exists).
- **n8n W1 (`codex-orchestrator`) and W2 (`codex-results`) are not built.** Per the spec's own build order this is a separate focused session; it's also a live mutation of the production PR-review-bot n8n instance, so it's being held for an explicit go-ahead.
- `codex-review.yml`'s `workflow_dispatch` path has not been exercised end-to-end (no PR to dispatch against yet, no secrets installed).

## Test plan
- [x] All 3 new workflow YAML files parse (`yaml.safe_load`)
- [x] All new shell scripts pass `bash -n`
- [x] All new Python scripts pass `py_compile`
- [x] `review-schema.json` is valid JSON
- [x] `codex-review-path-classifier.sh` self-tested against a real diff range
- [x] `codex-review-assemble-prompt.py` self-tested against the real prompt/memory files
- [x] `codex-review-claude-deep-extract-response.py` self-tested against both a well-formed and a malformed Anthropic response
- [x] Exec bits verified via `git ls-files -s` (this repo has `core.filemode=false`, so `chmod` alone doesn't get tracked)
- [ ] `workflow_dispatch` smoke test against a real closed PR (blocked on `CODEX_OPENAI_API_KEY`)
- [ ] Forced-failure test confirming `codex-watchdog.yml` flips the status to `failure`

### DIFF
--- a/.github/codex/REVIEW-MEMORY.md
+++ b/.github/codex/REVIEW-MEMORY.md
@@ -1,0 +1,58 @@
+# Codex review memory
+
+Curated rules the automated reviewer has learned across sessions. Each entry
+names the source so a rule can be re-evaluated if its origin turns out to be
+wrong. This file is injected into every review prompt (see
+`.github/codex/review-prompt.md`) -- keep it under ~200 lines and free of raw
+session excerpts or any identifiable data; it is not `audit-reports/` or
+`docs/legal/` territory, but treat it as reviewer-facing, not user-facing.
+
+When a human overrides a verdict, or a post-merge bug slips through review,
+append one short rule here in its own small PR. Do not batch memory edits
+into an unrelated PR.
+
+## Rules
+
+- **Register drift is a merge blocker, not a style nit.** The
+  `audit-artifacts-integrity` CI job runs five `--check` scripts
+  (`compliance-calendar-render.rb`, `compliance-notion-publish.rb`,
+  `document-register-render.rb`, `compliance-publication-status.rb`,
+  `capability-check.rb`). A PR that edits `audit-reports/FINDINGS.json` or
+  `audit-reports/DOCUMENT-REGISTER.json` without regenerating the
+  corresponding rendered artifact will fail this gate. Verify by actually
+  running the `--check` scripts, not by reading the diff.
+  -- source: repo convention, `reference_register_edit_regenerate_artifacts`.
+
+- **Stale provider-version wording is a real finding, not pedantry.**
+  Compliance and infra docs that name a specific model/CLI version
+  (`gpt-5.5`, `codex-cli 0.144.1`, etc.) drift silently. If a PR touches a
+  doc naming a provider/version, confirm the version claim against the live
+  tool (`codex --version`, vendor changelog) rather than assuming the doc is
+  current.
+
+- **Flaky `persistence-sync` timing failures on `main`-targeting PRs are CI
+  noise, not a regression the PR introduced.** `main` has no required
+  status checks, so this does not block merge; do not raise it as a
+  blocking finding unless the PR's diff touches `persistence.js` sync logic
+  directly. -- source: `reference_main_build_and_test_persistence_sync_flake`.
+
+- **A negative existence claim ("file X does not exist") requires the same
+  evidence standard as a positive claim.** Search all roots (`lib/`, `app/`,
+  `config/`), not only `app/`, before asserting absence. A prior review round
+  of this very pipeline's design spec wrongly claimed
+  `lib/json_api/lesson.rb` did not exist because the search was scoped to
+  `app/` only. -- source: this pipeline's own build spec, round 1 -> round 2
+  correction (FLAG-3 reversal).
+
+- **Policy changes must update every canonical consumer in the same PR.**
+  Before approving an edit to a cross-cutting policy file (e.g.
+  `instructions/shared/compliance.md`), grep `instructions/`, `commands/`,
+  and `agents/` for other files stating the same rule. A policy edit that
+  updates one consumer and leaves siblings contradicting it is a finding,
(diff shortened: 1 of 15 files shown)


swahlquist:

## Automated PR Review

### Verdict: `COMMENT`

| --- | --- | --- |

**Senior-dev:** The pipeline scaffold is well-structured with good fail-closed logic and input validation. Two concrete issues worth addressing before the workflow goes live: a missing HTTP-status check on the Anthropic curl call, and a code path in the path-classifier's tee invocation that silently drops output when GITHUB_OUTPUT is unset.

**Adversary:** Security and logic issues identified, including potential secret exposure, insufficient input validation, and race condition risk.

### Findings (8 unique)
- **[HIGH]** Secret exposure via environment variable interpolation - claude-fix.yml step 'Run headless Claude Code fix pass' uses echo ::add-mask::$ANTHROPIC_API_KEY. The GitHub Actions masking feature only works on exact matches; printing the secret to stdout could expose it in logs if the step fails before the mask is registered. _(flagged by: adversary)_
- **[MED]** claude-deep.sh: curl exit code checked but HTTP status is not - codex-review-claude-deep.sh uses curl -sS ... -o $RESPONSE_FILE with no -w '%{http_code}' capture and no check that the HTTP response is 2xx. A 4xx/5xx Anthropic error (rate-limit, bad key, etc.) writes an error JSON to RESPONSE_FILE; the extractor then exits 1 with 'no submit_review block', which t... _(flagged by: senior-dev)_
- **[MED]** Path-classifier tee idiom silently drops GITHUB_OUTPUT lines in local mode - The final block pipes through tee -a ${GITHUB_OUTPUT:-/dev/stdout} >/dev/null. When GITHUB_OUTPUT is unset the substitution is /dev/stdout, then >/dev/null silences it, so local self-test runs produce no visible output. This is a usability/debuggability issue for the self-test path the PR descriptio... _(flagged by: senior-dev)_
- **[MED]** Insufficient input validation for HEAD_REF - claude-fix.yml's HEAD_REF validation pattern ^[A-Za-z0-9._/-]+$ allows '.' and '/' anywhere, which could permit directory traversal if used unsafely in shell commands. Should restrict to safe branch naming conventions. _(flagged by: adversary)_
- **[MED]** Race condition in pr_number to head_sha binding - codex-review.yml's 'Bind pr_number to head_sha' step fetches the PR's current head SHA and compares it to the dispatched head_sha. If the PR is updated between this check and later steps (e.g., checkout), the review may act on a stale commit. _(flagged by: adversary)_
- **[LOW]** codex-review-post-fix-comment.sh opens file via bare open() without explicit encoding - The inline Python heredoc uses json.load(open(findings_file)) with no explicit encoding='utf-8'. On a runner whose locale is not UTF-8 this can raise a UnicodeDecodeError if any finding description contains non-ASCII. Use open(findings_file, encoding='utf-8') for consistency with the other Python sc... _(flagged by: senior-dev)_
- **[LOW]** Missing validation for findings_json input - claude-fix.yml accepts findings_json as a string input but only validates it's a string; the Python step json.load() may raise exceptions if the JSON is malformed, potentially halting the workflow without clear error context. _(flagged by: adversary)_
- **[LOW]** Watchdog cron schedule may miss pending statuses - codex-watchdog.yml's schedule trigger runs every 10 minutes but checks for statuses pending >30 minutes. A status set just after a sweep could remain pending for nearly 40 minutes before being flagged, exceeding the intended 30-minute timeout. _(flagged by: adversary)_

---

PR #610, opened 2026-07-15
fix(codex-review): pass -R explicitly to gh pr calls run before checkout

swahlquist:
## Summary
The first live workflow_dispatch smoke test of `codex-review.yml` (#607, run 29424797637) failed at the "Bind pr_number to head_sha" step with `fatal: not a git repository`. Root cause: `gh pr view` without `-R` auto-detects owner/repo from the local git remote, but that step deliberately runs before "Checkout PR head" (we don't want to trust an unverified `head_sha` enough to check it out first). With no checkout yet, `gh` had nothing to auto-detect against.

Fix: pass `-R "${{ github.repository }}"` explicitly on every `gh pr` call in the job, including the two in "Gather live state" that happened to work only because they run after checkout — an implicit ordering dependency that shouldn't have been load-bearing.

## Test plan
- [x] YAML parses (`yaml.safe_load`)
- [x] Every `run:` block passes `bash -n`
- [ ] Re-dispatch the smoke test once merged and confirm the bind-check step passes

### DIFF
--- a/.github/workflows/codex-review.yml
+++ b/.github/workflows/codex-review.yml
@@ -101,7 +101,7 @@ jobs:
         # to proceed -- and flip the status to failure -- if pr_number's
         # actual current head doesn't match the dispatched head_sha.
         run: |
-          ACTUAL_HEAD="$(gh pr view "$PR_NUMBER" --json headRefOid --jq '.headRefOid')"
+          ACTUAL_HEAD="$(gh pr view "$PR_NUMBER" -R "${{ github.repository }}" --json headRefOid --jq '.headRefOid')"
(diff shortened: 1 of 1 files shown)


swahlquist:

## Automated PR Review

### Verdict: `APPROVE`

| --- | --- | --- |

**Senior-dev:** Minimal, correct fix: every gh pr call in the workflow now carries an explicit -R flag, eliminating the implicit dependency on a prior checkout. No application code, PII surfaces, or user-facing behavior is touched.

**Adversary:** The diff correctly adds the -R flag to all gh pr calls, fixing the missing git repo error and eliminating an implicit ordering dependency.

### Findings (0 unique)
_No findings._

---

PR #584, opened 2026-07-12
Release: staging → main - prod catch-up (supersedes #583; SES observability, Ember 5.12, EU AI Act Art.50, security hardening)

swahlquist:
## Supersedes PR #583

PR #583 was opened directly from `staging` as head and conflicted with `main` in 6 files (main carries two direct-to-main backport PRs, #538 and #539, not reachable from `staging`). Since GitHub can't cleanly repoint a PR to a different head branch, #583 was closed as superseded. This PR is opened from a dedicated conflict-resolution branch, `scot/chore/staging-into-main-2026-07-12`, cut from `staging` with `main` merged in and all conflicts resolved.

## Current state

- **Prod (`main`) SHA:** `b03fb7cf1c4812cb5430c0dfc282e4c86b56a874` - confirmed live on Render `lingolinq-prod` (deploy `dep-d95ivhp9rddc73baqfrg`)
- **Staging source SHA:** `e80dc26d4d0c2ffd97fb5e0d7474b408e88705d8`
- **This branch:** `scot/chore/staging-into-main-2026-07-12`, `staging` + a merge of `main` with conflicts resolved

## Conflict files and resolution

Six files conflicted (all stemming from direct-to-main backports #538 "SigV4 PresignedPost signing" and #539 "gem CVE patches" landing 2026-07-06, in spots `staging` later touched independently):

- `Gemfile`, `Gemfile.lock`
- `app/frontend/app/components/beta-feedback-form.js`
- `app/frontend/app/services/content-grabbers.js`
- `lib/uploader.rb`
- `spec/lib/uploader_spec.rb`

**Resolution rule applied:** staging's version was used for each file **only after verifying, file by file, that it preserves the direct-to-main hotfixes** - not a blind "take staging."

[...]

PR #615, opened 2026-07-16
chore: untrack .planning/ so the existing gitignore rule applies

swahlquist:
## What

Removes ten `.planning/` files from the git index. Nothing else changes: 0 lines added, 1343 removed, and **every file stays on disk**.

## Why

`/.planning` has been at `.gitignore:95`, but the rule never took effect. The expired-link-UX GSD project committed `.planning/` at `c3a7efa09` (2026-07-10), and gitignore does not retroactively untrack files that are already committed. Those ten files have been tracked on `staging` ever since, and shipped to `main`/prod via #584.

The consequence is subtle and it bites in practice. The tracked control files describe one specific GSD project (expired-link UX). `.planning/` is a single slot, so any session that scaffolds a *different* project into it turns those files into modified **tracked** files, and local planning scratch silently rides along in an unrelated feature PR. A fresh checkout or new worktree re-materializes them, so archiving them locally does not help. Untracked files under `.planning/` *are* correctly ignored, which is what makes this easy to miss: `git status` stays clean until you touch one of the ten.

This is exactly what the gitignore rule intended all along. `.planning` keeps working as local GSD scratch, just without the index entries.

## Verification

- `git show --numstat 76e2c15c6` shows 0 additions across all ten files -- index removal only.
- Files confirmed present on disk after the change.
- `git check-ignore -v .planning` now resolves to `.gitignore:95:/.planning`, where it previously reported not-ignored.
- Working tree clean.

## Note for reviewers

One of the ten, `.planning/phases/02-disclosures-content/02-01-SUMMARY.md`, is a VPC leftover swept in by accident rather than an expired-link file. It comes out here with the rest; it remains on disk.


swahlquist:
OK to merge. Required CI is green and the diff only untracks .planning/ files that are already ignored. The failing codex-review/deep-pass status on this PR is a Codex runner sandbox/tooling false positive, not a PR defect: the review run could not execute read-only shell commands because bubblewrap failed with `bwrap: loopback: Failed RTM_NEWADDR: Operation not permitted`. The Node 20/24 note is also a GitHub Actions runtime warning for `actions/checkout@v4`, not a project Node version change. LingoLinq-AAC remains pinned to Node 20 for Ember 3.28.

PR #638, opened 2026-07-20
security(sentry): fail closed for COPPA in scrubber + transaction filter

swahlquist:
Narrow hardening of the Sentry COPPA scrubber (`config/initializers/sentry.rb`). No behavior change on the happy path; the changes only affect what happens when the child-user determination cannot be made cleanly. Scope is deliberately limited to two files.

## What changed
- **`CoppaSentryScrub.child_user?` fails closed.** When `coppa_parental_consent_pending?` raises (corrupt settings blob, decryption failure, lookup timeout), it now returns `true` so the event is scrubbed instead of shipped. `nil` / anonymous users still return `false` (an unauthenticated request is not a known child). A user object that does not respond to the predicate still returns `false`.
- **`TRANSACTION_FILTER` fails closed.** If the inner lookup/filter chain raises, it now returns `nil` (drops the transaction event) rather than shipping a potentially-child trace carrying URL global_ids and a stable per-IP fingerprint. Pairs with `child_user?`.
- **`keep_cache_error` escape-hatch tag value coercion.** Boolean `true` or the string `"true"` (any case), under either the symbol or string key, trips the hatch. Falsy / non-true strings still drop cache errors.
- Comment tone tightened (removed a legal conclusion and some editorializing); the technical rationale and the nil-user note are preserved.

## Fix status
| Item | Status | Evidence |
| --- | --- | --- |
| child_user? fail-closed on raise | Fixed | config/initializers/sentry.rb child_user? (rescue => true); spec `#child_user? (M1 fail-closed contract)` |
| nil / anonymous still non-child | Fixed | child_user? `return false unless user`; spec "returns false for a nil user" |
| TRANSACTION_FILTER fail-closed on raise | Fixed | config/initializers/sentry.rb TRANSACTION_FILTER (rescue => nil); spec "returns nil (drops event) when the inner pipeline raises" |
| keep_cache_error accepts bool/string "true", any case, either key | Fixed | keep_cache_error_tag? coercion; specs for string "true", string key, and falsy values |

## Not covered by this PR
- No changes outside these two files. Scrubber redaction rules, sampling, and the broader Sentry init are unchanged.

## Test
`DB_USER=scotw RAILS_ENV=test bundle exec rspec spec/initializers/sentry_spec.rb` -> 62 examples, 0 failures.

## Author-Model: opus-4.8


swahlquist:
## Codex Review

**Verdict:** NEEDS_HUMAN
**Route:** codex
**Head:** 3690eb3de48375586cbb69b9fc18a2fbdd8109ec
**Run:** https://github.com/lingolinq/LingoLinq-AAC/actions/runs/29714433214

Codex review completed.

| --- | --- | --- | --- |
| LS-1 | HIGH |  | The Codex sandbox is currently preventing required read-only inspection commands from running, so I cannot verify the PR body claims, drift checks, file modes, legal/audit references, or current CI state from the checkout. |

MelissaOneil (approved):

swahlquist:
## Codex Review

**Verdict:** REQUEST_CHANGES
**Route:** codex
**Head:** a96cb695e6fbaa1d30772f878eaf62a4aa4d4e18
**Run:** https://github.com/lingolinq/LingoLinq-AAC/actions/runs/30211375807


Codex review completed.

| --- | --- | --- | --- |
| F1 | MED | config/initializers/sentry.rb | The keep_cache_error escape hatch does not actually honor either key when both symbol and string tag keys are present and the symbol value is any truthy non-true value. |
| EVIDENCE-1 | HIGH | (diff-wide) | One or more chunk reviews blocked or were inconclusive. |

PR #640, opened 2026-07-20
Traci/fix/persistence sync flake

traciday:
## Summary

Fixes an intermittent failure in `persistence-sync-test.js` (~20-50% of runs)
that surfaced in two different ways — an assertion failure on
`expect(reloads).toEqual(...)` and a 15s timeout — both traced to a single
harness defect.

Also merges current `staging` into the branch so it lands cleanly.

## Root cause

`stubBoardReloadTracking` stubbed `reload` on board record **instances**
captured via `peekRecord` at setup time. Sync unloads and re-pushes board
records mid-traversal (`refreshBoardsInStore`, `store.push` in `sync_boards`),
and Ember Data returns a **new instance** each time — silently discarding the
stub. One defect, two symptoms:

- The board's real `reload()` ran and was never recorded, so the assertion saw
  a missing key for a reload that **actually happened** — an untracked-observation
  bug in the harness, not a missing reload in the code under test.
- That real `reload()` performed a genuine store fetch against stubbed ajax,
  which could hang, stalling `waitForSyncDoneAndSettled` into the 15s timeout.

## Fix

Stub `LingoLinq.Board.prototype.reload` keyed by board id, so tracking survives
record replacement. Untracked boards pass through to the real `reload()`.

Additionally, the "important ids" test polled `persistence.important_ids` only.
`utils/persistence.js` exports a Proxy with a `get` trap and **no** `set` trap,
so reads forward to `window.persistence` — which in a full-suite run can be a
different or torn-down service instance than the one that synced. That test
passed in module isolation and timed out only in the full suite. It now also
polls the resolved `persistenceTarget()`.

**Test-only change.** No application code is modified.

## Fix status

| Item | Status | Evidence |
|---|---|---|
| Reload tracking survives record replacement | Fixed | `persistence-sync-test.js` — prototype stub keyed by board id |
| Assertion flake on `expect(reloads)` | Fixed | 3 previously-flaky tests stable across full-suite runs |
| 15s timeout flake | Fixed | real `reload()` no longer fetches against stubbed ajax |
| "important ids" full-suite-only timeout | Fixed | now polls resolved `persistenceTarget()` |
| Branch current with `staging` | Done | merge `598350021`, no conflicts |

## Not covered by this PR

- No application code changes — harness only; does not alter sync behavior.
- Does not address the underlying `utils/persistence.js` Proxy having a `get`
  trap with no `set` trap. The test now works around it; the Proxy asymmetry
  remains and could bite other suites.
- No change to the other 18 pre-existing backend spec failures (see below).

## Verification

- **Frontend:** 1717 tests — 1677 pass, 40 skip, **0 fail**
- **Backend:** 6415 examples, **18 failures**, 51 pending — all pre-existing and
  unrelated (`add_premium_voice`, `track_protected_source`, plus long-standing
  TODO specs like *"should do something worth speccing"*). Confirmed by line
  number: the merge added `user_spec.rb:4493-4601`; all failures sit at
  2097-2200 and 3751-3784.
- **Merged specs all pass:** `article50_call_context`, `pii_scrubber`,
  `eu_jurisdiction`, `eval_narrator`, `ai_board_generator`, `ai_word_predictor`,
  `ai_api_log`, `eval_sessions_controller`.


swahlquist:
## Codex Review

**Verdict:** NEEDS_HUMAN
**Route:** codex
**Head:** 5983500219d925bb288156dbabe92828ddd2a378
**Run:** https://github.com/lingolinq/LingoLinq-AAC/actions/runs/29754906953

Codex review completed.

| --- | --- | --- | --- |
| CR-1 | HIGH |  | Required read-only inspection commands cannot run in this Codex sandbox, so I cannot verify the PR-body claims, changed file contents, drift checks, file modes, docs claims, or current local HEAD as mandated. |

MelissaOneil (approved):

PR #644, opened 2026-07-20
# fix(board-detail): Ember 5.12 regressions, styling refinements, a11y & Node 22

traciday:
# fix(board-detail): Ember 5.12 regressions, styling refinements, a11y & Node 22 - merged in from Scot's PRs

## Summary

Follow-on work to the board-detail redesign (#153), plus this cycle's merges:
Ember 5.12 regression fixes, accessibility/lint conformance, styling refinements,
the Ember-upgrade audit system, and a repo-wide Node 20→22 bump.

**Scale:** 314 commits · 426 files · +10,573 / −11,466
Current `staging` merged in cleanly (`b52e2868a`).

---

## 📖 How to review

Grouped by theme; risk is concentrated in one small section.
**If you review one thing, review the Ember 5.12 fixes and the Node bump.**

| Theme | ~commits | Review weight |
|---|---:|---|
| 🎨 Styling / design | 53 | Author-verified in browser; low. |
| ♿ A11y / lint | 43 | CI-enforced; objective. |
| 🖥️ board-detail | 25 | Includes the bug fixes below. |
| ⚠️ **Ember 5.12 migration** | 13 | **Highest risk — please scrutinise.** |
| 🧰 **Node 20 → 22** | — | **Infra — affects CI, Docker, Render deploy.** |
| 🔍 Ember-audit system | — | Additive tooling; no runtime effect. |

Despite the branch name, this is mostly `fix` (65) + `feat` (56); only 8 are `style`.

---

## ⚠️ Ember 5.12 fixes — the part that needs eyes

The migration required hand-rolled compensation for `{{on}}` not receiving
synthetic clicks, and that machinery caused real, silent regressions:

1. **Options (⋮) menu dead after an edit round-trip** — the chrome-dispatch guard
   bailed when neither `speak_mode` nor `edit_mode` was set, so raw_events
   dispatched the action *and* Ember's `{{on "click"}}` fired it → the toggle
   net-cancelled. `preventDefault()` on `mouseup` doesn't cancel the follow-up
   `click`; on `touchend` it does — which is why it appeared to "heal" with
   DevTools open. Fixed by deferring real mouse clicks to Ember in every mode.
2. **Speak mode lost on edit exit** — `board-detail.edit` is a child route, so
   leaving it never re-runs the parent's setupController; the child's
   resetController now restores `current_mode='speak'` unconditionally.
3. **Edit-sidebar drag never initiated** — drag handlers were wired to a
[...]

PR #642, opened 2026-07-20
compliance(register): file 3 pre-existing modal/consent/a11y defects found during Art.50 Phase 3 planning

swahlquist:
Register-only change. No application code is touched.

Three defects surfaced during VPC Phase 3 (EU AI Act Art. 50 disclosure modal) planning, while cloning `terms-agree` as the acknowledge-gate exemplar. All three **predate Phase 3** and are deliberately scoped **out** of it: Phase 3 builds the Art.50 modal correctly and guards against them, but does not repair them. Legacy repair is separate prioritized work.

Each is anchored to staging tip `2a10027a1` with a source-verified evidence snippet, filed `open` / `unassigned`. No dispositions, no severity downgrades — only Scot closes a finding.

| ID | Severity | Frameworks | Finding |
|---|---|---|---|
| `LL-104bfa61dc` | high | WCAG | terms-agree unreachable by switch scanning |
| `LL-53cb93fab1` | high | GDPR, FERPA | terms-agree replaced before consent is presented |
| `LL-58130aaefe` | medium | WCAG | modal-dialog missing `aria-labelledby` / `aria-describedby` |

### LL-104bfa61dc — switch scanning dead on terms-agree
Buttons use `la-btn` / `la-btn--ghost` with no `.modal_targets` wrapper and no bare `.btn`, so the scanner selector at `utils/modal.js:255` matches nothing. All four call sites (`routes/index.js:123,132`, `routes/bento.js:51,55`) also call `modal.open('terms-agree')` with no options object, so `scannable: true` is never passed. Both halves need fixing; either alone is insufficient.

Rated high rather than medium because the modal is `uncloseable` — a switch-scanning AAC user is not degraded, they are stuck, and that is a core LingoLinq population.

### LL-53cb93fab1 — terms-agree never presented on the sync path
`routes/index.js` terms_agree-missing branch never returns, so when `show_intro` is set it falls through to `modal.open('intro')`. On the synchronous fresh-model path, both opens happen in one run loop; Ember renders only the final `currentTemplate` (`intro`), so terms-agree never mounts. The first open resolves with `{replaced: true}` (`utils/modal.js:138`, `services/modal.js:92`). The user does not see or confirm Terms that visit; `terms_agree` stays false (no false-positive consent record).

[...]

PR #635, opened 2026-07-19
compliance(art50): AI call-site enforcement (shared call-context helper, Phase 4)

swahlquist:
## Phase 4: AI Call-Site Enforcement (the shared call-context helper)

The linchpin for the EU AI Act Article 50(1) deadline (2026-08-02). One shared helper stamps `jurisdiction` and `article_50_disclosure_shown` onto every `AiApiLog` row at all three AI call sites, which un-inerts `AiApiLog.purge_old_eu_logs!` (it previously matched zero rows) and gives the forthcoming disclosure modal the state it reads.

### What shipped
- **ENF-01 shared helper** `LingoLinq::Article50CallContext.for(user)` -> `{ jurisdiction:, article_50_disclosure_shown: }`. Single source of the full call context: owns the guarded reads, the lib `require_relative`s, and a scrubbed (class-only, no PII) `Rails.logger.warn` on fallback. All three `log_ai_call` wrappers (`ai_board_generator`, `ai_word_predictor`, `eval_narrator`) call only it and merge the two explicit params into the existing sink. No per-site re-implementation.
- **D-01 retention fail-safe** `EuJurisdiction.retention_stamp(user)` maps ONLY confirmed `:eu` -> `'EU'`; both `:non_eu` and `:unknown` -> `nil`. The `jurisdiction` column drives `purge_old_eu_logs!` (5yr delete), and those rows double as a HIPAA six-year audit trail (45 CFR 164.316(b)(2)), so an unsure user is never stamped `'EU'` and never deleted inside the floor. This is the opposite fail-safe direction from disclosure (which stays true for `:eu` AND `:unknown` and is unchanged).
- **B3 disclosure-shown state** on `User` under `settings['ai_transparency']`, cloning the shipped `grant_ai_consent!` (versioned, `with_lock(requires_new: true)`, idempotent, `SecureRandom.uuid` record_id, exactly one `AuditEvent`, `secure_serialize` inherited). Write API only; `shown` stays `false` in production until the Phase 3/5 modal writer ships (by design).
- **PN-02** dedicated `LingoLinq::Article50Disclosures::CURRENT_VERSION`, decoupled from `AiConsentDisclosures` so an Article 50 copy change never forces a COPPA re-consent (and vice versa).

### Decisions honored
[...]

PR #569, opened 2026-07-10
compliance(coppa): VPC Phase 2 -- AI data-sharing disclosure content (attested by Scot, pending es + outside counsel)

swahlquist:
## Summary

VPC Phase 2 (Disclosures Content) of the AI Data-Sharing VPC project (COPPA second-tier AI consent). Builds the versioned, i18n-ready disclosure the consent gate will reference, and reconciles the public privacy policy and internal legal docs to match.

- `LingoLinq::AiConsentDisclosures` module -- `CURRENT_VERSION = 1`, per-version metadata (vendors, retention, data categories) + content hash
- `GET /ai_consent/disclosures/:version` -- server-rendered disclosure (single source of truth; a future Ember modal fetches + caches it, per Phase 3), all 7 COPPA 16 CFR 312.4(c) direct-notice elements
- `privacy.hbs` extended (not replaced) -- AI vendor naming, disclosure link, and AiApiLog retention reconciled into a scoped table (EU 5yr enforced / children 12mo rolling-out / general 24mo rolling-out / IP 90-day enforced), replacing a stale blanket claim
- `docs/legal/AI_DATA_SHARING_CONSENT.md` + `docs/legal/AI_DATA_FLOW_CLASSIFICATION.md` -- rationale, four-bucket data classification, sub-processor basis, versioning/revocation policy, open counsel questions
- `User#ai_consent_granted?` now defaults its version argument to the new constant

**⚠️ DRAFT status -- not final.** Task 02-02.8 (Scot + counsel legal sign-off) is intentionally **not** part of this PR; it's a business/legal decision, not engineering work. Three items block that sign-off:
1. Whether scrubbed, neutral AI board generation can be treated as Non-personal (exempt) or must stay gated
2. Which verifiable-parental-consent method satisfies COPPA for this disclosure (email-plus/text-plus excluded)
3. Google Gemini 2.5 Flash is a real conditional fallback vendor (`lib/ai_board_generator.rb`, `lib/ai_word_predictor.rb`) hitting the AI Studio endpoint, not Vertex AI -- matches an already-attested open item in `AI_GOVERNANCE_MEMO.md` section 7, but blocks sign-off since the disclosure now names this vendor

This content must not be treated as final or shipped to a real parent until that sign-off happens (VPC Phase 3/4 are separately gated on it too).

## Verification

Independently verified by a second-pass gsd-verifier agent (not just the executor's own self-check):
- 13/13 phase must-haves confirmed against actual code, not just the SUMMARY.md's claims
[...]

PR #646, opened 2026-07-21
compliance(art50): Phase 3 - EU disclosure content, ack endpoint, accessible modal, first-AI-use gate

swahlquist:
VPC Phase 3 of the EU AI Act Article 50 milestone. Builds on Phase 4 (#635, the shared call-context helper and disclosure-shown user state) and Phase 2 (#569, the COPPA disclosure content pattern).

**Ships inert by design.** `article_50_disclosure` is deliberately NOT registered in `AVAILABLE_FRONTEND_FEATURES`; that is Phase 5 / RLL-01. Nothing in this PR is user-visible until the flag is enabled.

## The goal

An EU-or-unknown-jurisdiction user is told they are interacting with AI before their first AI interaction of any kind, acknowledges it once, and that acknowledgement is recorded through `User#mark_article_50_disclosure_shown!`, without ever interrupting AAC communication.

## What landed

| Plan | Delivers |
|---|---|
| 03-01 | Art.50 disclosure content: `Article50Disclosures` REGISTRY + content hash, `art50_v1.html.erb`, en + es locales, served via the existing `GET /ai_consent/disclosures/:version` |
| 03-02 | `POST /api/v1/users/:user_id/article_50_disclosure_ack` plus two unconditional booleans on user JSON and the Ember model |
| 03-03 | `ai-disclosure` component, the one shared `utils/article50_gate.js` helper, additive `labelledBy`/`describedBy` on the `modal-dialog` shell |
| 03-04 | The first-AI-use gate at all three entry points, plus the server backstop |
| 03-05 | Session-entry presentation on both dashboards plus a passive Preferences affordance |

## The decision reviewers should focus on

**Enforcement mode varies by surface; the acknowledgement state is shared.**

| Surface | Mode | Unacknowledged behavior |
|---|---|---|
| Board generation (2 frontend sites + server backstop) | BLOCK | modal blocks, resolves on acknowledge, generation proceeds |
| Eval narration | BLOCK | clinician-initiated, safe to block |
| **Word prediction** | **DEGRADE, never block** | silently skip the AI call, fall back to the non-AI path. No modal, no error, no interruption |

Word prediction fires inline on keystrokes in the speak path. A modal there would interrupt an AAC user mid-communication, which is the one thing this milestone's core value forbids. This is enforced mechanically, not by convention: `grep -c modal app/frontend/app/utils/ai_word_predictor.js` returns 0, and `needsAcknowledgement` never touches `modal.open`, so it cannot reach a dialog transitively either.

Consequence worth understanding: an unacknowledged in-scope user simply gets non-AI word prediction. DEGRADE is a one-way safety valve, so no acknowledgement means no AI interaction, and there is no undisclosed AI use to worry about.

## Safety properties verified against source

- **Inert:** the backstop checks the feature flag first in a short-circuiting `&&`, with a regression spec that applies maximum in-scope pressure (EU user, disclosure unshown) and still asserts success. An active backstop before the modal exists would lock EU and unknown-jurisdiction users out of board generation with no way to unblock themselves.
- **Ack cannot be forged:** `source` and `disclosures_version` are server-side constants, never read from params. Specs assert that posting `source=admin_backfill` and `disclosures_version=999` have zero effect. `allowed?(user, 'edit')` prevents acking on another user's behalf.
- **Fail-safe direction:** gating reads `EuJurisdiction.disclosure_required?` (true for `:eu` AND `:unknown`), never the `jurisdiction` retention column, which deliberately leaves `:unknown` null to protect the HIPAA 6-year floor. These two fail-safes point in opposite directions on purpose.
- **Accessibility:** the modal wraps actions in `.modal_targets` with the bare `.btn` class and is opened with `scannable: true`, so it is reachable by switch scanning. This is deliberate; see the pre-existing findings below.
- **Auto-speak untouched:** the `jump_to_speak` branch in `routes/index.js` is byte-for-byte unchanged.
- **Modal-promise trap guarded:** `modal.open()` resolves a superseded modal's promise with `{replaced: true}`, so a naive `.then()` would fire without user action. Both chained sites use an `onlyIfGenuinelyResolved` guard.

## Disclosure content truthfulness

- Vendors named are exactly Claude Haiku 4.5 (board generation, word prediction) and Claude Opus 4.7 (eval narration), verified against the runtime source. No dev-loop review tooling is named.
- Retention states the tiered rule (24 months general, 12 months rolling for children's data, overridden upward to up to 5 years EU regulatory and up to 6 years under the HIPAA audit floor at 45 CFR 164.316(b)(2), a hard floor), never a flat number.
- Vendor-side retention is stated separately from LingoLinq's own `AiApiLog` retention.

## Deliberately NOT in this PR

Three pre-existing defects were discovered during planning, verified against source, and filed as register findings in #642 rather than repaired here:

- `LL-104bfa61dc` (high): `terms-agree` is unreachable by switch scanning. This PR's component does not inherit that gap, but does not fix `terms-agree` itself.
- `LL-53cb93fab1` (high): `terms-agree` can be silently replaced by `intro` before the user agrees, so `terms_agree` may never be recorded. This PR guards its own Art.50 chain against that bump but does not repair it.
- `LL-58130aaefe` (medium): `modal-dialog` lacks `aria-labelledby`/`aria-describedby` across roughly 140 callers. This PR adds the optional arguments to the shell (verified backward-compatible, since Ember omits bound attributes that are `undefined`) but does not retrofit existing callers.

Also out of scope, per the phase plan: the feature flag registration, RET-01 retention attestation, A11Y-GATE-01, REG-01 (all Phase 5), F3 (deferred 2026-07-14), and any change to Article 50(2) marking, which ships unconditionally and is never flag-gated.

Eval narration has no server-side backstop this phase (client-gate only). Deliberate, and logged as threat T-03-04-07 for the Phase 5 register.

## Rebase note

#644 landed on staging after these plans were written. It migrated component invocation to angle-bracket syntax and added a `npm run lint:hbs` CI gate that fails on newly-introduced `ember-template-lint` violations. Two commits here adapt to it: `modal-container.hbs` uses `<AiDisclosure />`, and `ai-disclosure.hbs` uses `<ModalDialog @arg={{...}}>`. Behavior is unchanged; args map to the same `this.*` properties on the classic component.

## Verification

- Backend: 658 examples, 0 failures, 3 pending (pre-existing, unrelated).
- Frontend: 1766 tests, 1726 pass, 40 skip, 0 fail. Two consecutive clean runs post-rebase on Node 22.
- `npm run lint:hbs` (the new CI gate): clean.
- Phase verifier: 9/9 goal-backward criteria passed, checked against source rather than summary prose.

**Flake disclosure:** the first post-rebase full frontend run showed 1 failure that did not reproduce across two subsequent full runs. The log was truncated before the failing test name was captured, so it is recorded here as an unidentified flake rather than attributed to the known `persistence-sync` race. If CI shows an intermittent single failure, this is the prior history.

https://claude.ai/code/session_01GUynFU8HHfATmtsSBh1q9b


swahlquist:
## Compliance PR Checklist

swahlquist:
## Codex Review

**Verdict:** NEEDS_HUMAN
**Route:** codex
**Head:** 30e141f4513428d4a12e5446e2b2f4e42786addf
**Run:** https://github.com/lingolinq/LingoLinq-AAC/actions/runs/29816891162

Codex review completed.

| --- | --- | --- | --- |
| CR-1 | HIGH |  | The Codex sandbox prevented all required read-only inspection commands from running, so I could not verify the PR body claims, enumerate server entry points, run drift checks, inspect file modes, grep docs/legal and audit-reports, or confirm local head state. |

swahlquist:
## Review fixes pushed (5 commits)

Addresses the local review and the Codex relay. Verified: backend **663 examples / 0 failures** (was 658, +5 new specs), frontend **1766 tests / 1726 pass / 40 skip / 0 fail**, `lint:hbs` clean, `lint:js` unchanged from baseline on all touched files, all four artifact-integrity checks green.

### Blocker: board-generation gate destroyed its own modal

`generate-board` is modal-hosted (`modal-container.hbs:161`, opened from `new-board.js#generateWithAi`) and `modal.open()` replaces the showing modal. The gate fired from inside `generateLabels` tore down its own host: user types a prompt, clicks Generate, the disclosure replaces the form (input lost), they acknowledge, and the continuation bails on `isDestroyed`. No modal, no generation, no error.

Gate moved to `new-board.js#generateWithAi`, before the form exists. `generate-board.js` now carries no client gate; the server backstop enforces that endpoint. `create-board-new` and `eval-comprehensive-runner` are route pages, not modals, so they gate in place safely.

### Eval narration: gate stays on the clinician, deliberately

Contrary to the review suggestion. `EvalNarrator#payload_for_prompt` drops the student name (any key casing) and the etiology/diagnosis field before egress, then `PiiScrubber` runs with the resolved student's name blocklisted. Two different questions:

- *May this student's data be processed by AI?* The student. Enforced server-side by `ai_allowed_for?`, already correct.
- *Has the human interacting with this AI been told it is an AI?* Art. 50(1) informs the person **interacting**. That is the SLP at the keyboard.

Recording an acknowledgement against a student who never saw the notice would need a new `ARTICLE_50_DISCLOSURE_SOURCES` entry and would make the audit trail less truthful, not more. Reasoning is now in the code so it is not re-litigated.

### Disclosure text corrected

The notice said "clinical evaluation notes and assessment data" go to Claude Opus, and was silent on the name and diagnosis being stripped first. `slp_notes` does go, so that was not false, but the omission **overstates** exposure to exactly the audience the notice exists to inform. Now states the removals, with truthfulness gate 5 asserting both halves so the text and `payload_for_prompt` cannot decouple.

### Also fixed

| Item | Change |
|---|---|
| N+1 | `article_50_disclosure_*` scoped to `permissions['model']`. It sat past the `limited_identity` branch, putting an uncached org lookup on every list row (org pages paginate to 500). New spec asserts `EuJurisdiction` is never consulted on the list path. |
| Preferences trap | "View Notice Again" opened the uncloseable BLOCK modal; only exit was Acknowledge, with no cancel target for a switch-scanning user. Now a plain link to the notice. |
| Hanging promise | A bumped gate rejects with `GATE_NOT_ACKNOWLEDGED` instead of never settling. Both call sites surface a reason. |
| Dead wiring | `ai-disclosure`'s `onClose`/`onOpening`/`onClosing` were assigned after render as non-tracked properties, so the binding never updated. Removed. |

Version and URL single-sourced as `ART50_DISCLOSURE_URL` in `utils/article50_gate`. `03-UI-SPEC.md` §7.3 and the file table are corrected in place with the reasons, since both specified behavior that turned out to be wrong.

### Not addressed

Phase 3's `ai_disclosure_*` frontend strings are still absent from `public/locales/*.json`, so non-English locales fall back to English for the modal chrome. The notice body itself is translated (`config/locales/en.yml` + `es.yml`). Pre-existing on this branch, not introduced here; wants one `i18n_generator.rb` pass across the whole Phase 3 string set rather than a partial edit.

swahlquist:
## Codex Review

**Verdict:** NEEDS_HUMAN
**Route:** codex
**Head:** 08f144d966f21164afe2cdb417cf70edb962fa0d
**Run:** https://github.com/lingolinq/LingoLinq-AAC/actions/runs/29862444317

Codex review completed.

| --- | --- | --- | --- |
| CR-1 | HIGH |  | Required read-only inspection commands could not run in the Codex sandbox, so I could not complete the mandated blocker-first review checklist. |

swahlquist:
## Locale gap fixed (not deferred)

It was worse than locale-blind: the **Spanish notice shipped and was unreachable**. `config/locales/es.yml` has the full translation, `disclosures_controller#show_art50` already wraps the render in `I18n.with_locale` — but it reads `params[:locale]`, and nothing on the client ever sent one. Every request resolved to `I18n.default_locale` and served English.

Both halves were broken, so both are fixed:

- **Client** — `art50DisclosureUrl()` appends `i18n.langs.preferred` and drives all three consumers (the in-modal fetch, the "Read the Full Notice" link, and the Preferences link) so they cannot disagree. Skips the `backwards` debug pseudo-locale, percent-encodes the value.
- **Server** — `resolve_locale` now falls back through the primary subtag before giving up (`es-ES` → `es`, `es_MX` → `es`), mirroring the frontend's own `preferred`/`fallback` pair in `initializers/attempt_lang.js`. Browsers report a full regional tag, so the client fix alone would still have missed the allowlist and served English. Still an allowlist: the derived subtag is checked against `I18n.available_locales` exactly like the full tag, so an unknown base language degrades to the default.

Eight new tests (3 server, 5 client) including the `es-ES` and `es_MX` cases that were the actual failure mode.

**Verified:** backend **666 examples / 0 failures**, frontend **1771 tests / 1731 pass / 40 skip / 0 fail**, `lint:hbs` clean, `lint:js` unchanged from baseline on all touched files.

One genuine failure surfaced during this and is fixed: the `ai-disclosure` fetch test asserted an exact URL string. It now asserts the path plus the presence of the locale param, since the value varies with the test browser.

### Eval narration: confirmed staying on the clinician

Per Scot, and the reasoning is in the code at `eval-comprehensive-runner.js`. The two questions are separate: *may this student's data be processed by AI* is the student's, enforced server-side by `EvalNarrator.ai_allowed_for?` plus de-identification before egress; *has the human interacting with this AI been told it is an AI* is the SLP's. Recording an ack on a student record for a notice the student never saw would be worse audit evidence.

### Remaining known gap (logged, not a blocker)

Phase 3's `ai_disclosure_*` frontend strings are absent from `public/locales/*.json`, so the modal chrome (buttons, status text, errors) falls back to English in non-English locales. The **notice body itself is now correctly localized**, which is the Article 50 delivery obligation. The chrome wants one `i18n_generator.rb` pass over the whole Phase 3 string set rather than a partial hand-edit — Phase 5.

swahlquist:
## Codex Review

**Verdict:** NEEDS_HUMAN
**Route:** codex
**Head:** ac0407899f0538c446b97246c2bcc18e9145a780
**Run:** https://github.com/lingolinq/LingoLinq-AAC/actions/runs/29866524333

Codex review completed.

| --- | --- | --- | --- |
| CR-1 | HIGH |  | Required read-only inspection commands cannot run in the Codex sandbox, so the mandated review checklist cannot be completed against HEAD. |

PR #656, opened 2026-07-22
compliance(art50): Phase 5 - rollout flag (AVAILABLE-only), retention reconciliation, a11y

swahlquist:
## Art50 Phase 5 - Rollout, Retention & Monitoring

Finishes the EU AI Act Article 50 work (hard gate **2026-08-02**). Targets `staging`.
The disclosure feature stays **inert** until the flag is enabled; **nothing here goes live in production**.

Phases 1-4 (consent record, disclosure content, accessible modal + ack endpoint + first-AI-use gate, jurisdiction stamping) are already merged to staging. This PR does the rollout wiring, retention reconciliation, an accessibility pass, and the register/claim-surface truth-up.

---

### What is ACTIVE on staging after this PR
- The `article_50_disclosure` frontend flag is **registered in `AVAILABLE_FRONTEND_FEATURES` only** (RLL-01). It is OFF for everyone by default.
- Verified path to the client: `FeatureFlags.frontend_flags_for(user)` -> `lib/json_api/user.rb:94` -> `user.js:135 feature_flags: attr('raw')` -> `app-state.js:2538 computed('currentUser.feature_flags')` -> `appState.feature_flags.article_50_disclosure`, the single input `utils/article50_gate.js#needsAcknowledgement` reads.
- EU 5-year `AiApiLog` purge is now **functional** wherever Phase 4 is deployed (jurisdiction is stamped `EU` at the three AI call sites).
- Accessibility fix on the modal itself: `role="alert"` on the ack-failure message (WCAG 4.1.3) and a single stable button element across the error toggle (WCAG 2.4.3).

### What remains INERT until the flag is enabled
- Because the flag is AVAILABLE-only (not in `ENABLED_FRONTEND_FEATURES`), `needsAcknowledgement` returns `false` for everyone, so the modal is shown to **no user**. The entire Phase 3/4 disclosure path is dormant.
- Enabling the flag for EU accounts is the **hard 2026-08-02 release gate**, done **only on Scot's explicit sign-off** and **only after the production deploy** of Phases 3-5. This PR does not enable it.

### Retention: enforced vs decided/documented (RET-01)
Reconciled identically across `docs/legal/DATA_RETENTION.md`, `docs/legal/AI_DATA_FLOW_CLASSIFICATION.md`, `lib/tasks/scheduler.rake`, and `privacy.hbs`:

| Tier | Window | Status |
|---|---|---|
[...]

PR #652, opened 2026-07-22
docs(compliance): record GCP cutover posture

swahlquist:
## Summary
- updates infrastructure and compliance docs for the 2026-07-22 Gate 1 DNS cutover to GCP
- records Google Cloud Platform as active production hosting and Render as a write-frozen rollback fallback pending explicit decommission
- refreshes derived document-register hashes for the changed git compliance docs

## Finding / attestation boundary
- does not close, downgrade, or supersede any finding
- leaves LL-6619cc1811 open until the in-context Cloud Run rediss:// PONG is captured and Scot attests closure
- leaves LL-f150e0e828 open until the offboarding consent flow is functionally tested and Scot closes it
- keeps LL-aacae48768 / LL-7f7372e3eb as Render-tail items until Render fallback data and services are deleted or restricted
- marks the Compliance Program content as post-Gate-1 update pending Scot re-attestation

## Verification
- scripts/regenerate-register.sh
- git diff --check
- targeted stale-current-state sweep for Render/GCP planned-hosting wording
- confirmed FINDINGS.json was unchanged and named findings retain their existing statuses

## Notes
The historical 2026-06-18 status snapshot and older change-log rows still describe the pre-cutover state by design; they are not current-state operator instructions.

PR #648, opened 2026-07-21
compliance(register+subprocessors): disclose 5 raw localization/speech egress sinks; open LL-c38e7da48e + LL-1eb9a2435b

swahlquist:
## What this PR does
Corrects the subprocessor register to reflect the **actual** localization/speech egress surface, discovered while scoping which features truly reach a third party. None of this is runtime-behavior change; it is register + disclosure accuracy.

Five runtime flows send **raw, unscrubbed user content** to a third party (none pass through `lib/pii_scrubber.rb`):

| # | Sink | Data | Code | Contract basis |
|---|------|------|------|----------------|
| 15 | Google Cloud Translation | raw button-label text | `word_data.rb#query_translations` | Google Cloud BAA (§5.7); covered-service TBC |
| 16 | Google Cloud Text-to-Speech | raw spoken text | `tts.rb#generate_google`, `search_controller` | Google Cloud BAA; covered-service TBC |
| 17 | Abair Irish TTS (Trinity College Dublin) | raw spoken text | `tts.rb#generate_irish` | **No DPA on file** |
| 18 | Google Cloud Speech-to-Text | **raw user VOICE RECORDING (audio)** | `button_sound.rb#schedule_transcription` (auto, `after_save`) | Google Cloud BAA; covered-service TBC |
| — | consumer `translate_tts` | raw text | `search_controller.rb:408` | **No basis; dead in prod** (only hit when `GOOGLE_TTS_TOKEN` unset) |

Sink **#18 was missed by the first pass** and is the highest-sensitivity: an actual voice recording, uploaded automatically, that **cannot be PII-scrubbed** (the scrubber redacts text, not audio). Its discovery corrected my earlier "build a text identifier allowlist" recommendation — that control can't touch audio, and for the text flows the real control is BAA covered-service confirmation, not scrubbing.

## Findings (all OPEN — only Scot triages/closes)
- **LL-c38e7da48e** (medium) — Cloud Translation, text
- **LL-a167848115** (medium) — TTS (Google TTS + Abair), text
- **LL-1eb9a2435b** (medium, NEW) — Cloud Speech-to-Text, voice audio
- Cross-referenced as one triage cluster.

## Fix status
| Item | Status | Evidence |
| --- | --- | --- |
| Disclose all 5 sinks in the register | Fixed | `SUBPROCESSORS.md` rows #15–#18 + §5.8 |
| Correct control framing (BAA covered-service, not scrubber) | Fixed | `SUBPROCESSORS.md` §5.8 |
[...]

PR #631, opened 2026-07-19
compliance(anthropic): record executed HIPAA-Ready BAA + reconcile posture docs

swahlquist:
Records the executed Anthropic HIPAA-Ready BAA and reconciles the compliance docs that previously said the Anthropic runtime path had no BAA. Docs-only; no code or data-flow change.

## What happened
Anthropic HIPAA-Ready BAA executed and HIPAA readiness enabled **2026-07-18** on the runtime-dedicated **LingoLinq, LLC** Anthropic API org (confirmed by Scot). This is the model-provider BAA for the direct Anthropic runtime egress path, distinct from the AWS and GCP infrastructure BAAs already on file.

## Live enablement verification (2026-07-18)
Run against the deployed runtime `ANTHROPIC_API_KEY` (key value never logged):
- `POST /v1/messages` (Haiku 4.5) -> **HTTP 200** (eligible feature works, no ZDR required)
- `GET /v1/files` (non-eligible) -> **HTTP 400** "this endpoint is not available for HIPAA-regulated organizations without Zero Data Retention"

The 400 on a non-eligible feature is positive proof the runtime key is on a HIPAA-regulated org and org-wide enforcement is active. This resolves the key-identity blocker from the 2026-07-18 eligible-services scope review.

## Changes
- **New:** `docs/legal/ANTHROPIC_BAA_ACCEPTED.md` (parity with AWS/GCP records) + `DOCUMENT-REGISTER` row `DOC-ab3a8c3ed4`.
- **Reconciled (P4):** `COMPLIANCE.md` Anthropic row, `SUBPROCESSORS.md` row 4 + section-5.7 narrative + changelog, `GCP_BAA_ACCEPTED.md` scope note, `AI_GOVERNANCE_MEMO.md` (dated addendum, flagged for CEO re-attestation).
- Coverage recorded: Messages API only, in-scope models (Haiku 4.5 / Opus 4.7 / Sonnet), **no ZDR required**; excluded features (Files/Batch/Skills/Code-Exec/Computer-Use/Web-Fetch) blocked org-wide. Source: https://platform.claude.com/docs/en/manage-claude/api-and-data-retention

## Fix status
| Item | Status | Evidence |
| --- | --- | --- |
| Record executed Anthropic BAA | Fixed | `docs/legal/ANTHROPIC_BAA_ACCEPTED.md` |
| Register row + integrity | Fixed | `DOC-ab3a8c3ed4`; `document-register-render --check` OK (55 docs) |
[...]

PR #669, opened 2026-07-23
compliance(register): pin attested bytes with attestedContentHash (closes #661)

### LINKED ISSUE #661: compliance(register): add attestedContentHash so attested-file drift is detectable
Follow-up from #658.

## Problem

`audit-reports/DOCUMENT-REGISTER.json` stores a row's current `contentHash` and its `attestation.attestedDate`, but never the hash **as attested**. So an attested document's bytes can change with no signal: the render simply recomputes `contentHash` and CI stays green, while the row continues to assert an attestation that covered a revision which no longer exists.

This is not hypothetical. Two instances were found by hand during #658, and both passed green CI:

1. `docs/legal/COMPLIANCE_PROGRAM_OVERVIEW.md` - attested for **external release** 2026-07-09 (banner still reads `rev. 2026-07-09-c`), then edited by #649 and #652 on 2026-07-21. The subprocessor and hosting posture was rewritten and the COPPA offboarding section inverted from an open gap to an implemented control.
2. `docs/legal/AI_DATA_FLOW_CLASSIFICATION.md` - attested (provisional) 2026-07-09, then edited by #656 on 2026-07-22, moving two AI-log retention tiers from "Decided, rolling out" to "Decided, not yet enforced."

Both rows were held at `status: draft` with empty attestations in #658 as a manual correction. Nothing prevents the next occurrence.

The Drive-side twin of this hole is already documented in `meta.contentHashNote` (Drive hashes are operator-supplied with no automated refresh). This issue is about the **git** side, where the bytes *are* verifiable and the drift *is* mechanically detectable.

## Proposed fix

Add `attestation.attestedContentHash` alongside `attestedBy` / `attestedDate`, populated when an attestation is recorded. Then in `scripts/document-register-render.rb --check`, for any git row that is attested:

- `attestedContentHash` missing while `attestedBy` is present -> failure (an attestation with no pinned bytes cannot be verified)
- `attestedContentHash != contentHash` -> failure, naming the file and telling the operator that the attested revision no longer exists and re-attestation is owed

Scope decisions to settle in the PR:

- Whether an unattested-but-drifted row should be advisory rather than a hard failure.
- Whether drive/notion rows are exempt (they have no verifiable hash, so probably yes: enforce only where `canonicalSystem == git`).
- Backfill: existing attested git rows have no `attestedContentHash`. Backfilling from the current bytes would silently re-assert every existing attestation, which is the exact error this is meant to prevent. Safer to require Scot to confirm each, or to grandfather existing rows behind a dated exemption list that shrinks to zero.

Only Scot attests, so the backfill question is his call, not the implementer's.

## Also owed (separate from the schema work)

Re-attestation of the two documents above, the second being externally shareable.

swahlquist:
Closes #661.

## The hole

`audit-reports/DOCUMENT-REGISTER.json` stored a row's current `contentHash` and its
`attestation.attestedDate`, but never the hash **as attested**. An attested compliance document
could therefore be rewritten with a green build: the render recomputed `contentHash`, and the row
went on asserting an attestation that covered a revision which no longer existed. Both known
instances were caught by hand, never by CI.

## What this adds

`attestation.attestedContentHash` pins the sha256 of the bytes that were attested.
`ruby scripts/document-register-render.rb --check` now fails when:

| Condition | Why it fails |
|---|---|
| pinned hash != `contentHash` (git row) | the attested revision no longer exists; re-attestation is owed |
| attested git row pins nothing and is not grandfathered | an attestation with no pinned bytes cannot be verified |
| pin is not 64 lowercase hex | malformed |
| non-git row carries the field | Drive/Notion hashes are operator-supplied; pinning asserts a guarantee a network-free check cannot make |
| exemption is stale, duplicated, unresolvable, undated, or reasonless | the grandfather list must only ever shrink |

**Render never writes `attestedContentHash`.** Backfilling it in the same pass that recomputes
`contentHash` would make every attestation self-certifying, which is precisely the failure the
field exists to catch. Pinning is a human act recorded with the attestation.

## Backfill: decided by evidence, not by policy

The issue flagged backfill as Scot's call because populating from current bytes would silently
re-assert every attestation. Rather than do that or punt entirely, each of the 17 attested git rows
was classified by whether git shows a **modify** commit after its `attestedDate`. A commit that
*added* the file after that date is not drift (the attestation preceded the merge), which cleared
three rows a naive date compare had flagged.

- **10 rows with no post-attestation modification** were pinned. Git proves the current bytes are
  the attested bytes, so the pin states a fact rather than making an attestation.
- **7 rows modified after attestation** were **not** pinned. They sit on
  `meta.attestationBackfillExemptions` with the commits that moved them.

### Re-attestation owed (7) - Scot's queue, no agent decision made

| Record | Attested | Modified after by |
|---|---|---|
| `COMPLIANCE.md` | 2026-07-06 | #622, #631, #652 |
| `docs/legal/COMPLIANCE_POSTURE_REPORT.md` | 2026-07-16 | #622, #652 |
| `docs/legal/DATA_RETENTION.md` | 2026-06-21 | #569, #656 |
| `docs/legal/SUBPROCESSORS.md` | 2026-07-06 | #588, #622, #631, #648, #652 |
| `docs/legal/COPPA_VERIFICATION_2026-04-26.md` | 2026-06-21 | #597 |
| `docs/legal/PARENTAL_CONSENT_EMAIL.md` | 2026-06-21 | #597, #616 |
| `docs/legal/GCP_BAA_ACCEPTED.md` | 2026-07-16 | #631, #652 |

That list is the substantive finding of this PR: seven attested compliance records, several of them
externally shareable, whose attested revision no longer exists. Nothing here re-attests them.

## Downstream truthfulness

- `compliance-publication-status.rb` gains an **Attestation Integrity** section with a drift table
  and a re-attestation queue, and counts both in the footer.
- `document-register-notion-sync.rb` labels an unverified attestation
  `(re-attestation owed: attested bytes not pinned or no longer current)` rather than showing a bare
  `attestedBy attestedDate`, so the Notion mirror cannot read cleaner than the register.
- `docs/legal/README.md` rule 4 states the pin, which is what makes the existing
  "attestation freezes the artifact" rule enforceable rather than aspirational.

## Fix status

| Item | Status | Evidence |
|---|---|---|
| `attestedContentHash` schema + pin | Fixed | `audit-reports/DOCUMENT-REGISTER.json` `meta.attestationHashNote`; 10 rows pinned |
| Drift is a hard CI failure (git rows) | Fixed | `scripts/document-register-render.rb` `attestation_problems` |
| Grandfather list cannot become permanent cover | Fixed | `attestation_exemption_problems`; stale entry fails `--check` |
| Non-git rows exempt from enforcement | Fixed (by design) | field rejected on drive/notion rows |
| Render cannot launder a pin | Fixed | negative test 13: edit file, full render, `--check` still fails |
| Backfill of drifted rows | Not fixed by design | 7 rows queued for Scot; only Scot attests |

## Not covered by this PR

- **Re-attestation of the 7 drifted records.** Only Scot attests. Each clears by pinning the hash
  and deleting its exemption entry.
- **Drive and Notion rows.** Their hashes are operator-supplied with no automated refresh, so there
  is nothing offline CI can verify. Documented in `meta.contentHashNote` and unchanged here.
- **Historical reconstruction.** No attempt is made to recover what the attested bytes *were* for a
  drifted row; the register records that they are gone, not what they said.

## Verification

13 negative tests, every one confirmed to fire: pinned-hash mismatch, unpinned + unexempted, stale
exemption, malformed pin, pin on a drive row, unresolvable exemption id, missing reason, unparseable
`addedOn`, duplicate exemption, exemption on an unattested row, exemption location drift, non-array
list, and the end-to-end path (edit an attested file, run the full render, confirm `--check` still
fails).

Preflight, all green at head:

```
ruby scripts/document-register-render.rb --check       # OK
ruby scripts/compliance-publication-status.rb --check  # OK
ruby scripts/compliance-notion-publish.rb --check      # OK
ruby scripts/compliance-calendar-render.rb --check     # OK
ruby scripts/capability-check.rb --check               # OK
scripts/regenerate-register.sh                         # Done, no further changes
git diff --check                                       # clean
```

The advisory about non-git rows missing supplied hashes is expected. One new advisory is
intentional and prints every run until it reaches zero: `7 attested row(s) are grandfathered on
meta.attestationBackfillExemptions`.

## Author-Model: opus-4.8

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -141,6 +141,9 @@ jobs:
       - name: Verify document register render + git content hashes + bundle completeness
         run: ruby scripts/document-register-render.rb --check
 
+      - name: Verify attestedContentHash guards (drift + closed exemption set)
+        run: scripts/tests/attestation-hash-guard-test.sh
+
       - name: Verify compliance publication status report
         run: ruby scripts/compliance-publication-status.rb --check
 
--- a/audit-reports/COMPLIANCE-PUBLICATION-STATUS.md
+++ b/audit-reports/COMPLIANCE-PUBLICATION-STATUS.md
@@ -199,6 +199,26 @@ No retention rule has been approved. Nothing in this register is eligible for di
 
 _No record is under legal hold. A hold suspends all disposition for the rows it covers, and only Scot flips it._
 
+## Attestation Integrity
+
+17 attested git record(s). `attestation.attestedContentHash` pins the bytes that were attested; `ruby scripts/document-register-render.rb --check` fails when a pinned hash stops matching the file. Drive and Notion rows are out of scope: their hashes are operator-supplied, so there is nothing CI can verify.
+
+**No pinned attestation has drifted.** Every record that pins a hash still matches the attested bytes.
+
+### Re-attestation queue (7)
+
+Attested before the check existed and modified afterwards, so the attested revision no longer exists. The hash is deliberately not backfilled: pinning current bytes would re-assert an attestation Scot never gave. Each row clears when Scot re-attests the current revision.
+
+| Title | Location | Attested | Why it is unpinned |
+|---|---|---|---|
+| COPPA Final-Rule Verification | `docs/legal/COPPA_VERIFICATION_2026-04-26.md` | 2026-06-21 | attested 2026-06-21, then modified by 1 later commit(s); the attested revision no longer exists, so pinning current bytes would re-assert an attestation Scot never gave |
+| Compliance & Data Governance (COMPLIANCE.md) | `COMPLIANCE.md` | 2026-07-06 | attested 2026-07-06, then modified by 3 later commit(s); the attested revision no longer exists, so pinning current bytes would re-assert an attestation Scot never gave |
+| Compliance Posture Report | `docs/legal/COMPLIANCE_POSTURE_REPORT.md` | 2026-07-16 | attested 2026-07-16, then modified by 2 later commit(s); the attested revision no longer exists, so pinning current bytes would re-assert an attestation Scot never gave |
+| Data Retention Schedule | `docs/legal/DATA_RETENTION.md` | 2026-06-21 | attested 2026-06-21, then modified by 2 later commit(s); the attested revision no longer exists, so pinning current bytes would re-assert an attestation Scot never gave |
+| Google Cloud Platform BAA + CDPA + SCCs - Acceptance Record | `docs/legal/GCP_BAA_ACCEPTED.md` | 2026-07-16 | attested 2026-07-16, then modified by 2 later commit(s); the attested revision no longer exists, so pinning current bytes would re-assert an attestation Scot never gave |
+| Parental Consent Email (COPPA / under-13) | `docs/legal/PARENTAL_CONSENT_EMAIL.md` | 2026-06-21 | attested 2026-06-21, then modified by 2 later commit(s); the attested revision no longer exists, so pinning current bytes would re-assert an attestation Scot never gave |
+| Subprocessor Register | `docs/legal/SUBPROCESSORS.md` | 2026-07-06 | attested 2026-07-06, then modified by 5 later commit(s); the attested revision no longer exists, so pinning current bytes would re-assert an attestation Scot never gave |
+
 ## Supersession Chains
 
 A superseded record is never edited, renamed, or moved. It keeps its row and its membership in any frozen point-in-time binder; only the pointer is added.
@@ -258,4 +278,4 @@ The missing layer is a Google Docs publisher/refresh workflow. Until that exists
 
 ---
 
-_70 documents tracked. 53 stale review item(s). 22 Drive refresh item(s). 4 Notion hash item(s). 22 inferred retention class(es). 0 legal hold(s). 1 superseded record(s). 23 bundle gap(s) across 6 bundle(s)._
+_70 documents tracked. 53 stale review item(s). 22 Drive refresh item(s). 4 Notion hash item(s). 22 inferred retention class(es). 0 legal hold(s). 1 superseded record(s). 0 drifted attestation(s), 7 awaiting re-attestation. 23 bundle gap(s) across 6 bundle(s)._
--- a/audit-reports/DOCUMENT-REGISTER.json
+++ b/audit-reports/DOCUMENT-REGISTER.json
@@ -451,7 +451,166 @@
     },
     "bundleMatchNote": "bundleDefinitions[*].requiredDocs binds each requirement to a specific document by canonicalLocation (identity), not free-text title, so a doc retitled to a required string cannot satisfy a requirement. Each requiredDoc carries a human-readable title that --check verifies still matches the member at that location (register-drift guard). --check fails if a requiredDoc location is not a bundle member, or its title has drifted.",
     "bundleGapNote": "bundleDefinitions[*].gaps names artifacts a bundle genuinely needs that DO NOT EXIST as records yet. A gap is a deliberate, visible hole: it is rendered in DOCUMENT-REGISTER.md and never satisfied by inventing a row or by pointing a requiredDoc at an unrelated document. gaps never fail --check; an unmet requiredDoc always does. Close a gap by creating the real artifact and promoting it out of gaps into requiredDocs.",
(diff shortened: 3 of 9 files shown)


swahlquist:
## Codex Review

**Verdict:** REQUEST_CHANGES
**Route:** codex
**Head:** 72087d54889198e6c9f4f35ece9c3b94ff0925a7
**Run:** https://github.com/lingolinq/LingoLinq-AAC/actions/runs/29982860445


Codex review completed.

| --- | --- | --- | --- |
| CR-1 | HIGH | scripts/document-register-render.rb | Backfill exemptions can be added for future unpinned attestations, so the grandfather list is not actually constrained to pre-existing drift and does not 'only ever shrink'. |

swahlquist:
Both findings addressed at `18cc95531`.

## High: the exemption list was a waiver mechanism

Correct, and it defeated the point of the gate. Appending an id to `meta.attestationBackfillExemptions` would have cleared any future unpinned attestation, so "one-way, only shrinks" was a property of the prose, not of the code.

Fixed by freezing the grandfather set **in code, outside the data the check reads**:

```ruby
CLOSED_ATTESTATION_EXEMPTIONS = %w[
  DOC-9b299a785b DOC-4e3b7fb1fb DOC-bff9acf51f DOC-0387973005
  DOC-407d2c2bf4 DOC-4e6c9253b9 DOC-5b14b08908
].to_set.freeze
```

`meta.attestationBackfillExemptions` may now only ever be a **subset** of that set. An id outside it is rejected no matter how well-formed its entry is:

```
[FAIL] attestation exemption for "..." (DOC-...) is not in the closed grandfather set; that set is
frozen in scripts/document-register-render.rb and can only shrink. An attestation recorded after
this check landed must pin attestedContentHash - it cannot be exempted
```

Removing an id (Scot re-attested, hash pinned) is the only supported edit. Nothing is ever added. The set can be closed because the population it describes is closed: rows attested before the check existed. A new attestation pins its bytes at the moment it is recorded, which costs the attester nothing, since they are looking at the file.

The unpinned-row message also no longer offers an exemption as a remedy. It now says the list is closed and that adding an entry will not clear the failure, so the next person hitting it is not pointed at the loophole you found.

## Medium: tests were run but not committed

Also correct. Added `scripts/tests/attestation-hash-guard-test.sh`, wired into the `audit-artifacts-integrity` CI job. 16 cases, all passing:

```
  ok   live register passes unmutated
  ok   pinned hash no longer matches the file
  ok   attested git row pins nothing and is not grandfathered
  ok   pin is malformed
  ok   non-git row carries a pin CI cannot verify
  ok   a new row cannot be exempted into passing
  ok   exemption left behind after the row was pinned
  ok   exemption id resolves to nothing
  ok   exemption carries no reason
  ok   exemption date is unparseable
  ok   exemption is duplicated
  ok   exemption covers a row with no attestation
  ok   exemption location disagrees with its row
  ok   exemption list is not an array
  ok   edit an attested file, re-render, guard still fires
  ok   live register untouched
```

That covers every case you listed: pinned mismatch, missing pin, stale exemption, non-git pin, malformed hash, and the render refusing to backfill. The last is the end-to-end path (edit an attested file, run the full render, confirm `--check` still fails), which is what would catch a future change that made attestations self-certifying.

The harness works on a **copy** of the real register placed beside it in `audit-reports/`, not a synthetic fixture. The properties under test are properties of the real register plus the real repo (tracked file bytes, `docs/legal` completeness, `REPO_ROOT` containment), and a fixture faking all of that is what would rot first. Two guards keep it honest: a baseline case asserting the committed register passes unmutated, and a post-condition asserting the run left the live register untouched.

Supporting change: the render path is now derived from the register filename (`DOCUMENT-REGISTER.json` to `DOCUMENT-REGISTER.md`) rather than hard-coded, so a test register renders beside itself instead of comparing against and overwriting the real one. Identical result for the default path.

## Unchanged

The seven rows owing re-attestation stay unpinned and queued. Nothing here attests anything; only Scot attests.

swahlquist:
## Codex Review

**Verdict:** APPROVE
**Review status:** Approved
**Route:** codex
**Head:** 18cc9553126edda5734a31007e7f66e57fa8a2e2
**Run:** https://github.com/lingolinq/LingoLinq-AAC/actions/runs/29983509650


Codex review completed.

_No findings reported._

PR #667, opened 2026-07-23
security(tts): GA v1 Text-to-Speech endpoint + drop unauthenticated consumer fallback (BAA Pre-GA exclusion)

swahlquist:
## What & why
BAA-coverage hardening for the Google Cloud localization/speech egress surface documented in **#648** (findings **LL-a167848115**, **LL-c38e7da48e**, **LL-1eb9a2435b**).

Google's Cloud BAA covers Text-to-Speech as a **product**, but its canonical HIPAA page explicitly **excludes Pre-GA offerings from PHI**: *"Do not use Pre-GA offerings ... in connection with PHI"* ([cloud.google.com/security/compliance/hipaa](https://cloud.google.com/security/compliance/hipaa)). All six TTS call sites used the Pre-GA **`v1beta1`** endpoint, so that traffic sat outside BAA coverage.

## Changes
1. **Repoint TTS `v1beta1` → GA `v1`** — `lib/tts.rb` (x3) and `app/controllers/api/search_controller.rb` (x3). Verified drop-in against Google's v1 REST reference: `v1/text:synthesize` takes the same `input`/`voice`/`audioConfig`, and `v1/voices` takes `languageCode`. Translation (`/v2`) and Speech-to-Text (`/v1`) were already GA and are untouched.
2. **Remove the unauthenticated consumer `translate.google.com/translate_tts` fallback** — no DPA/BAA basis; must never receive user utterance text. Only reachable with `GOOGLE_TTS_TOKEN` unset (dead in prod, where it's set). Now returns `400 "no tts provider configured"`.
3. **Abair Irish TTS left untouched** — separate open CEO decision (DPA vs disable).

## Fix status
| Item | Status | Evidence |
| --- | --- | --- |
| TTS on GA `v1` endpoint | Fixed | `lib/tts.rb`, `search_controller.rb` (no `v1beta1` remains) |
| Consumer `translate_tts` removed | Fixed | `search_controller.rb` audio `else` → 400 |
| Regression tests | Fixed | `spec/lib/tts_spec.rb` (new) + audio specs; 62 examples, 0 failures |
| Abair DPA/disable | Not in scope | open CEO decision (LL-a167848115) |

## Not covered
- Abair Irish TTS DPA-or-disable (CEO decision).
- Voice-transcription (#18) consent/flag posture (register documents it; separate decision).

## Test evidence
`bundle exec rspec spec/lib/tts_spec.rb spec/controllers/api/search_controller_spec.rb` → **62 examples, 0 failures**.

[...]

PR #671, opened 2026-07-23
test(env_wrap): scope ENV overrides to the block + faithful restore

swahlquist:
## What & why
Tiny test-isolation fix for the `env_wrap` spec helper (flagged during the TTS security work, #667). Helper-only; no product code.

Two defects:
1. **Leak to siblings** — `env_wrap` attached its `before`/`after` hooks to the **enclosing** example group, so the ENV overrides applied to every sibling example in that group, not just the ones inside the block. (This is why a `GOOGLE_TTS_TOKEN` override leaked into a sibling "no token" test in `search_controller_spec`, which had to work around it.)
2. **Stale restore** — it snapshotted ENV once at spec-definition (file-load) time and "restored" an originally-**unset** key by setting it to `nil`.

## Fix
- Nest the block's examples in their own group so overrides are scoped to them only.
- Capture originals at **run** time in an `around(:each)`.
- **Delete** keys that were originally unset on teardown instead of leaving them set.

## Verification
All 10 `env_wrap`-using spec files pass: **545 examples, 0 failures**. No test relied on the old leak.

## Fix status
| Item | Status | Evidence |
| --- | --- | --- |
| Overrides scoped to block | Fixed | `spec/spec_helper.rb` (nested group + `around`) |
| ENV restored faithfully (delete-on-unset, runtime capture) | Fixed | `spec/spec_helper.rb` |
| No regressions | Fixed | 10 files, 545 examples, 0 failures |

## Author-Model: opus-4.8

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -99,15 +99,34 @@ RSpec.configure do |config|
 end
 
 def env_wrap(overrides, &block)
-  fallbacks = {}
-  overrides.each{|k, v| fallbacks[k] = ENV[k] }
-  before(:each) do
-    overrides.each{|k, v| ENV[k] = v }
+  # Run `block`'s examples with the given ENV overrides, then restore the prior state.
+  #
+  # Two isolation defects are fixed here vs the old implementation:
+  #   1. Scoping: the overrides now apply ONLY to the examples defined inside `block`,
+  #      by nesting them in their own example group. The old version attached its
+  #      before/after hooks to the *enclosing* group, so the overrides leaked into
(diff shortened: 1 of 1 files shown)


swahlquist:
## Codex Review

**Verdict:** APPROVE
**Review status:** Approved
**Route:** codex
**Head:** b6ab3f6fe38af087c0d82f36a18dc9214ad0073d
**Run:** https://github.com/lingolinq/LingoLinq-AAC/actions/runs/29987583479


Codex review completed.

_No findings reported._

PR #689, opened 2026-07-26
fix(word_data): cache word-frequency ranks instead of downloading a 1.7MB list per record

swahlquist:
## What and why

`WordData#assert_missing_priority` (`app/models/word_data.rb:15`) schedules the **instance** `assert_priority` for every record saved without a priority. With no `opts`, that path downloaded the entire ~1.7MB `language/english_with_counts.txt` from S3 to resolve a **single word's** rank.

Measured from CloudTrail S3 data events for 2026-07-07: **97,103 fetches of that one object = 159.7 GB of egress in a single day**, sustained at ~1.1 req/sec around the clock. It was the dominant line item on the AWS bill (S3 went $0.78 in April to $57.02 for Jul 1-26; storage itself is only ~$0.23/mo). Three separate bursts: Jun 8-9, Jun 30-Jul 3, Jul 6-8.

Attribution by burst (CloudTrail S3 data events, all anonymous `Typhoeus` GETs of the same key):

| Date | Source | Signed identity also seen on IP | Bucket | Volume |
| --- | --- | --- | --- | --- |
| Jun 8 | `107.3.91.150` | `lingolinq-dev-app` | `lingolinq-dev-static` | 97.9 GB |
| Jun 30 | `136.124.34.18` | **`lingolinq-cloudrun-prod`** | **`lingolinq-prod-static`** | 121.2 GB |
| Jul 2 | `34.96.47.200` | **`lingolinq-cloudrun-prod`** | **`lingolinq-prod-static`** | 100.0 GB |
| Jul 7 | `67.186.242.78` | local dev machine | `lingolinq-dev-static` | 159.7 GB |

**Two of the four bursts ran in production on Cloud Run**, not just on dev machines. `Async#schedule` enqueues to the **`default`** queue (`boy_band` `AsyncInstanceMethods#schedule` -> `schedule_for('default', ...)`), which is the same queue `BoardDownstreamButtonSet.schedule_once(:update_for, ...)` uses, and every worker type in the Procfile consumes `default`. So ~74k (Jun 30) and ~61k (Jul 2) slow jobs sat ahead of user-facing button-set regeneration for most of a day. This is not migration egress and it will recur on any bulk WordData import.

## Approach

Resque forks per job (v3.0.0 default, `boy_band` does not disable it, no `FORK_PER_JOB=false` in the Procfile), so a class-level in-process memo would be discarded between jobs and fix nothing. The ordered list is cached instead as a Redis `word => rank` hash, making a per-record lookup one `HGET`.

Three correctness details:

- **First-occurrence ranks preserved.** `counts.index(word)` returns the first match; a naive `hmset` loop would let a duplicated word overwrite with its later index and silently change scores. The builder skips words it has already ranked.
- **Lock losers get a correct answer.** A burst of concurrent jobs all miss at once. One wins the build lock; the rest fall back to an uncached lookup rather than reading an empty cache and persisting an unscored priority.
- **Failed downloads are never cached.** An empty or truncated response would poison the cache for the full TTL. A rebuild is only accepted at >= `FREQUENCY_RANK_MIN_WORDS` (1000), and that case returns `nil`, matching pre-cache behavior on a failed download.

## Fix status

| Item | Status | Evidence |
| --- | --- | --- |
| Per-record S3 download removed | Fixed | `app/models/word_data.rb:291` uses `WordData.frequency_rank` |
| Rank cached across forked jobs | Fixed | `assert_frequency_ranks` Redis hash, `word_data.rb:210-238` |
| Concurrent-build stampede | Fixed | `SET NX EX` lock, `word_data.rb:216`; `:busy` falls back |
| Failed download not cached | Fixed | `word_data.rb:220`; spec "should not cache a failed or truncated download" |
| First-occurrence rank semantics | Fixed | `word_data.rb:223-226`; spec "should use the first occurrence of a duplicated word" |
| Duplicate download code in two places | Fixed | both now call `self.frequency_counts` |
| `assert_priority` test coverage | Fixed (was zero) | 10 new specs, `spec/models/word_data_spec.rb` |

Scoring behavior is otherwise unchanged: the batch path still indexes into the list it already passes via `opts['counts']`, thresholds are untouched, and a failed download still contributes no frequency score.

One intentional edge-case change: previously, `opts` present but without a `'counts'` key produced no frequency score at all. It now resolves through the cache. No caller does this (the only caller, `self.assert_priority:172`, always supplies `'counts'`), so this is dead-path cleanup rather than a behavior change in practice.

## Verification

```
bundle exec rspec spec/models/word_data_spec.rb
62 examples, 0 failures, 1 pending
```

The pending example is pre-existing (`update_activities_for`, "Not yet implemented"). The regression guard is "should download the list only once across many lookups": 50 lookups, exactly one `Typhoeus.get`.

## Not covered by this PR

- **No backfill.** Records already written with a too-low priority during the affected runs are not corrected. `WordData.assert_priority(relation)` (the batch path) can re-score them and was already efficient; running it is a separate operational decision.
- **The `after_save` fan-out itself is unchanged.** One job per record is still enqueued; this PR makes each job cheap rather than removing the N-jobs-for-N-records pattern.
- **No feature flag.** Outward behavior is identical and there is no UI surface; this is a background-job performance fix.
- **Unrelated, found while investigating:** `lingolinq-logs-prod` has never received a log. S3 server access logging is configured on both prod buckets but the target bucket has `ObjectOwnership: BucketOwnerEnforced` (ACLs disabled) and no bucket policy granting `logging.s3.amazonaws.com`, so neither delivery mechanism is available. Worth a separate fix if any compliance doc claims S3 access logging is in place.

## Author-Model: opus-5

https://claude.ai/code/session_013FGBa9VbPW3FeF5xwyfF2V

### DIFF
--- a/app/models/word_data.rb
+++ b/app/models/word_data.rb
@@ -165,6 +165,11 @@ class WordData < ApplicationRecord
   FREQUENCY_RANK_TTL = 30.days.to_i
   FREQUENCY_RANK_LOCK_KEY = 'word_data/en/frequency_ranks/building'
   FREQUENCY_RANK_LOCK_TTL = 10.minutes.to_i
+  # The hash is written in slices, so it is built under a staging key and swapped
+  # in with an atomic RENAME. Readers therefore only ever observe the live key as
+  # absent or complete, never half-populated (a partially visible hash would hand
+  # back nil for words not yet written, and that nil gets persisted as a score).
+  FREQUENCY_RANK_BUILD_KEY = 'word_data/en/frequency_ranks/staging'
   # A truncated or failed download must never be cached, so a rebuild is only
   # accepted when it yields at least this many words.
   FREQUENCY_RANK_MIN_WORDS = 1000
@@ -230,12 +235,20 @@ class WordData < ApplicationRecord
         next if word.blank? || ranks.key?(word)
         ranks[word] = idx
       end
+      # Build under the staging key, then swap. A reader that arrives mid-build
+      # sees no live key, fails to take the lock, and falls back to an uncached
+      # lookup rather than reading a hash that is missing most of its words.
+      redis.del(FREQUENCY_RANK_BUILD_KEY)
       ranks.each_slice(5000) do |slice|
-        redis.mapped_hmset(FREQUENCY_RANK_KEY, slice.to_h)
+        redis.mapped_hmset(FREQUENCY_RANK_BUILD_KEY, slice.to_h)
       end
-      redis.expire(FREQUENCY_RANK_KEY, FREQUENCY_RANK_TTL)
+      # Set the TTL before the swap; RENAME carries it over, so the live key is
+      # never left without one.
+      redis.expire(FREQUENCY_RANK_BUILD_KEY, FREQUENCY_RANK_TTL)
+      redis.rename(FREQUENCY_RANK_BUILD_KEY, FREQUENCY_RANK_KEY)
       :ready
     ensure
+      redis.del(FREQUENCY_RANK_BUILD_KEY)
       redis.del(FREQUENCY_RANK_LOCK_KEY)
     end
   end
--- a/app/models/word_data.rb
+++ b/app/models/word_data.rb
@@ -154,6 +154,92 @@ class WordData < ApplicationRecord
     end
   end
 
+  # The English word-frequency list lives in S3 as a ~1.7MB file, ordered
+  # highest-frequency first. Scoring a word only needs that word's rank, but the
+  # per-record path (assert_missing_priority -> schedule(:assert_priority)) used
+  # to re-download the entire file for EVERY word saved. Resque forks per job, so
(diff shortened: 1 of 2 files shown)


swahlquist:
## Codex Review

**Verdict:** REQUEST_CHANGES
**Route:** codex
**Head:** 270bfd7f2ceb4212704633e28993189b4d1bb908
**Run:** https://github.com/lingolinq/LingoLinq-AAC/actions/runs/30223868385


Codex review completed.

| --- | --- | --- | --- |
| partial-rank-cache-visible-during-build | HIGH | app/models/word_data.rb | The Redis frequency-rank hash can be treated as ready while it is only partially populated, causing concurrent jobs to persist incorrect nil frequency ranks for words not yet written. |
| EVIDENCE-1 | HIGH | (diff-wide) | One or more chunk reviews blocked or were inconclusive. |

PR #686, opened 2026-07-26
# Board-detail refinements, action fixes & image uploads

traciday:
# Board-detail refinements, action fixes & image uploads

**29 files · +2,840 / −1,042**

| Area | Files | Churn |
|---|---|---|
| 🎨 Board-detail styling | `app.scss` | +1,288 / −482 |
| 🔗 Actions / links / video | 3 files | +556 / −200 |
| 🧩 Board-detail template | 2 files | +79 / −316 |
| 🖼️ Images & uploads | 4 files | +153 / −12 |
| ↔️ board-alt parity | 2 files | +56 / −15 |
| ✅ Tests / docs / i18n | 17 files | +708 / −17 |

---

## 🎨 Board-detail — speak & edit chrome

`app/frontend/app/styles/app.scss` (+1288/−482) · `templates/user/board-detail.hbs` (+76/−313) · `components/board-detail-grid.hbs` (+3/−3)

- **Speak mode:** flat shell/main surfaces so the thin lighter seam at the top is gone; a frosted-glass compose (sentence) bar; a **circular mic that scales with button size** on short screens; the word-prediction rail seated on the dark board surface; a Home button + an in-folder Back button; the inline sidebar reworked as one glass panel.
- **Folder display modes:** tab-labels, colored-corner, and default folder-back geometry driven from a single source, so folder rows stay aligned across modes.
- **Edit mode:** the sticky session bar now shows the board's **title + icon** (so you always know which board you're editing), relabels `Done Editing → Save` and `Discard → Cancel`, adds an "Editing" status badge, and moves the "Take a tour" trigger into the edit panel. Brighter edit-rail/panel gradients, darker light-mode labels, and short-screen space reclaimed around the preview grid.
- **Grid:** the `hide-label` class now propagates to the card + label span, so "hide the label when the picture is shown" actually hides it.
- Dark mode maintained throughout.

## 🔗 Actions, links, videos & folders

`components/button-settings.js` (+152/−3) · `components/button-settings.hbs` (+202/−111) · `controllers/user/board-detail.js` (+202/−86)

- **Link buttons no longer navigate to the old board.** Switching a button's action clears the conflicting fields across **every** entry point (Action dropdown + Quick-action shortcut), and `select_button` reads the authoritative board button instead of a stale render copy.
[...]

PR #692, opened 2026-07-27
# 🔍 Find-a-button across the whole board tree, faster board opens, landing/privacy copy, and login self-heal

traciday:
# 🔍 Find-a-button across the whole board tree, faster board opens, landing/privacy copy, and login self-heal

Branch: `traci/styling/styling-updates` → `staging`

✅ Staging is fully merged in (`origin/staging` is an ancestor of HEAD), so this is a clean fast-forward-able diff.

> ⚠️ **Scope note:** the branch has 351 commits ahead of staging, but most of that work already landed in staging via squash-merges (notably #686). The **real diff is 33 files, +747 / −275** — that is what this description covers.

---

## 🎯 Why this PR exists

Five independent threads, all verified against the current head:

1. 🔍 **Find-a-button was blind above the current board.** On a sub-board it only searched that board's descendants, so a word on a parent, sibling, or root board could not be found.
2. 🐌 **Board opens felt ~4s slow** even on a cache hit, because the button-set generation ran on the critical path.
3. 👻 **The loading overlay didn't react** under Ember 5.x, so it could stick or never appear.
4. 🔑 **A stale browser token wedged login** until the user manually cleared browser storage.
5. 📣 **Landing/marketing copy overstated our position** on languages, WCAG, AI, and analytics.

---

## 1. 🔍 Find-a-button — search the whole tree, and climb back up

**Problem.** `find-button` built its button set from the *current* board and walked **down**. On a sub-board, everything above you was invisible to search.

**Fix.** Resolve the navigation **root** first, then build that board's whole-tree set.

| | Change | File |
|---|---|---|
| 🌳 | New `_resolveSearchRoot()` — reads `app_state.board_detail_nav_history[0]` (the board the session started on) and builds from there | `components/find-button.js` |
| ⚓ | Anchor return-navigation to the searched tree's own root in speak mode, so `button_steps` can emit a `true_home` step | `models/buttonset.js` |
| ⬅️ | `true_home` / `home` steps drive the **Back** button on board-detail (`[data-bd-action='go_back']`) instead of Home — climbing the tree instead of jumping to the session home board | `controllers/application.js` |
| 🧭 | Guided navigation into a sub-board now pushes nav-history, so Back actually renders | `controllers/application.js` |
| ▶️ | Board-detail's render path never reached `resume_scanning()`, so a guided sequence stalled the moment it entered a sub-board — now resumes, guarded on an actual board change | `utils/edit_manager.js` |

⚠️ **Sharp edge worth reviewing:** button sets are now keyed on the **numeric `global_id`**, never the ember-data record `id`. A board loaded by key has `id == "user/boardname"` and stashes the real id in `_actual_id`. Mixing the two gave the root key-form ids while every descendant was numeric, so the guided highlight would reach the root and then fail to locate the target button. Fixed in `_loadOrBuildButtonSet`, `_buildLocalButtonSet`, `process_board`, and the `from_board_id` passed to `find_sequence` / `find_buttons`.

🛟 `_resolveSearchRoot` falls back to the current board when not in speak mode, when nav history is empty (deep link), or when the root record won't resolve.

🚩 `find_sequence` is behind the existing `find_multiple_buttons` feature flag (`lib/feature_flags.rb`); `find_buttons` is the unflagged path and both were updated identically.

---

## 2. ⚡ Board opens — take button-set generation off the critical path

`setupController` called `model.load_button_set()` inline. For an uncached set that's `POST /api/v1/buttonsets/:id/generate`, which makes the server build the entire find-a-button hierarchy — several seconds on a large board, while the grid itself paints in ~12ms.

⏱️ Now deferred to a `runLater(…, 750)` after paint, matching the existing `warm_images` / `prefetch_linked` pattern. It bails if the component is torn down or the user has already navigated elsewhere. **find-a-button is user-invoked, so nothing waits on it.**

🎬 Also: the board-collection drawer now returns its `Transition` so it can clear the "Opening your board" overlay when the board actually settles, rather than relying on its 8s safety timeout.

---

## 3. 👻 Loading overlay — real reactivity

The component computed `show` / `message` from `app_state.loading_overlay_message`, which was never declared on the service. An undeclared property that is only `.set()` later can fail to notify template consumers under Ember 5.x.

- 📌 Declared `loading_overlay_message: null` on the service
- 🔗 Template binds the service property **directly**; dropped both intermediate computeds and a leftover debug probe `<div>`
- 🧹 Cancelled the orphaned 15s guard timer on hide — it was keeping the run loop alive and tripping Ember's pending-timers assertion in tests
- 🧪 **New test:** `tests/components/app-loading-overlay-test.js` covers the reactive data contract

---

## 4. 🔑 Login self-heal for a stale browser token

The server rotates the browser token (used as `client_secret`) and returns the current one in the `BROWSER_TOKEN` header on **every** response — including the 400 it sends when your token is stale. A tab left open past the token window would send a stale secret, get `"Invalid client_secret for client_id"` *before the password was ever checked*, and stay wedged until browser storage was cleared manually.

- 🔁 `session.authenticate` detects a `client_secret` error, persists the fresh token the server just handed back, and retries **once** (guarded by `browser_token_retried`)
- 💾 `persistence.refresh_browser_token_from_response` keeps the stored token current on every response, so it doesn't go stale mid-session

### 🚪 Entry points (CLAUDE.md P2)

| Entry point | Enforced at | Test |
|---|---|---|
| Password login (`grant_type: password`) | Rails `/token` — unchanged; this is a client-side retry of a **rejected** request, not a new grant | ⚠️ Not covered — see below |
| SPA transition / cached content | No new path; token refresh is passive bookkeeping on responses | n/a |
| Direct API call | Unchanged — the server still validates `client_secret` on every attempt | n/a |

🔒 **This does not weaken authentication.** The retry re-sends the same password with a token the server itself just issued; a wrong password still fails, and the retry is capped at one attempt.

---

## 5. 📣 Landing, privacy and accessibility copy

Marketing claims brought in line with what we can actually support:

| | Was | Now |
|---|---|---|
| 🌐 | "10+ languages" / "10+ language support" | "Multiple languages" |
| ♿ | "WCAG 2.1 AA compliant" | "Working toward WCAG 2.1 AA" |
| 🤖 | AI board generation described as unconditional | "opt-in AI … your content is never used to train AI models" |
| 📊 | Analytics described as unconditional | "Opt in to track … you can turn analytics off at any time" |

✨ Plus a new **Our Mission** section and a **Privacy Commitment** section on landing-alt, and an `aria-label` describing the hero mockup for screen readers.

🧽 The beta-welcome Original/Short layout toggle was removed — the page now always renders the Short version.

🌍 **i18n:** 16 new keys generated and merged across all 13 locale files (`i18n_generator.rb --generate` + `--merge`; TOTAL DUPS 0 / MISSING 0). Non-English locales carry `***` placeholders pending translation.

---

## 6. 🛠️ Operational tooling

- 🔧 **`rake extras:rebuild_button_sets`** — rebuilds every board's downstream button set **sequentially**. `BoardDownstreamButtonSet` coordinates shared sub-boards through a per-board Redis traversal cache and a freshness guard, so concurrent rebuilds of overlapping trees race and can produce empty sets. One board at a time is deliberate. Clears stale `traversed/button_set/*` keys first.
- 🪣 **`.env.op.template`** — documents `UPLOADS_S3_NO_ACL=1`. The dev bucket has Object Ownership = "Bucket owner enforced" (ACLs disabled), so `acl=public-read` uploads fail with `AccessControlListNotSupported`, which silently breaks board copy and button-set generation. Verified honored at `lib/uploader.rb:190` and `lib/uploader.rb:332`.

---

## ✅ Fix status

| | Item | Status | Evidence |
|---|---|---|---|
| 🔍 | Find-a-button searches parent/root boards from a sub-board | ✅ Fixed | `components/find-button.js` `_resolveSearchRoot` |
| ⬅️ | Guided highlight climbs back up via Back on board-detail | ✅ Fixed | `controllers/application.js` (`bd_back` branch) |
| ▶️ | Guided sequence stalled on entering a sub-board | ✅ Fixed | `utils/edit_manager.js` (resume on board change) |
| 🔑 | Button-set ids keyed on numeric `global_id` | ✅ Fixed | `find-button.js`, `models/buttonset.js` |
| ⚡ | Board open blocked on button-set generation | ✅ Fixed | `routes/user/board-detail.js` (deferred 750ms) |
| 🎬 | Board-collection overlay cleared on settle, not by timeout | ✅ Fixed | `controllers/user/board-detail.js` |
| 👻 | Loading overlay reactivity under Ember 5.x | ✅ Fixed | `services/app-state.js`, `components/app-loading-overlay.*` |
| 🧹 | Orphaned 15s overlay guard timer | ✅ Fixed | `services/app-state.js` (`runCancel`) |
| 🔐 | Stale browser token wedging login | ✅ Fixed | `services/session.js`, `services/persistence.js` |
| 📣 | Landing copy overstating languages / WCAG / AI / analytics | ✅ Fixed | `landing-alt-page.hbs`, `features.hbs`, `public/locales/en.json` |
| 🌍 | i18n coverage for new strings | ✅ Fixed | 13/13 locale files, 0 missing |
| 🔧 | Button-set rebuild tooling | ➕ Added | `lib/tasks/extras.rake` |
| 🪣 | ACL-disabled bucket documented | ➕ Added | `.env.op.template`, honored at `lib/uploader.rb:190,332` |

## ⚠️ Not covered by this PR

- 🧪 **No automated test for the browser-token retry.** The self-heal is verified by the reproduction in `docs/task-management/2026-07-25-stale-browser-token-login-wedge.md`, not by a spec. 👉 Requesting a manual check: log in, leave the tab idle past the token window, log out, log back in — should succeed without clearing storage.
- 🧪 **No automated test for the guided find-a-button climb.** The multi-board highlight sequence needs a rendered board tree. 👉 Requesting a manual click-test: from a **sub-board** in speak mode, search a word that lives on the **root** board, then confirm the highlight walks Back up the tree and lands on the correct button.
- 🔧 **`rake extras:rebuild_button_sets` has no spec** and is an operator tool. It is read-mostly per board but does write button sets; run it against dev first.
- 🪣 **`.env.op.template` is shared config.** It only documents a dev-bucket variable and changes no runtime default (`UPLOADS_S3_NO_ACL` is unset in prod ⇒ ACLs still applied), but flagging it since it is outside my usual UI scope.
- 🌍 **Non-English locale strings are `***` placeholders**, not translations.

## 🧪 Verification

| | Check | Result |
|---|---|---|
| ✅ | Ember test suite | **1818 tests / 1778 pass / 40 skip / 0 fail** |
| ✅ | RSpec (`image_spec`, `button_image_spec`, `word_data_spec`, `board_spec`) | **0 failures** |
| ✅ | `sass` compiles `app.scss` | clean |
| ✅ | All 13 locale files parse as JSON | pass |
| ✅ | `i18n_generator.rb --generate` / `--merge` | TOTAL DUPS 0 / MISSING 0 |
| ✅ | Staging merged in | `origin/staging` is an ancestor of HEAD |


traciday:
Commit: fix(ci): allowlist landing_a11y_* i18n keys in gitleaks; make uploader ACL spec env-independent

Before pushing I re-ran gitleaks against the new range (abd6d8c8c..5589a53ab, 275 commits — one more than the failing run, since the fix commit is now inside the scanned range): no leaks found, exit 0. That matters because the config change is itself part of the range being scanned; verifying the old range wouldn't have proven the actual CI invocation passes.

The commit message records both the what and the why-not-the-obvious-alternative:

Why not just rename the key — gitleaks scans a commit range, so the literal survives at 5af1f9bba no matter what HEAD looks like, and a rename would churn all 13 locale files for nothing.
Why the existing locales allowlist didn't already cover it — these keys also live in the .hbs template, outside that path glob.
Evidence the allowlist is narrow — it suppressed the i18n keys while still catching a planted GitHub PAT and a generic secret in the same scan.
The spec fix — it was reading ambient ENV, so anyone following the new .env.op.template guidance would hit a local failure CI never saw; it also never exercised the UPLOADS_S3_NO_ACL branch its own name promised. The new assertion was confirmed to fail against a deliberately broken use_acl before I restored it.
CI should now go green across all six jobs — every one was verified locally first, including the full 6573-example RSpec suite under CI-equivalent conditions.

One thing that hasn't changed: stats_spec.rb:710 will still fail if you run bundle exec rspec locally without TZ=UTC. Pre-existing, untouched by this branch, invisible to CI.

MelissaOneil (approved):

PR #702, opened 2026-07-29
fix(codex-review): resolve deep-pass to a terminal state on every exit path

swahlquist:
Phase 2 of `ai-company-brain outputs/plans/2026-07-28-codex-review-gate-plan.md`.
Scope is **gate reliability only**. No routing, credential, compliance-toggle, or
branch-protection change.

## Problem

`codex-review/deep-pass` wrote a terminal status in only two places: the
input-validation guards at the top of the job, and the final resolve step.
Anything failing between them, notably the reviewer auth step, left the status
`pending` and delegated the problem to `codex-watchdog.yml`, which by design waits
30 minutes.

Observed 2026-07-28: auth failed in about 90 seconds while `deep-pass` read
"Codex review running" for 41-65 minutes on #690, #696 and #697.

**The watchdog was working correctly.** It found all three and flipped them to
`failure`. The primary workflow was misreporting its own state, so a fast failure
looked like a slow review. This PR fixes the reporting and keeps the watchdog as
the independent backstop.

## Changes

**1. Attempt-scoped run identity.** `RUN_URL` is now job-level env and includes
`/attempts/<run_attempt>`. `runs/<run_id>` alone is identical across re-runs, so
attempt 2 could match attempt 1's terminal status in the resolve step's verify loop
(reporting false success) and inherit attempt 1's `created_at` (defeating the
watchdog clock). All six `target_url` writes and `run-chunks --run-url` use it.

**2. Terminal-status lifecycle step, `if: always()`.** Covers success, failure,
cancellation, re-run on the same SHA, and the status write itself failing.
`if: failure()` is insufficient because it does not run on cancellation.
Idempotent, retries the write 3x, fails the job loudly if it cannot write, and
fails closed if a green job produced no verdict.

**3. Watchdog anchors on the earliest pending status for the attempt in force**,
read from the plural `/statuses` endpoint. `run-chunks.py` re-POSTs a pending
heartbeat per chunk, each minting a new `created_at`, so anchoring on the latest
entry let a heartbeating run push the clock forward indefinitely and outrun the
30-minute sweep, in exactly the hang case the backstop exists for.

## Pre-push verification

- 44 existing codex-review unit tests pass (envelope 27, evidence 11, chunks 6)
- both workflows parse; `bash -n` clean over every embedded `run` block
- jq anchor logic checked against synthetic multi-attempt heartbeat data

## Live verification, 2026-07-29 (attended)

All five cases exercised. Cases 2, 3 and 4 pass outright; case 1 is deferred with
[...]

PR #705, opened 2026-07-29
docs(process): temporary --admin exception policy for codex-review/deep-pass

swahlquist:
Resolves Fork 2 of the codex-review gate plan. Companion to #702 (the gate-reliability fix); deliberately separate so that PR stays narrow.

## Why

`codex-review/deep-pass` is a required status check on `staging` with **no funded reviewer route** since 2026-07-27. Demoting it to advisory was rejected: `enforce_admins` is false so the override already exists, and demotion would silently remove the gate for every future PR rather than for the specific PRs affected by this outage.

**The check stays required. Branch protection is unchanged.** This policy governs the override instead, so each bypass is a deliberate, recorded act.

## Fields

- **Approver:** Scot Wahlquist (`swahlquist`), sole approver
- **Expiry:** 2026-08-28, or earlier when an authorized reviewer route passes the five-case verification in #702
- **Audit log:** `audit-reports/deep-pass-admin-overrides.md` (added here, empty)

## What it says

It states plainly that `gh pr merge --admin` **bypasses branch protections broadly**, not just `deep-pass`: in one command it can also override `rspec`, `build-and-test`, `audit-artifacts-integrity`, `secret-detection` and `codex-review-tests`. **An `--admin` merge is not equivalent to a passing reviewer gate.**

Five conditions must all hold:

1. `deep-pass` is failing because no route is funded, evidenced by the run failing at the auth step, not the review step
2. every other required check is green **on its own**
3. the PR is not on a data-bearing path per `scripts/codex-review-path-classifier.sh`
4. a **named human** actually reviewed the diff
5. a listed approver authorized that specific PR

**The approver does not silently count as both reviewer and approver.** A separate human diff reviewer is required whenever one is available; where none is, the entry must record `SAME AS APPROVER - no second reviewer available` explicitly, so the collapse is visible and countable at the lapse review.

Required-check results are enumerated individually in each entry rather than summarized as "all green," since `--admin` bypasses them all at once.

## Temporary by construction

It lapses on the earlier of the expiry date or a verified route. If the outage is unresolved at expiry it must be **explicitly renewed** with a new date, never rolled over. A review on lapse asks two questions, the second being whether the human-review condition held every time or decayed into a formality under time pressure.

## Not in this PR

Provider routing, credentials, `CODEX_COMPLIANCE_PATHS`, and the sync-config repair all remain out of scope.

https://claude.ai/code/session_019eWWTEBEwJBqbkwmyku22n

### DIFF
--- a/audit-reports/deep-pass-admin-overrides.md
+++ b/audit-reports/deep-pass-admin-overrides.md
@@ -1,0 +1,7 @@
+# deep-pass `--admin` override log
+
+One entry per override, per `docs/process/deep-pass-admin-exception-policy.md`.
+Every field in that policy's evidence block is mandatory. Do not abbreviate
+required-check results to "all green": `--admin` bypasses all of them at once.
+
+_No overrides recorded yet._
--- a/docs/process/deep-pass-admin-exception-policy.md
+++ b/docs/process/deep-pass-admin-exception-policy.md
@@ -1,0 +1,172 @@
+# `--admin` merge exception policy: codex-review/deep-pass
+
+**Status: APPROVED 2026-07-28.** All three open fields set by Scot. Effective on
+merge into LingoLinq-AAC.
+
+Resolves Fork 2 of the codex-review gate plan (ai-company-brain
+`outputs/plans/2026-07-28-codex-review-gate-plan.md`).
+Audit log: `audit-reports/deep-pass-admin-overrides.md`.
+
+---
+
+## Why this exists
+
+`codex-review/deep-pass` is a required status check on `staging`. As of 2026-07-27
+it has **no funded reviewer route**: `CODEX_OPENAI_API_KEY` was deleted deliberately
+and `CLAUDE_REVIEW_API_KEY` never existed. Every run since has failed at
+authentication.
+
+Two options were considered. Demoting the check to advisory was **rejected**:
+`enforce_admins` is false, so the override already exists, and demotion would
+silently remove the gate for every future PR rather than for the specific PRs
+affected by this outage. **The check stays required.** This policy governs the
+override instead, so each bypass is a deliberate, recorded act.
+
+## Scope
+
+Applies only to merges blocked by `codex-review/deep-pass` while no authorized
+reviewer route is funded and verified. It does not apply to any other failing
+required check.
+
+## What `--admin` actually does
+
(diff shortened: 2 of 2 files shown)


swahlquist:
## --admin merge under a ONE-TIME EMERGENCY EXCEPTION

This PR **adds** the exception policy, so the policy cannot govern its own merge; it is not yet in effect. Recorded here to the same standard the policy will require from the next override onward.

```
--admin merge, one-time emergency exception (policy not yet active)

PR:                   #705
Head SHA:             668c367610bca313690f2829e4f3e02755fd3673
Timestamp (UTC):      2026-07-29T07:55Z
Reason for exception: Lands the deep-pass --admin exception policy itself.
                      deep-pass has no funded reviewer route, so this PR cannot
                      pass the gate it exists to govern. Bootstrapping only.

Auth-failure evidence (condition 1):
  deep-pass run URL:  https://github.com/lingolinq/LingoLinq-AAC/actions/runs/30425705022
  Failing step:       Authenticate Codex CLI    <- auth, NOT the review step
  Run conclusion:     failure
  Status timeline:    05:39:18Z pending -> 06:20:56Z failure
                      "Timed out after 41min with no result (watchdog)"
                      (default-branch code; the 41 min is the defect #702 fixes)

Human diff reviewer:  Scot Wahlquist (SAME AS APPROVER - no second reviewer available)
Approver:             Scot Wahlquist (swahlquist)

Required-check results (condition 2, each green on its own):
  rspec:                      pass
  build-and-test:             pass
  audit-artifacts-integrity:  pass
  secret-detection:           pass
  codex-review-tests:         pass
  security-scan:              pass
  comment-preview-url:        skipping (not required)
  codex-review/deep-pass:     failure  <- the check this exception covers

Data-bearing path (condition 3): no
  classifier: data_bearing=false compliance_path=true
              (compliance_path=true because this PR adds audit-reports/
               deep-pass-admin-overrides.md; inert while CODEX_COMPLIANCE_PATHS=allow)
```

Note: `--admin` bypasses branch protections **broadly**, not just `deep-pass`. Every other required check was verified green on its own before use, as the policy requires.

Follow-up: this override must be back-filled into `audit-reports/deep-pass-admin-overrides.md` once merged.

PR #718, opened 2026-07-31
fix(ai): stop telling operators to set the retired ANTHROPIC_API_KEY

swahlquist:
## Problem

Runtime AI moved to AWS Bedrock (`lib/ai_client.rb`), but four operator-facing diagnostics still tell you to set `ANTHROPIC_API_KEY`.

`resolve_api_config` returns nil purely on `AiClient.configured?`, which checks `bedrock_region.present? && aws_credentials.present?` and never looks at `ANTHROPIC_API_KEY`. So the single message an operator sees when board generation or prediction seeding is dark points at a variable the code no longer reads. Setting it changes nothing, the feature stays off, and nothing says why.

Found while tracing which credential the generate-board feature actually needs. The stale message is what made that question hard to answer.

## Sites corrected

| File | Line | Was |
|---|---|---|
| `lib/ai_board_generator.rb` | 17 | method comment "Requires ANTHROPIC_API_KEY" |
| `lib/ai_board_generator.rb` | 30 | `dev_diag` in `generate_words` |
| `lib/ai_board_generator.rb` | 240 | `dev_diag` in the second generator path |
| `lib/ai_prediction_generator.rb` | 29 | `puts` in `generate` |

The remaining `ANTHROPIC_API_KEY` mentions in `lib/` are all negations ("no longer configures this path", "intentionally not built"), including the existing one in `ai_client.rb:14`.

## What the new messages say

They name the variables actually consulted (`BEDROCK_AWS_KEY`, `BEDROCK_AWS_SECRET`, and a region from `BEDROCK_AWS_REGION` / `AWS_REGION` / `AWS_DEFAULT_REGION`) and surface the two non-obvious failure modes already documented in `ai_client.rb` but invisible to an operator reading only the error:

- **Both halves of the pair are required.** `credential_pair` returns nil unless both are present and non-blank, selected atomically so a partial rollout cannot mix a dedicated Bedrock key with a generic S3/SES secret.
- **`AWS_KEY` / `AWS_SECRET` are not a fallback.** They are the Cloud Run S3/SES least-privilege credentials (`scripts/gcp/iam/lingolinq-cloudrun-s3-ses-policy.json`) and lack Bedrock invoke permission. Per the comment at `ai_client.rb:42-46`, falling back to them previously made `configured?` true while every request returned AccessDenied, which is the worst version of this failure: it looks configured and fails at call time.

`ANTHROPIC_MODEL` is unchanged and still honoured as a model-id override, normalized to Bedrock form.

## Scope

Message text only. No control flow, no configuration resolution, no model routing, no change to which credentials are read.

## Verification

- `ruby -c` clean on both files.
- `spec/lib/ai_board_generator_spec.rb` and `spec/lib/ai_word_predictor_spec.rb`: **29 examples, 0 failures**.
- No spec asserts the old diagnostic text (checked before rewording).

## Related

Bedrock cutover context: AWS BAA signed 2026-02-07, Anthropic HIPAA-Ready BAA executed 2026-07-18 as the coverage bridge. This PR does not change compliance posture; it makes the failure message point at the right place.

https://claude.ai/code/session_01G975WXqu1Kz4UYU1qEaLUh


MelissaOneil (approved):

swahlquist:
Superseded by #727, which now carries this PR's work.

Verified before closing rather than assumed: this PR did contain unique changes not in #727 at the time, so they were ported rather than discarded.

**Ported in 5dbc4e478** (`security(ai): gate the ANTHROPIC_MODEL override and pin the Bedrock endpoint`):
- The richer `dev_diag` / `puts` operator diagnostics in `lib/ai_board_generator.rb` (both call sites) and `lib/ai_prediction_generator.rb`, naming **both** accepted credential pairs, stating that both halves are required and a partial pair is ignored, listing all three region variables, and explaining that `AWS_KEY`/`AWS_SECRET` are deliberately rejected as the S3/SES least-privilege pair. #727's original text named only the `BEDROCK_*` pair and was strictly less complete.
- The expanded `generate_words` header comment.

**Ported in 1bc058289** (`compliance(ai): name AWS Bedrock as the processor...`):
- LEARNINGS.md rule 5 (diagnostics must mention both credential pairs).
- The `CAPABILITY-LEDGER.json` `currentEvidence` line-anchor corrections, `lib/ai_board_generator.rb` 43 -> 58 and 134 -> 149. Independent confirmation the port was faithful: applying this PR's comment additions to #727 shifted those anchors to exactly the same two line numbers.

Not ported, deliberately: nothing. Every change here is represented.

#727 additionally fixes the underlying reason the diagnostics mattered. Runtime AI was dead because `AiClient` only built the Mantle client and this AWS account has no Mantle entitlement (403 on every model even with admin credentials). #727 adds the classic `bedrock-runtime` plane, which is entitled and verified working end to end.

PR #731, opened 2026-08-02
fix(rebuild-button-sets): sweep copied trees the root pass cannot reach

swahlquist:
Follow-up to #724 / #726, raised by the Codex review pass.

`extras:rebuild_button_sets` printed "sub-boards reference their root's set via source_id, so the whole tree is covered" unconditionally. That claim is false for copied trees, leaf boards, and standalone boards.

## Why the root pass misses them

Root selection requires `upstream.empty? && downstream.any?`. A copied tree never satisfies it: every board in the copy carries an upstream link back to the original it was copied from, so nothing in it qualifies as a root. Those boards also do not inherit a `source_id`, because the original root's `linked_board_ids` reference its own children, not the copies.

## Measured in prod, not theorized

After the full root rebuild on 2026-08-02 (451 roots, 0 errors), **144 boards that had buttons still reported `button_count=0`** -- copies owned by `bob`, `scotw`, `melissaoneil`, `prodtest…`, plus the public `lingolinq/crisis-vocabulary` (48 buttons).

Rebuilding exactly those 144 via `update_for` repaired **144/144, 0 errors**, taking prod from 180 empty sets to 36 -- and all 36 remaining are boards with genuinely zero buttons (`of_which_board_has_buttons=0`).

## Change

Pass 2 folds that sweep into the task: any set with `button_count=0` whose board has buttons is rebuilt directly. Boards that remain empty are listed and folded into the nonzero exit, so the task can no longer report coverage it did not achieve. The false claim is deleted.

## Fix status

| Item | Status | Evidence |
|---|---|---|
| False coverage claim | Fixed | claim deleted; pass-2 comment states the real constraint |
| Copied/leaf/standalone trees unreachable | Fixed | pass 2 sweeps by `button_count=0 AND board.buttons.any?` |
| Silent partial success | Fixed | stubborn boards folded into `empty`, task exits nonzero |

## Not covered by this PR

- No spec. This task has no existing spec coverage and exercising it needs a
  multi-board copied-tree fixture; the behavior was verified against prod instead.
- The second latent loss path at `extra_data.rb:55` (throttled upload still deletes
  the attribute) remains open from #724.

## Author-Model: opus-5

### DIFF
--- a/lib/tasks/extras.rake
+++ b/lib/tasks/extras.rake
@@ -334,6 +334,11 @@ end
 # freshness guard, so concurrent/queued rebuilds of overlapping board trees race
 # and can produce empty sets. One board at a time is the correct, prod-proven mode
 # (this mirrors the button-set half of extras:fix_prod_setup).
+#
+# Runs in two passes. Pass 1 rebuilds true roots (no upstream, has downstream). Pass 2
+# sweeps every remaining board whose set is empty but whose board HAS buttons -- copied
+# trees, leaves, and standalone boards are all structurally unreachable from pass 1. See
+# the pass-2 comment for why. Exits nonzero if any board with buttons ends up empty.
 task "extras:rebuild_button_sets" => :environment do
   # Preflight. Without REMOTE_EXTRA_DATA the detach-to-S3 path is a no-op, and this
   # task used to run to "completion" while writing button_count=0 for every set over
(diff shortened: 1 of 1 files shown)


swahlquist:
**Admin override rationale (approved by Scot, 2026-08-02)**

> Admin override approved for PR #731 because ordinary CI is green, the diff is limited to `lib/tasks/extras.rake`, Codex/Claude review found no code blocker, and the only unmet branch-protection item is the known unsatisfiable `codex-review/deep-pass`, tracked in #733. This does not approve bypassing substantive failed checks.

State at merge: 6/6 checks green (rspec, build-and-test, audit-artifacts-integrity, secret-detection, security-scan, codex-review-tests), 1 file changed, +55/-2, `mergeable=MERGEABLE`, `mergeStateStatus=BLOCKED` solely on the absent deep-pass status.

Prod was already repaired before this merge; this PR prevents the next operator from hitting the copied-tree coverage gap, it does not fix a live break.

PR #696, opened 2026-07-28
release: staging into main 2026-07-28

swahlquist:
## Summary

Promotes current `staging` to `main` for the production deploy path.

This release includes the Article 50 transparency prerequisites while keeping the `article_50_disclosure` flag OFF.

## Required Article 50 PRs included

| PR | Status on staging | Evidence |
| --- | --- | --- |
| #635 Phase 4 call-context/jurisdiction stamping | Included | merge commit `4a50c522f802ee7662ddbc8fb31fc62ffb3328fb` |
| #646 Phase 3 disclosure content, ack endpoint, modal, first-AI-use gate | Included | merge/squash commit `8800d1c6757c51f4a0aa7ce83da5dacda0929d73` |
| #656 Phase 5 AVAILABLE-only flag registration, retention reconciliation, a11y | Included | merge/squash commit `74266b41f0c3b5eddf265e6d85db9ebbe6d7d228` |
| #694 `LL-a9d6d5a46b` link contrast blocker | Included | merge/squash commit `99c3da99568a7d0c7fb38a404e795f5a76070cce` |

## Flag state

`article_50_disclosure` remains AVAILABLE-only in `lib/feature_flags.rb`; it is not in `ENABLED_FRONTEND_FEATURES`. This release must deploy with the flag OFF. Enabling the flag remains a separate Scot sign-off step after production verification.

## Verification so far

- Compared `main...staging`: staging has the release contents to promote.
- Confirmed `lib/feature_flags.rb` on staging registers `article_50_disclosure` in `AVAILABLE_FRONTEND_FEATURES` and leaves it out of the visible `ENABLED_FRONTEND_FEATURES` block.
- #694 CI passed and Claude Code deep review approved the contrast fix.

## Not covered by this PR

- Does not enable `article_50_disclosure`.
- Does not close the Notion finding `LL-a9d6d5a46b`; closure should wait for production deploy evidence and verification.
- Does not perform the post-deploy AT/manual Article 50 modal walkthrough.

## Author-Model

Codex GPT-5

PR #681, opened 2026-07-24
feat(ai): route runtime AI to Claude on AWS Bedrock (off direct api.anthropic.com)

swahlquist:
Switches all four Tier 1 runtime AI seams from the direct `api.anthropic.com` endpoint to **Claude on AWS Bedrock** (Bedrock Mantle Messages API) via a new shared `lib/ai_client.rb`, on dedicated AWS creds. No direct-Anthropic fallback.

## Why this is compliant (not a regression)
- Amazon Bedrock is a **HIPAA-eligible AWS service, excluding Fable/Mythos models** (verified against AWS's HIPAA-eligible-services reference, 2026-07-24). Runtime inventory (Haiku 4.5, Opus 4.7) is on the eligible side.
- Inference stays **inside AWS's HIPAA boundary**, covered by the existing AWS account BAA (`docs/legal/AWS_BAA_ACCEPTED.md`, account 2390-4478-5114). The `ANTHROPIC_BAA_ACCEPTED.md` line-124 "AWS BAA doesn't extend to Anthropic egress" note was about the *direct third-party endpoint*, not an in-AWS service — it does not apply to Bedrock.
- The executed Anthropic HIPAA-Ready BAA (2026-07-18) remains on file as a still-available direct path.

## What changed
- `Gemfile`: `anthropic ~> 1.23` → `~> 1.36` (`Anthropic::BedrockMantleClient`; locks 1.58.0). Mantle signs via `aws-sdk-core` (already present) — no new AWS gem.
- `lib/ai_client.rb`: single sanctioned construction point; returns nil when AWS unconfigured (callers keep the existing "AI not configured" degrade path).
- Model ids → Bedrock form (`anthropic.claude-haiku-4-5`, `anthropic.claude-opus-4-7`); `eval_narrator` exact-ID allowlist repointed to Bedrock ids (fail-closed control preserved).
- `scripts/ai-endpoint-guard.sh` + CI step: fails the build if any runtime seam constructs a direct `Anthropic::Client` or reads `ENV['ANTHROPIC_API_KEY']`.
- Legal reconciliation: dated supersession notes in `ANTHROPIC_BAA_ACCEPTED.md` / `AWS_BAA_ACCEPTED.md` / `AI_GOVERNANCE_MEMO.md` + regenerated `DOCUMENT-REGISTER`.

## ⚠️ Required before merge (only Scot can do these)
[...]

PR #725, opened 2026-08-01
compliance(baa): retract the unverifiable Bedrock credential attestation and reconcile the corpus

swahlquist:
## What

`docs/legal/AWS_BAA_ACCEPTED.md` carried this under Scot's 2026-07-27 CEO re-attestation:

> **Verified 2026-07-27:** the deployed Cloud Run service `lingolinq-web` signs Bedrock with a credential that resolves to this account, region us-west-2.

It could not have been verified. This PR retracts it, records the evidence, and reconciles every other place in the corpus that repeated the same claim.

**Reviewed twice.** A dual review (Codex `gpt-5.5` + Claude adversary) of the first head found that the correction itself committed the error class it retracts. That is fixed here. An adversary apply-check then verified every fix landed and surfaced two residuals, also fixed.

## Why the original claim is wrong

`AiClient.aws_credentials` signs Bedrock only with `BEDROCK_AWS_KEY` + `BEDROCK_AWS_SECRET`, or `AWS_ACCESS_KEY_ID` + `AWS_SECRET_ACCESS_KEY`. It deliberately does not fall back to `AWS_KEY` / `AWS_SECRET`, the S3/SES pair.

| Check | Result |
| --- | --- |
| Bedrock credential env vars, all 11 revisions `00001-2vn` to `00011-l7f` | none, on every revision |
| Serving revision on 2026-07-27, `00010-95c` | none |
| `BEDROCK_AWS_*` in `deploy-cloudrun.yml` on `staging` | absent |

Never true at any point. **Root cause**, recovered from the capability ledger during review: the `anthropic-model-provider-baa` row recorded the verification with the parenthetical *"residual: shared app AWS credential, not a dedicated Bedrock credential."* Someone believed `AWS_KEY`/`AWS_SECRET` were signing Bedrock. They cannot.

## The Critical this PR also fixes (self-inflicted)

The first version of this correction claimed no data reached any model endpoint "since the 2026-07-24 routing change." **Also false, and by the same reasoning error.** The Bedrock commit `abd6d8c8c` (#681) merged 2026-07-27 11:45 MDT, but `00010-95c` was created 2026-07-24T23:21Z, three days earlier, and ran pre-#681 code constructing `::Anthropic::Client.new(api_key: ENV['ANTHROPIC_API_KEY'])` with the key mounted.

Now stated as four scoped claims:

1. Bedrock egress **never** occurred, on any revision (verified).
[...]

PR #722, opened 2026-08-01
compliance(runbook): attest BREACH_RUNBOOK v2.2.1 + correct the attestation record

swahlquist:
## What

Attests **BREACH_RUNBOOK v2.2.1** and corrects the attestation record that shipped with #703.

#703 merged the corrected runbook bytes (`0ee1b92e`) but left the register recording `attestedDate` 2026-07-28, the day the **superseded** bytes (`fbdf49a1`) were attested. The pin therefore asserted that the attester had reviewed content authored a day later. That is the specific failure mode `attestedContentHash` exists to catch: `scripts/document-register-render.rb:175` defines it as recording "the bytes Scot actually signed off on", so the hash and the date have to move together.

**No substantive runbook prose changed.** The three content corrections themselves (New York Ed Law 2-d seven-day cap, the §4.2 kill list, the §11 vendor contacts) already merged in #703 and are live on staging. This PR fixes the record over them.

## What changed

- **Version label `v2.2` to `v2.2.1`**, so each label identifies exactly one byte-set. `v2.2` now unambiguously means `fbdf49a1`, which remains in merge history.
- **`attestedDate` 2026-07-28 to 2026-07-31**, with 2026-07-28 appended to `priorAttestations` rather than silently overwritten. `lastReviewed` / `nextReviewDue` to 2026-07-31 / 2027-07-31.
- **Changelog split in two.** The `v2.2` entry now describes only what was attested on 2026-07-28 and closes with "Attested ... against hash `fbdf49a1...`". A new `v2.2.1` entry records the three post-attestation corrections with their citations (8 NYCRR § 121.10(b) and (e), EU AI Act Art. 50(1), `AI_FEATURES` in `lib/feature_flags.rb`) plus the bookkeeping problem itself.
- **Register note rewritten** to describe the correction rather than the defect. The prior note claimed the revision "corrects the ... NY-Ed-Law notification prose", which described the version that broke it.
- **Changelog no longer asserts Object Lock "supersedes" MFA-delete**; it states what changed. The §4.1 body claim is retained because it carries its own justification (compliance mode blocks deletion before retention expiry under any credentials).
[...]

PR #734, opened 2026-08-03
fix(ai): prod hotfix - route runtime AI over the classic Bedrock plane

swahlquist:
Production hotfix onto `main`, mirroring the pattern used for #726. Cherry-picks the Bedrock runtime stack already merged to `staging` (#727 `a1083e4bd`, #719 `57bd9f729`) **without** dragging the other 49 staging commits into production.

## Why a hotfix instead of deploying staging

Production Cloud Run runs `main` (currently `796390240`, revision `lingolinq-web-00012-x8z`). Deploying `staging` would have shipped **51 commits**, bypassing the `release/staging-into-main` process. This keeps the production delta to the Bedrock runtime path only.

## Problem

Runtime AI has been non-functional since #681, silently: callers degrade to "AI is not configured" rather than erroring.

`AiClient` built only `Anthropic::BedrockMantleClient`, and AWS account `239044785114` has **no entitlement on the bedrock-mantle plane**. Every model returns 403 `not available for this account`, and it does so even with **admin** credentials and `bedrock-mantle:CreateInference` on `Resource: "*"`, which rules out IAM. An access request is open with AWS.

The **classic** `bedrock-runtime` plane is entitled and working. The planes carry **different model catalogs**, so this adds classic alongside Mantle rather than replacing it.

### Verified live, end to end

Through the app's own `AiClient.build` with the production Bedrock credentials:

```
plane classic / us-west-2 -> Anthropic::Helpers::Bedrock::Client
anthropic.claude-haiku-4-5 -> us.anthropic.claude-haiku-4-5-20251001-v1:0
SUCCESS  reply "hot"  usage in=21 out=4
```

| Model | us-west-2 | us-east-1 | us-east-2 | Reading |
|---|---|---|---|---|
| Haiku 4.5 (`us.` profile) | OK | OK | OK | entitled |
| Opus 4.5 (`us.` profile) | AccessDenied | AccessDenied | AccessDenied | in catalog, unsubscribed |
| Opus 4.7 | ValidationException | ValidationException | ValidationException | **not in the classic catalog** |

## What is in this PR

- **Dual-plane `AiClient`** selected by `BEDROCK_PLANE` (classic default, mantle opt-in). Unrecognized values degrade to classic, not to an unentitled plane. The clients take different secret keywords (`aws_secret_key` vs `aws_secret_access_key`); a spec reflects on both real signatures so a gem rename fails CI, not production.
- **`bedrock_model()` fixed.** It tested `start_with?('anthropic.')` and so mangled `us.anthropic.claude-haiku-4-5-20251001-v1:0` into `anthropic.us.anthropic....`. Classic **requires** the profile form; the bare id returns `ValidationException: on-demand throughput isn't supported`. Legacy dated overrides (`claude-haiku-4-5-20251001`, the value the old `.env.example` documented) also resolve.
- **`ANTHROPIC_MODEL` allowlist** (`ALLOWED_RUNTIME_MODELS`, fail-closed) on the three seams that carry student data and previously had no model gate. `canonical_alias` strips the region prefix and version suffix so a Covered Model disguised as `us.anthropic.claude-fable-5-...` is caught, not just the bare alias.
- **Bedrock endpoint pinned.** Both clients resolve `base_url ||= ENV.fetch("ANTHROPIC_BEDROCK_BASE_URL", ...)`, so leaving it nil let an env var silently redirect all Tier 1 egress, still SigV4-signed, off the BAA'd path. `build` now passes the region-derived URL explicitly.
- **Article 50 / consent disclosures corrected.** The public page said data goes to "Anthropic's commercial API" under "a zero-data-retention agreement". It goes to AWS Bedrock. AWS is now named as the processor and Anthropic as model provider only (on Bedrock the provider cannot access prompts or completions). No ZDR claim is made, because AWS's `default` retention mode permits retention for abuse prevention and the account mode has not been set. Opus 4.7 removed; eval narration is stated as not active.
- **`deploy-cloudrun.yml`**: mounts `BEDROCK_AWS_KEY` / `BEDROCK_AWS_SECRET` (from #719).
- **Capability ledger**: evidence anchors moved to lines 58 / 149 (required, `capability-check` validates them at HEAD) and the plane facts corrected.

## Conflicts

**None.** Main and pre-merge staging were byte-identical on all 19 runtime files, verified before applying. The Bedrock architecture already existed on main via the #696 release; main simply carried the Mantle-only variant.

## Deliberately excluded

`docs/task-management/LEARNINGS.md`, which has independent staging drift and is documentation rather than runtime. It ships at the next release.

## Deploy prerequisites (all confirmed)

- Secret Manager: `BEDROCK_AWS_KEY` and `BEDROCK_AWS_SECRET` both exist with an **enabled** version, values confirmed by SHA-256 to be the credentials verified against Bedrock (no plaintext printed).
- `roles/secretmanager.secretAccessor` granted on both to `lingolinq-run@lingolinq-prod.iam.gserviceaccount.com`, the Cloud Run runtime SA.
- `AWS_REGION=us-west-2` already set, so no `BEDROCK_AWS_REGION` is needed.

## Fix status

| Item | Status | Evidence |
| --- | --- | --- |
| Word prediction reaches Claude | Fixed | live call above |
| Board generation / prediction seeding | Partial: same client + model path, not individually executed | identical `AiClient.runtime_model(...)` resolution |
| Eval narration | **Not fixed** | Opus 4.7 absent from the classic catalog; stays on its template fallback |
| `bedrock_model` mangling profile ids | Fixed | regression spec |
| `ANTHROPIC_MODEL` unvetted-model path | Fixed | allowlist + `canonical_alias`; specs cover disguised Fable/Mythos |
| `ANTHROPIC_BEDROCK_BASE_URL` redirect | Fixed | explicit `base_url`; spec sets a hostile value |
| Art 50 page named wrong processor | Fixed | AWS/Anthropic split; ZDR claim withdrawn |

## Not covered by this PR

- **Eval narration.** Needs routing to Haiku 4.5 (changes disclosed model names) or Mantle entitlement.
- **Attested corpus untouched**, pending live verification: `SUBPROCESSORS.md` does not list Bedrock; `AWS_BAA_ACCEPTED.md` pins "region us-west-2" while AWS documents that with cross-region inference "retained inputs and outputs are stored in destination regions". Both need re-attestation.
- **Bedrock account data-retention mode is unresolved** (`inherit`, so model default). The runtime credential is correctly denied `bedrock:GetAccountDataRetention`, so this cannot be read without admin credentials. Setting it to `none` is what would permit a true zero-retention claim.
- **`article_50_disclosure` flag is off**, so the corrected page reaches nobody via the modal.
- **`aws-sdk-core` 3.245.0 to 3.254.0** (required by `aws-sdk-bedrockruntime`), which also signs S3/SES.

## Post-merge deploy verification

Deploy from `main`, then confirm: deploy run SHA is this merge commit; Cloud Run revision advanced from `lingolinq-web-00012-x8z`; env contains both Bedrock variables; deployed code carries `ALLOWED_RUNTIME_MODELS` and the pinned base URL; `AiClient.configured?` true (**necessary, not sufficient**); AWS caller identity is `239044785114`; board-generation smoke succeeds; `AiApiLog` records `us.anthropic.claude-haiku-4-5-20251001-v1:0`; eval narration still falls back.


swahlquist:
Codex blocker before approval: one production-facing disclosure inconsistency remains.

lib/lingo_linq/ai_consent_disclosures.rb now removes eval narration from the vendor/features/model list because Opus 4.7 is not active on classic Bedrock, but the same metadata still lists clinical evaluation notes as a data category and says revocation stops data being sent for evaluation narration. That tells parents/consent consumers eval data is still part of the AI-sharing flow even though this PR says eval narration is inactive and egresses nothing.

Please align the consent metadata with the runtime fact before merge: either remove eval narration from the active consent data categories/revocation text, or explicitly mark it as inactive/future in the same style as Article 50. Add or adjust a spec so this cannot regress, for example no active consent data category or revocation text implies eval-narration egress while Opus 4.7/eval narration is not in the active vendor model/features set.

Everything else I checked looks aligned: CI is green, guard passed, Bedrock secret mounts are present, model allowlist canonicalizes regional/profile forms, and Bedrock base_url is pinned instead of inherited from env.

swahlquist:
Codex follow-up: I cannot approve b0af83157 yet.

High: app/frontend/app/templates/privacy.hbs:67 still says runtime AI runs on Anthropic commercial API, names Opus 4.7 for eval summaries, and claims zero-data-retention. After this PR deploys, runtime board and word AI goes to AWS Bedrock Haiku 4.5, eval narration remains inactive/local fallback, and account-level Bedrock zero retention is explicitly unresolved.

High: app/views/ai_consent/disclosures/v1.html.erb:50 plus config/locales/en.yml:135 still renders ZDR copy whenever trains_on_data is false. The metadata now says no ZDR guarantee, but the rendered consent page would still tell users there is a zero-data-retention agreement.

Medium/High adjacent: config/locales/en.yml:126 still presents clinical eval notes as sent when SLP requests an AI summary. Consent metadata was fixed to say eval narration is inactive and sends nothing; the rendered page should match.

Recommended fix: update these rendered surfaces in #734, add specs that render or search the actual pages for no Anthropic commercial API, no active Opus 4.7 claim, no zero-data-retention guarantee, and eval inactive wording. Then I will re-check and approve if CI is green.

swahlquist:
## Rendered-surface blockers fixed (9abe7bd22)

All three confirmed and fixed, plus two more of the same class found while tracing them.

| Surface | Was | Now |
|---|---|---|
| `privacy.hbs:67` | "Anthropic's commercial API", Opus 4.7 for eval summaries, ZDR claimed | Amazon Bedrock under the AWS BAA, Haiku 4.5 only, eval summaries marked inactive, ZDR guarantee explicitly declined |
| `v1.html.erb:50` + `en.yml:135` | ZDR sentence selected from `trains_on_data == false` | renders each vendor's own `training_note` from the registry |
| `en.yml:126` | clinical eval notes presented as sent | "nothing is sent today" |
| `public/locales/*.json` (12 of 13) | Google Gemini fallback, disabled 2026-07-09 (#570) | corrected |
| `en.yml` `training_confirmed_zdr` / `training_unconfirmed` | both asserted claims the Bedrock path cannot support | retired to neutral text |

### On `v1.html.erb:50`

Worth separating from the string fixes: the defect was the mechanism. The view chose legal prose from a boolean, and on Bedrock **both** entities are `trains_on_data == false` for different reasons (AWS does not train on it; Anthropic never receives it). So the page asserted a zero-data-retention agreement *twice*, for a vendor that does not have one and for a vendor that is not in the path. Patching the string would have left the trap armed for the next vendor added.

### Guards

`spec/support/ai_disclosure_claims.rb` holds one banned-claim list, shared by the rendered-page specs and the static-file specs. Two copies of a compliance rule is a slower way to get a contradiction.

They ban **assertions, not vocabulary**. A match negated within 60 characters is treated as a disclaimer, because "this is **not** Anthropic's commercial API" and "we do **not** claim a zero-data-retention guarantee" are the accurate sentences and are the ones a district reviewer most needs. A substring ban would have deleted exactly the honest disclaimers. The known limit is documented in the module: an unrelated negator nearby will suppress a hit. It is a drift backstop, not a proof of truthfulness, and it is tuned to avoid false alarms because a guard that cries wolf on accurate copy gets deleted and then catches nothing.

A final test ties rendered copy to the registry's active feature list, so re-enabling `eval_narrator` fails until the user-facing copy is updated in the same change.

### Negative-tested

Each guard was verified by reintroducing the specific defect:

| Reintroduced | Result |
|---|---|
| privacy.hbs names Anthropic's commercial API | 2 failures |
| privacy.hbs advertises Opus 4.7 | 3 failures |
| privacy.hbs asserts a ZDR agreement | 2 failures |
| en.yml presents eval data as sent | 2 failures |
| a locale file re-adds Gemini | 1 failure |

It also caught an unrelated `es.json` revert I caused during that work and had not noticed, which is the case it exists for.

### Why the artifact checks missed this

They only read the register and its generated artifacts. Nothing in CI read `privacy.hbs` (Ember, never loaded by Rails specs), `public/locales/*.json`, or the consent copy in `en.yml`. That is the hole these specs close. Correcting the registries genuinely did not correct what a parent reads.

### Scope

`main`-based hotfix only. `#725`/`#722` untouched. The identical defects still exist on `staging` and need the same fix ported before the next release PR, or this silently reappears.

105 examples, 0 failures across the disclosure and AiClient specs. All five register artifact checks green.

swahlquist:
## `privacy.hbs:76` fixed (b479d1655) — CI green

Confirmed the blocker. Word prediction does send; evaluation summaries do not, and the sentence bundled them together.

Rewritten so each feature gets its own accurate statement, plus the condition for turning eval on: word prediction sends under the separate AI consent; AI-drafted evaluation summaries are inactive and send nothing, and enabling live vendor egress would require an updated disclosure and a fresh consent first. The same string in all 13 `public/locales` files carried an older, vaguer variant and is now consistent.

### The guard was the real defect

`BANNED_CLAIMS` is an exact-phrase denylist, so it only knows wordings someone already caught. This sentence named no model, no vendor, and no retention term, so nothing matched. That is two consecutive review passes where the same false claim appeared in different words, which means adding the phrase would only have set up a third round.

Replaced with a **structural rule**: any sentence tying evaluation summaries to data leaving LingoLinq must also state the feature is inactive. It matches on shape, not vocabulary.

**Negative-testing caught a false negative in my first version**, which is worth reporting because the rule looked correct and was green against real code. I planted a novel paraphrase:

> "evaluation summaries are shared with our AI provider whenever a clinician requests one"

and it **passed**. A later clause in the same sentence mentioned producing a draft "on our own systems", and the inactivity marker accepted that as reassurance — a weak marker let a contradicting clause vouch for the claim. Two fixes:

- sentence splitting now includes semicolons, so a claim cannot borrow an adjacent clause's marker
- only strong markers count (`inactive`, `sends nothing`, `nothing is sent`, `never leaves`)

Locale checking also widened from `privacy_sharing_ai_vendors` alone to all five user-visible AI strings. Scoping it to one key is exactly how the stale `privacy_special_ai_consent_intro` shipped.

### Stated limit

Every pattern is English. The locale files are untranslated English fallbacks today so the guard sees them, but a genuinely translated false claim would pass silently. Translated compliance copy needs human review; this file does not substitute for it.

### Verification

| Reintroduced | Result |
|---|---|
| the exact sentence from this review | fails |
| a novel paraphrase naming no model or vendor | fails |
| a locale-file eval-egress variant | fails |
| all three fixed | passes |

107 examples, 0 failures. All five register artifact checks green. All six CI checks pass on `b479d1655`. Pushed head verified clean of the old string.

swahlquist:
## Governance exception on record: admin bypass of a functioning review gate

This PR was merged with `gh pr merge --squash --admin`, which bypassed `main`'s CODEOWNER review requirement.

**This was not a broken CI gate.** All six required checks passed on their own. The bypassed control was the human review requirement, and it was functioning as designed. `enforce_admins` is currently disabled on `main` (turned off 2026-07-15 for the Render-to-GCP migration and not yet restored), which is the only reason the bypass was mechanically possible.

Recording it explicitly rather than leaving it implicit in the merge metadata:

| | |
|---|---|
| Control bypassed | `main` CODEOWNER review, 1 required approval |
| Reason | production runtime AI non-functional; live consent page serving false processor, model, and retention claims |
| Authorized by | Scot, 2026-08-03 |
| Merged commit | `420aada88` |
| Reviews performed | Codex, three passes (`9abe7bd22`, `b479d1655`, and the pre-hotfix stack); no human CODEOWNER approval |
| Standing until | `main` review process is staffed |

The hotfix content was reviewed, and reviewed adversarially. What was skipped is the *human approval step*, not review as such. Treating this as an exception rather than a precedent: it should not become the default path for `main` hotfixes, and `enforce_admins` still needs restoring before MVP release.

Evidence that the bypass was warranted, captured from live production immediately before the deploy landed:

| Claim | Live in production pre-deploy |
|---|---|
| "commercial API" | present |
| "Claude Opus 4.7" | present |
| "zero-data-retention agreement" | present |
| "Amazon Bedrock" | absent |
| "Amazon Web Services" | absent |

The public consent page named neither the actual processor nor the actual host.

PR #737, opened 2026-08-03
fix(coppa): finish offboarding re-consent gaps for seat reclaim

MelissaOneil:
Stamp family COPPA on automated license expiry, require birth month/year server-side on org remove, and export-then-delete when a guardian declines or the consent window expires so district minors are not left on a consumer trial without a lawful basis.


swahlquist:
## Compliance PR Checklist

PR #744, opened 2026-08-04
fix(audit): decouple audit-merge --sha restamp from evidence anchoring + gate register structure in CI

swahlquist:
## What and why

Two defects handed off from the PR #742 session.

### 1. `audit-merge.rb --sha` was overloaded and corrupted evidence anchors

`--sha` carried two unrelated jobs: stamping `meta.auditedSha`/`auditedRef`/`auditedDate`, **and** overwriting every incoming finding's `evidence.sha`. Adding a finding outside a real `/audit-run` therefore had no correct invocation. Reproduced at `c5e85b2a5` before the fix, with a throwaway finding whose snippet sits at line 310 in both commits:

| Run | Invocation | `meta.auditedSha` after | `evidence.sha` |
|---|---|---|---|
| A | `--sha c5e85b2a5` (true HEAD) | **restamped** | `c5e85b2a5` |
| B | `--sha 20953ab3d` (register's own sha, to dodge the restamp) | untouched | **`20953ab3d`** — a commit the snippet was never verified against |

Run A falsely asserts a whole-tree re-audit, which `meta.auditedShaPriorNote` documents as a governance act needing Scot's sign-off plus an intervening-commit analysis. Run B is the dangerous one: `citation-check.rb` matches per LINE, so a wrong anchor passes **green** whenever the snippet lands on the same line in both commits. PR #742 only caught the equivalent case because line 182 vs 152 happened to differ, and worked around it by restoring the three `meta` keys by hand — which nothing enforced.

**Fix:** `--no-restamp`. Evidence anchors at `--sha`; `meta` is left byte-identical. Mirrors `promote-finding.rb`, which never touches the pointer for exactly this reason. `--no-restamp` with `--ref` now fails closed, since `--ref` only ever writes `meta.auditedRef`.

### 2. CI could merge a malformed register green

`audit-artifacts-integrity` ran only the render `--check` scripts. None of them reads the field *shapes* the register's consumers depend on. Reproduced: a finding whose `source` was a bare String instead of an object exits 0 under `citation-check.rb` and every render check, then hard-crashes `promote-finding.rb` for the **entire** register with `Hash#dig': String does not have #dig method (TypeError)` at `promote-finding.rb:372`.

**Fix:** two complementary gates, both stdlib-only and git-free so they are as cheap to gate on as the existing checks.

- `scripts/register-lint.rb` — structural validator (field shapes, enum membership against the register's *own* `meta` enums, id uniqueness). Precise, actionable errors. Runs over both registers.
- `scripts/tests/register-consumer-smoke-test.sh` — runs the real consumers over each committed register with empty input, asserting exit 0 **and** byte-identical output. Catches whatever the linter's predicate list failed to anticipate.

Both are wired into `ci.yml` and into `regenerate-register.sh`'s local verify bundle.

## Register content change (provenance labels only)

`register-lint` surfaced two pre-existing live rows whose `source.kind` sat outside the register's declared `meta.sourceEnum`. Each row's own `notes` settles the mapping, so these are label corrections — **status, severity, disposition, and evidence are untouched**:

| id | was | now | why |
|---|---|---|---|
| `LL-11db0dc848` | `audit` | `pr-review` | `pr: 412`; notes are `promote-finding.rb`'s verbatim PR-review provenance string |
| `LL-7f7372e3eb` | `audit-review` | `manual` | `pr: null`, hand-filed; identical shape to the existing `manual` rows |

No attested document's `attestedContentHash` pin is touched (only `contentHash` for `FINDINGS.json` and `audit-reports/README.md`, both `attestation: {}`), so **no re-attestation is owed**.

## Fix status

| Item | Status | Evidence |
|---|---|---|
| `--sha` no longer forces a false restamp | Fixed | `scripts/audit-merge.rb` `--no-restamp` opt + guarded meta block |
| `--sha` no longer silently re-anchors evidence to a stale commit | Fixed | same flag; `meta` verified byte-identical while `evidence.sha` = true commit |
| `--no-restamp` + `--ref` ambiguity | Fixed (fails closed) | `scripts/audit-merge.rb` die() on the combination |
| Summary artifact can no longer imply a re-audit that did not happen | Fixed | `summary.restampedMeta` |
| `/audit-run` behavior unchanged | Verified no-op | default-path output **byte-identical** to `git show HEAD:scripts/audit-merge.rb` over the same register |
| Malformed register row merges green | Fixed | `scripts/register-lint.rb` + `scripts/tests/register-consumer-smoke-test.sh`, both gated in `audit-artifacts-integrity` |
| Governance rule stated where callers read it | Fixed | `CLAUDE.md`, `audit-reports/README.md`, both `*-audit-run` SKILLs, promote-finding SKILL, ember-upgrade README |

## Verification

- Both defects reproduced **before** the fix (tables above)
- `--no-restamp`: `meta` byte-identical, `evidence.sha` = true commit
- Default path: byte-identical output vs the pre-change script over the same register
- Summary change purely additive (`restampedMeta` added; no key removed, all shared values equal)
- `register-lint` green on both live registers; catches all 12 mutation cases (String `source`/`disposition`, missing `evidence`, bad status/severity/framework/disposition-state, duplicate id, non-numeric `evidence.line`, non-String `notes`, non-object finding, non-array findings) and exits 1 on unparseable/missing files
- Smoke test green on both registers; negative control reproduces the exact reported `TypeError`
- `scripts/regenerate-register.sh` green — all eight checks including both new gates
- `ruby scripts/citation-check.rb` green; `git diff --check` clean; `ci.yml` parses; exec bit set on the directly-invoked test script

## Not covered by this PR

- `audit-merge.rb` and `promote-finding.rb` remain a hand-synced pair for their PII/secret patterns and governance constants (LEARNINGS L5567). This PR does not extract that shared logic; `--no-restamp` touches neither. Flagged, not fixed.
- `citation-check.rb` stays out of CI — unchanged policy; it needs git history, and gating it repo-wide is a broader decision than this PR.

## Author-Model: opus-5


swahlquist:
## Codex Review

**Verdict:** NEEDS_HUMAN
**Route:** codex
**Head:** cfc7bf6c2c555ab70eee4b1acecd5fa275774504
**Run:** https://github.com/lingolinq/LingoLinq-AAC/actions/runs/30956614018


Codex review completed.

_No findings reported._

PR #771, opened 2026-08-09
feat(board-picker,board-detail): 🏠 pick-for-home that works + 🔒 a watertight board lock

traciday:
## 🎯 What

Supervisors can now set a communicator's home board from `/board-picker` — a flow that had never worked — and the board lock ("Stay on this Board") is enforced on every exit from board-detail with its off-switch guaranteed reachable. Also fixes a production bug that permanently silenced UI feedback sounds, plus board-detail / board-picker / caseload styling and responsive work.

## 🔍 Why

- 🚫 **Pick-for-home never worked.** Three independent bugs stacked on each other, and the QA script meant to prove it could not run at all — it imports Playwright, the repo only committed Puppeteer.
- 🔒 **The board lock had a hole in the front door.** A supervisor *safety* control that keeps a communicator on the board their therapist chose, it had drifted to guarding 2 of the page's ~12 exits while still displaying as engaged. Found by round-4 adversarial review (`250186f3e`); the remainder by round 5, below.
- 🔇 **`speecher.load_sound` destroyed the URLs it was caching.** Any app boot with the extras layer unready wiped all twelve CDN feedback-sound URLs to `undefined`, silencing sounds for the rest of the session.

## 🧪 Testing

Run locally on the merged branch, not inferred from CI:

| Check | Result |
|---|---|
| `npx ember test` | ✅ **1892 tests, 1854 pass, 38 skip, 0 fail** |
| `bundle exec rspec` (`TZ=UTC`, CI-equivalent) | ✅ 6740 examples, 0 attributable failures |
| `spec/controllers/api/{boards,users,logs}_controller_spec.rb` | ✅ 669 examples, 0 failures |
| `npm run lint:js:ci` (new blocking gate) | ✅ `new=0` |
| `npm run lint:hbs` | ✅ clean |
| `audit-artifacts-integrity` (6 checks) | ✅ 6/6 pass |
| Board-picker QA script, real browser | ✅ **27/27**, supervisor→supervisee end to end |

Manually click-tested: supervisor pick-for-home (`?user_id=`), the self flow (no query param), and board-lock Back / Home / folder / drawer / sidebar exits with the lock on and off.

Two traps worth knowing:

⏰ **`TZ=UTC` matters.** `spec/lib/stats_spec.rb` builds its window from `Date.today` (local) against `.utc` timestamps. In a US timezone after midnight UTC, 10 examples fail locally that are green in CI.

🎲 **The suite has a wandering flake.** A red `ember test` job here is most likely that, not this code — the failing test moves every run. Detail under *Not covered*.

---

## 🗄️ Database migrations

- [x] **This PR contains no migration.**

## 🚀 Production impact

- [x] **No new environment variable or secret is required.** No `deploy-cloudrun.yml` secret-set change needed.
- [x] **Rollback is understood.** Frontend, three API controllers and views only — no schema change — so pinning the previous revision is sufficient and safe, with no backward-compatibility caveat to weigh. Check the current pin first (`gcloud run services describe lingolinq-web --region us-central1 --format='value(spec.traffic)'`), then `update-traffic --to-revisions REVISION_NAME=100`. If the cause is a bad commit, revert or fix it rather than re-running the deploy, since the next successful deploy moves traffic to its own verified revision and undoes the pin. Worker pool untouched.

---

## ✅ Fix status

| Item | Status | Evidence |
|---|---|---|
| Supervisor pick-for-home works end to end | ✅ Fixed | QA script 27/27 |
| Board-picker **self** flow | ✅ Fixed | `controllers/board-picker.js` `_resolve_setup_user` |
| Board lock enforced on all board-to-board exits | ✅ Fixed | `board_lock_blocks_exit()` ×5 call sites; 3 acceptance tests |
| Board-lock off-switch reachable while engaged | ✅ Fixed | `board_lock_control_visible` + section OR |
| Feedback sounds no longer destroyed on boot | ✅ Fixed | `speecher.js` `cache_sound_locally` |
| Cancelled drag no longer wedges the UI | ✅ Fixed | `raw_events.js` `cancel_drag` |
| `allowed?` / `allows?` double-render | ✅ Fixed | 3 controller specs, 669 examples |
| `runCancel` → `clearTimeout` on a native timer | ✅ Fixed | baseline shows one fewer `no-runloop` |
| Frontend lint gates green after the staging merge | ✅ Fixed | `new=0`; `lint:hbs` clean |
| Wandering `waitsFor` test flake | ❌ **Not fixed** | see below |

## ⚠️ Not covered by this PR

- 🎲 **The wandering test flake.** Provoked by this branch's new app-booting `board-lock-test`: acceptance modules boot the whole app, that boot mutates shared singletons, and the mutation outlives the module — killing a random later test on a ~5.5s `waitsFor` timeout, a different victim each run. Bisect: acceptance modules excluded → 2/2 clean; included → 4 of 5 runs flaked. **Four** harness-level remedies were tried and all failed (`QUnit.testDone` hung the suite; `QUnit.moduleDone` had no effect; auto-loader exclusion was a no-op because ES imports hoist; dynamic `import()` never resolves in this AMD build). Dead ends and the correct next direction — restore the specific mutated singleton, as the speecher fix did — are in `docs/task-management/LEARNINGS.md`.
- 🔓 **23 bare `allows?(@api_user, '…')` calls** in `app/controllers/` missing the scopes argument. Assessed: they fail **closed**, not open — bare `allows?` uses raw `permission_scopes` and appends `'*'`, which never intersects the `'full'` permission rules require. **Availability bugs, not vulnerabilities**, denying integration and dev-key devices. Not bulk-edited; several need per-site judgement.
- 🎲 **`copy_board_links` "protected vocabulary owner" spec**, flaky ~1 in 4. Pre-existing, unrelated.
- 🙋 **Four product decisions** needing a human: label truncation at 667×375; the ~2.7px symbol on very small cards; the removed consent-page CTAs (a legal surface — recommend restoring or explicit sign-off); two missing regression specs (cancelled touch/blur, label truncation at N viewports).

## 🚪 P2 — entry-point enumeration

The supervise-for-copy widening is a real authorization change.

| Entry point | Enforced at | Test |
|---|---|---|
| Fresh server-rendered navigation | `api/users_controller.rb` — supervise-only home-board slice | `spec/controllers/api/users_controller_spec.rb` (+62) |
| Client-side SPA transition (Ember) | Same server check; Ember flags are UX only | Board-picker QA script, 27/27 |
| Direct API call (raw params) | `api/{users,boards}_controller.rb`; `process_home_board` still gates on `board.allows?(self,'view')` or `allows?(updater,'share')` | `spec/controllers/api/boards_controller_spec.rb` (+23) |
| Offline / cached content | No new offline surface; the picker requires a live session | — |

**Not covered:** no new offline path was introduced, so none was tested. The widening still requires an existing supervise relationship; it grants no access to boards the supervisor could not already see.

## 🤖 Author-Model: opus-5

---

<details>
<summary>📦 <b>Change breakdown</b> — 101 files, +7025 / −1331 vs <code>staging</code></summary>

### 🏠 Board picker — supervisor sets a communicator's home board *(headline)*

First real browser execution surfaced three independent defects:

1. `Utils.uniq(list)` threw on any one-argument call — `compare.toString()` dereferenced before the type check — killing `User#org_board_keys`, which `edit_manager#copy_board` reads before copying. Every pick died with a generic toast and **zero** API calls.
2. `raw_events#modalDialogClickRelease` synthesized a click for `.modal-content` buttons; modern `{{on "click"}}` handlers also got the native click, so the CTA fired twice — two copies, the second `POST /boards` 400 "board key already in use", error shown *after* a copy existed.
3. The backend guard required the payload to contain **only** `preferences`; Ember's `user.save()` sends the whole record (~25 keys), so it never matched.

### 🔒 Board lock

Enforcement consolidated into one decision point, `board_lock_blocks_exit()`, called from every board-to-board exit: Back (checked *first*, before the in-session-history branch that used to transition and return), Home, folder navigation, the Board Collections drawer, and the quick sidebar.

⚠️ Deliberately **not** a route-level `willTransition` guard — that would also block leaving for settings, dashboard or logout, which this lock has never done and which would strand the user.

Unchanged and intentional: on a communicator-only account the Session section stays hidden, so a locked communicator cannot release their own lock. Documented at the site so it is not "fixed" later.

### 🔇 `speecher.load_sound`

`persistence.store_url_now` early-returns a bare `{url, type}` — no `local_url`, no `data_uri` — when the extras layer is not ready. `speecher[attr] = data.local_url || data.data_uri` therefore assigned **`undefined`**, destroying the CDN URL instead of caching it. Fixed by keeping the existing URL as the final fallback — the intent `load_sound`'s own error path documented but never applied on the success path.

### ⚙️ Backend — three API controllers

| File | Change |
|---|---|
| `boards_controller.rb` | `allowed?` double-render fix; supervise widening scoped to copies |
| `users_controller.rb` | supervise-only home-board update slice |
| `logs_controller.rb` | double-render fix; pass scopes to `allows?` |

No models, no libs, no migrations.

### 🎨 Styling / 🔧 tooling

Board-detail responsive work, label-fit, caseload, search, board-picker, consent-page layout, dashboard sections; `app.scss` +1781; 13 locale files regenerated. `.eslint-todo` regenerated for the new blocking gate; Playwright made a real devDependency so the QA script can run.

</details>

<details>
<summary>🕵️ <b>Adversarial reviews</b> — five rounds, findings and rejections</summary>

Every round's findings were verified against the code before acting. Several were wrong; those are listed too, because "reviewer said so" is not evidence.

**Rounds 1–3** — `find-a-button`, speak bar, sidebar editor, guided tours, eval/org views. Commits distinguish real findings from false positives, with the false positives documented in code rather than silently ignored (`fix(review): …`).

**Round 4 (`250186f3e`)** — board lock menu items had been dropped from the template while the page kept enforcing the lock (`toggle_sticky_board` was live code reachable from no UI), and the check had drifted to 2 of ~12 exits. Also: a cancelled touch activated board chrome.

**Round 5 (this PR)** — five findings, four confirmed, one rejected:

| # | Finding | Verdict |
|---|---|---|
| 1 | "Assign a Home Board For Me" broken for the **self** flow | ✅ Fixed |
| 2 | Cancelled edit-mode drag wedges the UI | ✅ Fixed |
| 3 | Board-lock toggle hidden with the Session section | ✅ Fixed |
| 4 | `runCancel` on a `setTimeout` token | ✅ Fixed |
| 5 | Progressive render breaks root filtering | ❌ False positive |

**1️⃣ The self flow.** `setupController` had lost `controller.set('setup_user', currentUser)`, replaced by `observer('user_id', 'appState.currentUser')` — which cannot fire on the self flow, since opening `/board-picker` with no `user_id` changes neither key. `setup_user` stayed null and the action short-circuited to *"Home board update failed unexpectedly"*.

⚠️ Why it survived: the QA script reporting **27/27** runs `--supervisee`, i.e. the query-param path only. **The self path was never exercised.** A green QA run stood in for proof the feature worked when it had only tested half of it.

**3️⃣ How the board lock regressed twice.** `dc67d28aa` removed `sticky_board` from the customize-menu registry **and** from the section's visibility check — correct and self-consistent at the time. `250186f3e` restored the item and its buttons but not the visibility term, and left a comment asserting the item *"was removed from this menu"*. That false comment is almost certainly why the omission stayed invisible. Deleted, and the term restored.

**5️⃣ Why #5 was rejected.** The claim was that `filterRootBoards` "requires the fully accumulated set". Tracing `board-roots.js:33-65`, the two passes collapse to a per-board decision — rule 2 drops any board with `copy_id && copy_id != id` regardless of the list, and rule 1 only fires where rule 2 already would, or self-referentially. No such documentation exists either. The predicted symptom cannot occur, so no change was made.

</details>


swahlquist:
## Compliance PR Checklist

MelissaOneil (approved):

PR #782, opened 2026-08-11
fix(flusher): sweep LogSnapshot on user erasure (LL-1e2ab28aab)

### LINKED ISSUE #775: Flusher: sweep LogSnapshot on user erasure (LL-1e2ab28aab)
Follow-up to #721, which added the `ButtonSound` / `UserVideo` sweeps. Same class of gap, one model further.

## The problem

`LogSnapshot` rows survive user erasure. They are keyed by `user_id` and created through `api/snapshots_controller.rb#create`, but nothing deletes them when the user is deleted:

- `Flusher` never references `LogSnapshot` (`lib/flusher.rb`)
- `user.rb` declares no `dependent: :destroy` for it
- `log_session.rb` declares no `has_many` at all
- `db/schema.rb` declares **no foreign keys**, so there is no DB-level cascade either

A hard delete therefore leaves a row carrying a supervisor-supplied label, a date range, a `device_id`, and a `location_id`, keyed to a user that no longer exists.

Tracked as **LL-1e2ab28aab** (medium, accepted). Context in #774.

## Scope of this task

1. Add a `LogSnapshot.where(user_id: user.id)` sweep to `Flusher.flush_user_content`, alongside the `ButtonSound` / `UserVideo` sweeps added in #721. Same `flush_record` pattern.
2. Add regression coverage proving a hard delete removes the user's `LogSnapshot` rows.

That is the whole task.

## Explicitly NOT in scope

Please don't touch any of these; they are handled separately and some are CEO-only:

- `audit-reports/FINDINGS.json` — including marking this finding closed. Closure is Scot's.
- Register metadata, `DOCUMENT-REGISTER.json`, or any `contentHash` / attestation field
- Bundle definitions or bundle policy
- Anything under `docs/legal/`, especially the attested files and the `2026-08-09_*` successor drafts
- `UserExtra` and `ContactMessage`. They are named in the same unverified set but were never traced, so no claim exists about them either way. If you want to look at them, treat it as separate work rather than widening this one.

## Notes

- The retained `location_id` already dangles: `ClusterLocation` is swept at `lib/flusher.rb:19`. So this is about the snapshot row itself, not the location it points at.
- `LogSnapshot` is individually deletable by the user via the snapshots API today. The gap is only in the erasure path.
- No S3 objects are involved, unlike `ButtonSound` / `UserVideo`, so no `Uploadable` interaction to worry about.

MelissaOneil:
## Summary
- Add `LogSnapshot.where(user_id:)` sweep to `Flusher.flush_user_content`, same pattern as the ButtonSound / UserVideo work from #721.
- Add stub expectation + regression coverage that a hard-delete flush removes only that user's snapshots.
- No S3 / Uploadable path; rows only. Does not touch FINDINGS.json, the document register, or `docs/legal/**`.

Closes #775. Finding LL-1e2ab28aab remains for Scot to close after verify.

## Fix status
| Item | Status | Evidence |
|---|---|---|
| LogSnapshot swept on `flush_user_content` | Fixed | `lib/flusher.rb` (LogSnapshot block after UserVideo) |
| Regression coverage | Fixed | `spec/lib/flusher_spec.rb` — flush_user_content examples |
| FINDINGS.json / register / docs/legal | Not fixed (out of scope) | issue #775 explicit exclusions |

## Not covered by this PR
- Closing LL-1e2ab28aab in the findings register (Scot)
- `transfer_user_content` for LogSnapshot (not requested; erasure path only)
- UserExtra / ContactMessage Flusher gaps

## Test plan
- [x] `DB_USER=melis bundle exec rspec spec/lib/flusher_spec.rb` (53 examples, 0 failures)
- [ ] CI rspec green

## Author-Model: grok-4.5


swahlquist:
## Compliance PR Checklist

- [ ] Retention policy reviewed (flusher.rb changed)

PR #793, opened 2026-08-12
release: staging into main 2026-08-12 (deploys the Bedrock BAA account assertion)

swahlquist:
Promotes `staging` (`5d6ec6042`) to `main`. **Merging this deploys to production automatically** — `deploy-cloudrun.yml` triggers on `push: branches: [main]`.

The stated purpose is to land the Bedrock BAA account assertion (finding `LL-1b0d78dbe6`, #768/#785) in production. That is one of roughly fourteen PRs in this promotion; the rest are listed below so nobody reads this as a Bedrock-only release.

## Pre-merge verification

| Check | Result |
|---|---|
| Article 50 disclosure NOT enabled | **Confirmed.** `lib/feature_flags.rb` is byte-identical between `main` and `staging`. `article_50_disclosure` remains in `AVAILABLE_FRONTEND_FEATURES` (`:50`) and absent from `ENABLED_FRONTEND_FEATURES` (`:57+`). This release cannot enable it. |
| Prod hotfix #734 (classic Bedrock plane) preserved | **Confirmed** by trial merge. |
| Prod hotfix #726 (button-set destruction) preserved | **Confirmed.** `app/models/board_downstream_button_set.rb` is unchanged by the merge. |
| Merge conflicts | **None.** Trial merge with `--no-commit --no-ff` completed clean. |

The trial merge removes exactly 10 lines from `main`'s `lib/ai_client.rb`. All 10 are from the #768 refactor: the inline `Rails.logger.warn` guard in `runtime_model` (extracted to `emit_log`) and the three-line region resolution (moved into `bedrock_region` with format validation). None touch plane selection, so #734's classic-plane routing is intact.

## What ships

**Bedrock / AI**
- #768 `sts:GetCallerIdentity` account assertion; fails closed on mismatch, unverifiable probe, malformed account, or malformed region. Deploy pins `BEDROCK_EXPECTED_AWS_ACCOUNT=239044785114` and reads it back on the web service and the worker pool.
- #785 closes `LL-1b0d78dbe6`; headline 12 High → 11 High.
- #788 closes the code side of `LL-16ef84ad9a` — the word-prediction cache no longer keys on the raw pre-scrubber utterance, is organization-scoped, sweeps on TTL, and is mutex-guarded.
- #789 Render dev/staging Bedrock config. **No production effect**: the manifest excludes Render prod, and production runtime is Cloud Run.

**Security / compliance**
- #776 stops the seed granting site admin to the example user
[...]

PR #796, opened 2026-08-12
DIAGNOSTIC (do not merge): identify why persistence test 1473 fails in CI

swahlquist:
**Do not merge. Diagnostic only.** This exists to make one CI failure legible; it will be reverted.

## Why

`persistence DSAdapter updateRecord - should update a locally-created record that hasn't been persisted yet` (test 1473) fails on every recent run of `staging` and on all three runs of the release PR #793.

It is **not** a regression from #791, #788, or #789. The same test failed on `staging` before any of them merged:

| Tree | Result |
| --- | --- |
| `f65bfd515` 20:08:12 | pass (last green) |
| `89b33370d` 20:08:30 | fail |
| `07ab1b24e` (#788 merge) | fail |
| `1991380d6` (#791 merge) | fail |
| `5d6ec6042` (#789 merge) | fail |
| #793 x3 | fail |

Seven consecutive failures. First appearance was `3a7ab8350` (#781) after 13 consecutive passes.

## Why instrumentation rather than a fix

The failure cannot be reproduced locally: 8/8 passes in isolation, 6/6 across the whole `persistence` module, 4/4 alongside `board-detail-cache`. The full suite will not run locally (Acceptance and `app_state` tests hang under a bare puppeteer harness), so CI is the only place it exists.

The failure is a **rejection, not a run-loop hang** — `expect(final_error).toEqual(null)` receives an object, with `TypeError: Cannot read properties of null (reading 'id')` in the browser log. `persistence.find` has seven rejection paths and the assertion currently prints only `[object Object]`.

## What this captures

Into the assertion value, so the TAP reporter prints it:

- which path fired: `SAVE_RESOLVED_NULL` / `SAVE_REJECTED` / `FIND_REJECTED`, plus the serialized error
- module-vs-`window` persistence divergence, which decides the branch at `persistence.js:4511`
- `known_missing` on both roots, the first check `find` performs
- `extras_ready`, covering the "extras not ready" branch

Existing assertions are untouched, so a pass is still a genuine pass.

## Note on .eslint-todo

Regenerated because the lint gate runs before the Ember test step; a stale baseline fails the build and the tests never run. The regenerate **removes 11 stale `persistence-test.js` entries and adds none** — leftovers from #791 replacing `later(` with `setTimeout(`. Nothing new is grandfathered; gate reports `new=0`.

## Fix status
| Item | Status | Evidence |
| --- | --- | --- |
| Root cause of test 1473 | Not fixed (diagnostic only) | this PR adds no fix |
| `_mineListBusy` leak (separate) | Not fixed | `app/controllers/user/index.js:1062-1067` |

## Not covered by this PR
- Any fix. This is instrumentation only and must be reverted.
- The `_mineListBusy` stranding bug found while investigating (`user/index.js:1062-1067`): the success path early-returns without clearing and the failure path clears only on `list_id` match, so a superseded Mine fetch can strand the flag `true` permanently, making `_wait_while_mine_list_busy` poll for its full 120s cap. Reported separately; not confirmed as the cause of this test failure.

## Author-Model: opus-5

### DIFF
--- a/app/frontend/.eslint-todo
+++ b/app/frontend/.eslint-todo
@@ -1,7 +1,7 @@
 # ESLint baseline (grandfathered findings). Do not hand-edit.
 # Regenerate: npm run lint:js:todo
 # Format: file|ruleId|line|column|severity|messageHash
-# Generated: 2026-08-11T17:43:30.146Z
+# Generated: 2026-08-12T12:47:57.694Z
 app/app.js|ember/no-runloop|102|5|2|0843889163d3
 app/app.js|ember/no-runloop|168|7|2|0843889163d3
 app/app.js|ember/no-runloop|193|7|2|0843889163d3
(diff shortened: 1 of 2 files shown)


swahlquist:
Purpose served. The instrumentation identified the root cause: `fake_dbman.store_internal` deleted the record and re-inserted it only after a randomized `wait_call` delay, leaving a window in which a concurrent `find()` correctly returned `{error: "no record found"}`.

Captured payload from the failing run:

```
FIND_REJECTED | mod_online=false | win_present=true | win_is_mod=false | win_online=false
| win_destroyed=false | rec_id="tmp_0.79005943527877881786541854782" | km_mod_hit=true
| km_win_hit=true | km_mod_n=1 | km_win_n=1 | extras_ready=true
| detail={"error":"no record found"}
```

That ruled out the online/offline divergence at `persistence.js:4511` (both roots read false), `known_missing` contamination (`km_n=1`, and the one key is the record's own, set by `find`'s own reject handler), and extras-not-ready.

Fix is in #797. Closing unmerged as intended.

